PR #6, opened 2019-01-25
Google Storage index support

tom-butler:
add gs-to-tar,

This tool will find a list of files for a prefix, filter them to find the metadata, and add them to a tarball, for later use in the dc-index-from-tar tool. 

updated dc-index-from-tar to support multiple protocols.

### DIFF
--- a/dea/apps/index_from_tar.py
+++ b/dea/apps/index_from_tar.py
@@ -42,7 +42,7 @@ def from_tar_file(tarfname, index, mk_uri, mode, **kwargs):
               is_flag=True, default=False)
 @click.option('--gzip', is_flag=True, help='Input is compressed with gzip (needed when reading from stdin)')
 @click.option('--xz', is_flag=True, help='Input is compressed with xz (needed when reading from stdin)')
-@click.option('--protocol', type=str, default='s3://', help='Override the protocol for working with data in other environments, i.e gs://')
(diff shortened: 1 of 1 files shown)


dea/apps/index_from_tar.py
```diff
@@ -51,7 +52,8 @@ def cli(input_fname,
         verify_lineage,
         ignore_lineage,
         gzip,
-        xz):
+        xz,
+        protocol):
```
Kirill888: I think we should also accept just `s3` or `gs` for `--protocol` option.

```python
if not protocol.endswith('://'):
    protocol = protocol + '://'
```

setup.py
```diff
@@ -19,10 +19,12 @@
                       'lmdb',
                       'aiobotocore',
                       'botocore',
+                      'google-cloud-storage'
```
Kirill888: how big is this dependency, should we consider making that an optional dependency?

dea/apps/index_from_tar.py
tom-butler: rasterio currently supports the following schems:

```python
SCHEMES = {
    'ftp': 'curl',
    'gzip': 'gzip',
    'http': 'curl',
    'https': 'curl',
    's3': 's3',
    'tar': 'tar',
    'zip': 'zip',
    'file': 'file',
    'oss': 'oss',
    'gs': 'gs',
}
```

tom-butler: Done

PR #7, opened 2019-01-30
dc-index-from-tar '--protocol' option fixes

tom-butler:
- Pass protocol through to the mk_url function
- Add default to help output
- generalise function names


dea/apps/index_from_tar.py
```diff
@@ -6,14 +6,14 @@
 from dea.index import from_yaml_doc_stream
 
 
-def from_tar_file(tarfname, index, mk_uri, mode, **kwargs):
+def from_tar_file(tarfname, index, protocol, mk_uri, mode, **kwargs):
```
Kirill888: This is not needed, the idea is that `mk_uri` captures whatever extra parameters be it `protocol` or credentials for url shortener.

dea/apps/index_from_tar.py
```diff
@@ -64,7 +64,7 @@ def cli(input_fname,
     if ignore_lineage:
         auto_add_lineage = False
 
-    def mk_s3_uri(name):
+    def mk_uri(protocol, name):
```
Kirill888: `protocol` is already captured don't pass it in

Also move `protocol` normalisation if statement out of this function, no point doing it on every call.

PR #8, opened 2019-01-31
gs-to-tar preserves directory structure, make logs docker friendly

tom-butler:
Preserves directory structure so files will be in `gs://bucket/filepath.yaml` format when loaded using dc-index-from-tar.

Modify how logs are written as flushed logs don't show up in docker orchestration tools (such as kubernetes)

### DIFF
--- a/dea/apps/gs_to_tar.py
+++ b/dea/apps/gs_to_tar.py
@@ -1,6 +1,7 @@
 import os
 import tarfile
 import click
+import shutil
 from google.cloud import storage
 
 @click.command('gs-to-tar')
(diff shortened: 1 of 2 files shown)


Kirill888:
Hold on, are you writing files to disk before adding to tar?

Cause I already have code that writes directly to tar without going to filesystem first, also preserving timestamp of the file that s3 reports. 

Also why does this include changes from pervious PR?

Kirill888:
I'll merge, but we need to change tar to not go to disk, I'll take out the write side of tar into lib.

tom-butler:
Fixed bad merge, 

This is for a small prototype to evaluate the capabilities and compatibility of Datacube and Google cloud. Once we have a tech demo available I fully intend to integrate the google cloud indexing scripts better, and modify them to follow the find -> tar -> index process the s3 scripts follow.

PR #10, opened 2019-02-21
Add prototype THREDDS-to-tarfile script

GypsyBojangles:
Add prototype THREDDS-to-tarfile script to allow downloading ODC dataset yaml files from THREDDS endpoints and prepare them for indexing via index-from-tar.

### DIFF
--- a/dea/apps/thredds_to_tar.py
+++ b/dea/apps/thredds_to_tar.py
@@ -1,0 +1,83 @@
+import os
+import tarfile
+import click
+import shutil
+import urllib.request
+
+from urllib.parse import urlparse
+from thredds_crawler.crawl import Crawl
+
+
+@click.command('thredds-to-tar')
+@click.option('--thredds_catalogue', '-c', type=str, required=True, help="The THREDDS catalogue endpoint")
+@click.option('--thredds_fileserver', '-f', type=str, required=True, help="The THREDDS HTTP fileserver endpoint")
+@click.option('--skips', '-s', type=str, multiple=True,
+              help="Pattern to ignore when THREDDS crawling")
+@click.option('--select', '-t', type=str, required=True,
+              help="Target file pattern to match for yaml")
+@click.option('--workers', '-w', type=int, default=4, help="Number of thredds crawler workers to use")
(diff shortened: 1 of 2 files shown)


Kirill888:
It's possible to write directly to tar without going to FS first:

https://github.com/opendatacube/dea-proto/blob/master/dea/io/tar.py#L63

dea/apps/thredds_to_tar.py
```diff
@@ -0,0 +1,89 @@
...
+    Example:
+
+       \b
+       Download files in directory that match `*yaml` and store them as a tar
+        > thredds-to-tar -c "http://dapds00.nci.org.au/thredds/catalog/if87/2018-11-29/"
+        -f "http://dapds00.nci.org.au/thredds/fileServer/" -t ".*ARD-METADATA.yaml" -s '.*NBAR.*' -s '.*SUPPLEMENTARY.*'
+         -s '.*NBART.*' -s '.*/QA/.*' -w 8 --outfile 2018-11-29.tar.gz
+
+    """
+    file_retriever = urllib.request.URLopener()
```
Kirill888: Is this thing ok to share between threads? Or is it better to have one per thread?

GypsyBojangles:
Now streams all files to memory, then outputs to tar (I found i couldn't stream directly to tar without corrupting the archive). 
Reduced the number of required parameters as it should be possible to infer the THREDDS HTTP file server endpoint.

dea/apps/thredds_to_tar.py
GypsyBojangles: This is NOT threadsafe and has been changed.

Kirill888:
Yes `tar.write` is not thread-safe, options are (1) locking in app code (2) decoupling Read/Write threads via `Queue` (3) load all concurrently then write all serially

(2) is more likely to be performant and allows for cleaner "back-pressure" control. 

Given that we now have three different small file datasources that all go to tar it's worth making tar output into a class. I might take out logic in `s3-to-tar` into a  thread-safe class at some point. For now this solution will do, so long as you are not reading too much at a time and it's a "prototype" just to get some data into db to play with.

PR #11, opened 2019-05-27
Metadata export to new structure

ashoka1234:
### Problem
This is a script to transform metadata in datacube db onto the newish structure, in particular involving `grid` based structure of bands. 

### Solution
The implementation involves two code flows, one for ingested products and another for non-ingested products, though both have substantial overlap of code at a higher level. The difference of the two flows is with the way `grids` are computed. For `ingested` products grids are computed once while for indexed products, the grids are computed for every dataset. Grids are specified in a config file, and those bands not specified in the config is assumed to have `default` grid. Consistency of bands of the same grid is currently not verified.

Output `metadata` is currently written to a yaml file in the specified output directory and this directory contents currently has flatter structure which may need change depending on the context. Some sections of the new dataset may still need tuning, in particular `properties` section. Do we need to automate extraction of all possible `properties`? For example, every field of `metadata-doc` that does not fall onto another section could be viewed as a property and placed onto properties section. 

### Testing
Only manual testing is currently done. Test scripts could be written if required.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/export_md.py
+++ b/apps/dc_tools/odc/apps/dc_tools/export_md.py
@@ -12,10 +12,13 @@ The config file contains the grid mappings to band names:
 from pathlib import Path
 import yaml
 import click
+import logging
 
 from datacube.testutils.io import native_geobox
 from datacube.storage import BandInfo
 from datacube import Datacube
+
+_LOG = logging.getLogger(__name__)
 
 
 @click.group(help=__doc__)
@@ -39,7 +42,7 @@ def transform(index, product, config, output_dir, limit):
     dataset_type = index.products.get_by_name(product)
 
     with open(config, 'r') as config_file:
-        cfg = yaml.load(config_file)
+        cfg = yaml.load(config_file) or dict()
 
     # Is this a ingested product?
     if dataset_type.grid_spec is not None:
@@ -56,13 +59,17 @@ def transform_ingested_datasets(index, product, config, output_dir, limit):
 
     dataset_ids = index.datasets.search_returning(limit=limit, field_names=('id',), product=product)
 
-    for cc, dataset_id in enumerate(dataset_ids):
+    grids_done = False
+    for dataset_id in dataset_ids:
 
         dataset = index.datasets.get(dataset_id.id, include_sources=True)
 
-        if cc == 0:
+        if not grids_done:
             # setup grids
             grids = get_grids(dataset, config.get('grids'))
+
+            # got to try to compute grids until shape is not default [1, 1]
+            grids_done = all([[1, 1] != grid['shape'] for _, grid in grids['grids'].items()])
 
         dataset_sections = (grids,) + _variable_sections_of_metadata(dataset, config)
         _make_and_write_dataset(get_output_file(dataset, output_dir), *dataset_sections)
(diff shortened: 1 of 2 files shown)


apps/dc_tools/odc/apps/dc_tools/export_md.py
```diff
@@ -0,0 +1,264 @@
...
+from datacube.testutils.io import native_geobox
+from datacube import Datacube
+
+
+@click.group(help=__doc__)
+@click.option('--datacube-config', '-c', help="Pass the configuration file to access the database",
+              type=click.Path(exists=True))
+@click.pass_context
+def cli(ctx, datacube_config):
+    """ Used to pass the datacube index to functions via click."""
```
Kirill888: This string is used by `click` to provide `--help` output, so it needs to be user friendly.

apps/dc_tools/odc/apps/dc_tools/export_md.py
```diff
@@ -0,0 +1,264 @@
...
+def transform_ingested_datasets(index, product, config, output_dir, limit):
+    """
+    Transform the metadata of ingested product. The product-wide fixed sections
+    of metadata such as 'grids' is computed just once.
+    """
+
+    dataset_ids = index.datasets.search_returning(limit=limit, field_names=('id',), product=product)
+
+    # Compute 'grids' section
+    dataset_id = next(dataset_ids)
```
Kirill888: This code duplication is a bit confusing.

Would be easier to read in this format:

```python
for cc, datasets_id in enumerate(datasets_ids):
    if cc == 0:
      ..do_setup_things..
    ..do_usual_things..
```

This is python `if` inside loop is not as costly in relative terms.

apps/dc_tools/odc/apps/dc_tools/export_md.py
```diff
@@ -0,0 +1,264 @@
...
+      }
+    }
+    """
+    grids_map = {m: grid for grid in band_grids for m in band_grids[grid]}
+    measurements = dataset.measurements
+    for m in measurements:
+        if grids_map.get(m):
+            measurements[m]['grid'] = grids_map[m]
+        else:
+            measurements[m]['grid'] = 'default'
```
Kirill888: We prefer to not include `default` grid, it is assumed.

Kirill888: Also, can we have a check that removes `band` key if it's equal to `1`

apps/dc_tools/odc/apps/dc_tools/export_md.py
```diff
@@ -0,0 +1,264 @@
...
+      {
+        'datetime': time,
+        'odc:creation_datetime': creation_time,
+        ...
+      }
+    }
+    """
+    props = dict()
+    props['datetime'] = [dataset.time.begin, dataset.time.end]
+    props['odc:creation_datetime'] = dataset.indexed_time
```
Kirill888: Can we change this to `odc:processing_datetime`

apps/dc_tools/odc/apps/dc_tools/export_md.py
ashoka1234: Done

apps/dc_tools/odc/apps/dc_tools/export_md.py
ashoka1234: Done

apps/dc_tools/odc/apps/dc_tools/export_md.py
ashoka1234: Done

apps/dc_tools/odc/apps/dc_tools/export_md.py
ashoka1234: I have written something like
```
# Remove 'band': 1 if present
        if measurements[m].get('band') == 1:
            measurements[m].pop('band', None)
```
Hope this is what you meant! Could this value be '1' instead of 1. I have noticed `ls5_level1_scene` has `number: '1'` for measurement band info. Could this `number` field related to this issue?

Kirill888: Good point it can be all sorts of weird, so proper way is to construct `datacube.storage.BandInfo` class for every measurement and record `.band` and `.layer` instead of getting directly from measurement dict. `BandInfo` will perform proper resolving of the `.band` including dealing with stacked netcdfs.

https://github.com/opendatacube/datacube-core/blob/develop/datacube/storage/_base.py#L66

Kirill888: `.layer` can be `None`  (tiff images) and will need to be omitted in that case.

ashoka1234: Done using BandInfo ( Also omitting `band` if it is `None`)

apps/dc_tools/odc/apps/dc_tools/export_md.py
```diff
@@ -0,0 +1,271 @@
...
+        ...
+      }
+    }
+    """
+    grids_map = {m: grid for grid in band_grids for m in band_grids[grid] if grid != 'default'}
+    measurements = dict()
+
+    for m in dataset.measurements:
+        band_info = BandInfo(dataset, m)
+        measurements[m] = {'path': band_info.uri}
```
Kirill888: this is fully resolved path (absolute uri), we want to record relative path though.

PR #16, opened 2019-06-24
Index Replacement Enhancement Proposal Edits

omad:
I think this still needs a lot of tightening up, but it also needs to be shared around...

### DIFF
--- a/docs/db-core-interface.md
+++ b/docs/db-core-interface.md
@@ -1,32 +1,52 @@
-# Introduction
+# ODC Enhancement: Replace the Database API
 
-**This is an evolving document**
+**Date** 2019/06/12
 
-- Why do we need to change DB layer
-- How we gonna change it
+**Author** Kirill Kouzoubov  (@Kirill888)
 
-# Problems with the Current State
+**Version** >= ODC 1.7
 
-## Interface is poorly defined
+# Summary
 
-Supposedly there is high-level|low-level split in the DB driver and it should be
-possible to define a new low-level driver that uses on non-pg backend, in
-practice the responsibility of hi|lo levels is not well understood/articulated,
-and with only one low-level backend we are not sure how leaky this boundary is.
+The current Database API, used throughout the ODC Core codebase, is
+responsible for all access to the ODC Index. It was designed to be highly
+flexible in terms of Metadata it can handle, while maintaining high
+performance access to search across multiple dimensions.
 
-## Too much complexity pushed into DB driver
+Unfortunately, it is showing it's age, and has several assumptions and design
+flaws which are limiting future ODC development.
 
-- Lineage traversal/verification
-- Dataset to product auto-matching
-- The whole metadata indirection mechanism
+This proposal is for a complete replacement of the Database API.
 
-"Metadata indirection" is of particular concern as far as alternative backend
-implementations go. User defined search fields with custom aggregate functions
-are way too complex. It is essentially a json DSL that gets compiled into
-SqlAlchemy expressions, too hard to re-implement, might not even be possible for
-simpler db backends like sqlite.
+# Background
 
-This is an example from current EO metadata spec:
+The current Data
+
+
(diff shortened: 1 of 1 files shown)


docs/db-core-interface.md
```diff
@@ -44,101 +75,107 @@ This is an example from current EO metadata spec:
...
+Broken down, this means something like:
 
-- Define search field named `lat` which is of type `Range<float>`
+- Define a search field named `lat` which is of type `Range<float>`
   - where range starts at: `min(data.extent.coord.ur.lat as float, data.extent.coord.lr.lat as float, ...)`
   - and stops at:  `max(data.extent.coord.ur.lat as float, data.extent.coord.lr.lat as float, ...)`
 
-This then gets compiled into a sql expression when queries are made.
+This is then compiled into an SQL expression to generate table indexes per
+product, and and an identical expression when queries are made.
```
Kirill888: and and

docs/db-core-interface.md
```diff
@@ -44,101 +75,107 @@ This is an example from current EO metadata spec:
...
-    but can not change locations of individual bands
+    - Can not store lineage information for *external datasets*
+    - No way to reference an existing lineage dataset. It's necessary to include
+      the entire metadata document of the parent datasets.
+- Too much immutability
+    - Limited ability to change indexed data in place
+    - Not possible to delete or rename *Products* or *Metadata Types*
+    - Inconsistent mutability: eg. it's possible to change the location for
+      a while dataset, but not change locations of individual bands
+    - This is really frustrating for anyway learning or experimenting with ODC.
```
Kirill888: anyway anyone

alexgleith:
Hey @Kirill888 and @omad, I've added a summary section after the Introduction.

Please add/remove/clarify if needed... I think it's ready to share, though.

docs/db-core-interface.md
```diff
@@ -2,35 +2,40 @@
...
+Instead of storing heterogeneous dataset structures in the database with a lot of
+flexibility in what is indexed (in a database index sense), a new database
+structure with required fields stored as database columns, and only ancilliary
+data in the dynamic `json` metadata column will be implemented. This will result
+in simpler, more maintainable code in the ODC Core, much simpler indexes and easier and likely faster
+querying.
 
+Key changes suggested are documented in detail below, but include the following:
 
+- Add required *geometry* field(s) with defined CRS
```
Kirill888: geometry is not a required field, we still need to support non-raster datasets like telemetry data for provenance data, rather when indexing geo-registered raster data geometry is passed on to the DB layer as is, in the native form rather then being reduced to a lat/lon bounding box. Same goes for spatial component of the query.

docs/db-core-interface.md
```diff
@@ -2,35 +2,40 @@
...
+- Add required *temporal* field(s) with defined time zone
+- Remove custom indexing and replace with standard indices
```
Kirill888: We still support querying with arbitrary/custom dataset properties, we removing ability to query qarbitrary json subtree and removing "vritual preperties" which a aggregations of some json subtrees.

docs/db-core-interface.md
```diff
@@ -1,32 +1,68 @@
...
+assumptions and design flaws which are limiting future ODC development.
+
+This proposal is for a replacement of the ODC Index and Database API and
+a change in the internal model of *Datasets*.
+
+## Summary
+
+Instead of storing heterogeneous dataset structures in the database with a lot of
+flexibility in what is indexed (in a database index sense), a new database
+structure with required fields stored as database columns, and only ancilliary
```
Kirill888: We are not trying to mandate DB structure, rather we are aiming to define an internal API between core and DB that would allow for more efficient DB structures, particularly for spatial data.

alexgleith: Ok, I seem to have over simplified. I'll need to talk to you and Damien and get on the same page.

PR #18, opened 2019-07-12
Add EO3 support for dc-index-from-tar

Kirill888:
New command line option `--eo3` that will deal correctly with the new style metadata.

Computes `grid_spatial` and `extent` before creating Dataset objects to be added to the DB

Limitations:
- lineage support is not implemented


Kirill888:
@santoshamohan FYI

santoshamohan:
Thanks Kirill!

PR #37, opened 2020-04-01
Feat/refactor s3 find

whatnick:
Async await does not yet work for this refactor.

```
s3-find s3://dea-public-data/geomedian-australia/v2.1.0/L8/**/*.yaml

Traceback (most recent call last):
  File "/home/ubuntu/miniconda3/envs/odc_tools/bin/s3-find", line 11, in <module>
    load_entry_point('odc-apps-cloud', 'console_scripts', 's3-find')()
  File "/home/ubuntu/miniconda3/envs/odc_tools/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/odc_tools/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/miniconda3/envs/odc_tools/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ubuntu/miniconda3/envs/odc_tools/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/ubuntu/odc-tools/apps/cloud/odc/apps/cloud/s3_find.py", line 38, in cli
    stream = s3_uri_to_stream(uri, skip_check)
  File "/home/ubuntu/odc-tools/libs/aio/odc/aio/__init__.py", line 367, in s3_uri_to_stream
    raise e
  File "/home/ubuntu/odc-tools/libs/aio/odc/aio/__init__.py", line 363, in s3_uri_to_stream
    for i, o in enumerate(stream):
  File "/home/ubuntu/odc-tools/libs/aio/odc/aio/__init__.py", line 478, in find
    raise IOError(f"Failed to list: {url}")
OSError: Failed to list: s3://dea-public-data/geomedian-australia/v2.1.0/L8/
```

### DIFF
--- a/libs/aio/odc/aio/__init__.py
+++ b/libs/aio/odc/aio/__init__.py
@@ -1,3 +1,5 @@
+import logging
+
 import aiobotocore
 import asyncio
 from types import SimpleNamespace
@@ -292,7 +294,7 @@ async def s3_walker(url, nconcurrent, s3,
     return step
 
 
-def s3_uri_to_stream(uri: str, skip_check: bool):
+def s3_find_uri(uri: str, skip_check: bool):
     """Build generator from supplied S3 URI pattern
     Arguments:
         uri {str} -- S3 URI to create generator for
@@ -332,7 +334,7 @@ def s3_uri_to_stream(uri: str, skip_check: bool):
     try:
         qq = parse_query(uri)
     except ValueError as e:
-        # TODO: Log this error
+        logging.error(f"S3 Find Error {ve}")
         raise e
 
     s3 = S3Fetcher()
@@ -357,16 +359,12 @@ def s3_uri_to_stream(uri: str, skip_check: bool):
         else:
             stream = do_dir_query(qq)
 
-    yaml_list = []
-
     try:
         for i, o in enumerate(stream):
-            yaml_list.append(o.url)
+            yield o.url
     except Exception as e:
-        # TODO: Log this error
+        logging.error(f"S3 Find Error {e}")
         raise e
-
-    return yaml_list
 
 class S3Fetcher(object):
     def __init__(self,
--- a/libs/aio/odc/aio/__init__.py
+++ b/libs/aio/odc/aio/__init__.py
@@ -293,6 +293,17 @@ async def s3_walker(url, nconcurrent, s3,
 
 
 def s3_uri_to_stream(uri: str, skip_check: bool):
+    """Build generator from supplied S3 URI pattern
+    Arguments:
+        uri {str} -- S3 URI to create generator for
+        skip_check {bool} -- Validity check flag
+    Raises:
+        e: ValueError if parsing URI fails
+    Returns:
+        [type] -- [description]
+    Yields:
+        [type] -- [description]
+    """
     def do_file_query(qq, pred):
         for d in s3.dir_dir(qq.base, qq.depth):
             _, _files = s3.list_dir(d).result()
@@ -317,8 +328,6 @@ def s3_uri_to_stream(uri: str, skip_check: bool):
 
     def do_dir_query(qq):
         return (SimpleNamespace(url=url) for url in s3.dir_dir(qq.base, qq.depth))
-
-    flush_freq = 100
 
     try:
         qq = parse_query(uri)
@@ -348,7 +357,16 @@ def s3_uri_to_stream(uri: str, skip_check: bool):
         else:
             stream = do_dir_query(qq)
 
-    return stream
+    yaml_list = []
+
+    try:
+        for i, o in enumerate(stream):
+            yaml_list.append(o.url)
+    except Exception as e:
+        # TODO: Log this error
+        raise e
+
+    return yaml_list
 
 class S3Fetcher(object):
     def __init__(self,
--- a/libs/aio/odc/aio/__init__.py
+++ b/libs/aio/odc/aio/__init__.py
@@ -294,17 +294,13 @@ async def s3_walker(url, nconcurrent, s3,
     return step
 
 
-def s3_find_uri(uri: str, skip_check: bool):
-    """Build generator from supplied S3 URI pattern
+def s3_find_glob(glob_pattern: str, skip_check: bool):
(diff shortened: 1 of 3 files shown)


libs/aio/odc/aio/__init__.py
```diff
@@ -292,6 +292,82 @@ def default_guide(url, depth, base):
...
+        elif qq.file is not None:
+            stream = do_file_query2(qq)
+        else:
+            stream = do_dir_query(qq)
+
+    yaml_list = []
+
+    try:
+        for i, o in enumerate(stream):
+            yaml_list.append(o.url)
```
Kirill888, line 364: I think you don't want to return list here, instead of adding things to List just `yield o.url`
But do consider error handling as well

Kirill888: Also doesn't looks like you use `i` so probably don't need `enumerate`

libs/aio/odc/aio/__init__.py
```diff
@@ -292,6 +292,82 @@ def default_guide(url, depth, base):
     return step
 
 
+def s3_uri_to_stream(uri: str, skip_check: bool):
```
Kirill888, line 295: I think `s3_uri_to_stream` is a confusing name, I would assume that it streams content of some large file as bytes/byte-chunks or something like that, from name alone

whatnick: `s3_uri_to_generator` is a better name ? Naming stuff is hard.

Kirill888: it's not `uri`, it's `glob`, and what is `stream`?

libs/aio/odc/aio/__init__.py
whatnick: The library does not use a logger infrastructure. Do you have a design in mind or should we just do logging.warning at the top level logger.

Kirill888:
As far as async goes, it's not meant to be async at the interface, internally it is, but externally it's just an iterator.

whatnick:
Still need to resolve this :
```
  File "/home/ubuntu/odc-tools/libs/aio/odc/aio/__init__.py", line 399, in _close
    await s3.close()
AttributeError: 'ClientCreatorContext' object has no attribute 'close'
```

libs/aio/odc/aio/__init__.py
Kirill888: For logging we just use python logging elsewhere. So I guess same here. But logging is separate from error handling, in this case you are just returning a lazy sequence of URIs from a single glob pattern, so there isn't much to do error handling wise. I mean underlying layers should be doing it properly though.

Next step where you actually fetch data will need it though.

Kirill888:
Maybe it should be wrapped in `with` what version of `aiobotocore` are you testing with? Things might have changed since when this was written

whatnick:
> Maybe it should be wrapped in `with` what version of `aiobotocore` are you testing with? Things might have changed since when this was written

I am on `aiobotocore==1.0.0` . Use context manager on S3Fetcher ?

whatnick:
@Kirill888 this is a breaking API change for [aiobotocore==1.0.0](https://github.com/aio-libs/aiobotocore/releases/tag/1.0.0) , works with 0.12.0 , is worth pinning to that version in deployments. Will create a separate issue and fix for breaking API change.

libs/aio/odc/aio/__init__.py
```diff
@@ -292,6 +294,78 @@ def default_guide(url, depth, base):
...
+        if qq.glob is not None:
+            pred = norm_predicate(glob=qq.glob)
+            stream = do_file_query(qq, pred)
+        elif qq.file is not None:
+            stream = do_file_query2(qq)
+        else:
+            stream = do_dir_query(qq)
+
+    try:
+        for i, o in enumerate(stream):
```
Kirill888, line 363: remove enumerate if you don't use it, in fact what is the point of this loop, you can just return iterator like this:
```python
return (o.url for o in stream)
```

In fact why do we need to turn into strings? Why not return a richer object, what if you want to check timestamp or size, turning that to a string is trivial afterwards

libs/aio/odc/aio/__init__.py
```diff
@@ -292,6 +294,78 @@ def default_guide(url, depth, base):
...
 
+def s3_find_uri(uri: str, skip_check: bool):
+    """Build generator from supplied S3 URI pattern
+    Arguments:
+        uri {str} -- S3 URI to create generator for
+        skip_check {bool} -- Validity check flag
+    Raises:
+        e: ValueError if parsing URI fails
+    Returns:
+        [type] -- [description]
```
Kirill888, line 305: please fill those.
Let's change output to a stream of objects rather than just strings, whatever that comes out of find.

libs/aio/odc/aio/__init__.py
Kirill888: Actually I'm not sure, what else is in this there except for `.url`?

libs/aio/odc/aio/__init__.py
```diff
@@ -292,6 +294,78 @@ def default_guide(url, depth, base):
...
+            stream = do_file_query2(qq)
+        else:
+            stream = do_dir_query(qq)
+
+    try:
+        for i, o in enumerate(stream):
+            yield o
+    except Exception as e:
+        logging.error(f"S3 Find Error {e}")
+        raise e
```
Kirill888, line 367: Just `return stream` instead this whole block. No need to create extra two depths of iteration just for the sake of logging.

Kirill888: In fact even better: don't assign stream at all, and just return instead of assignment.

libs/aio/odc/aio/__init__.py
whatnick: `s3_glob_to_keys` is a better name ?

PR #40, opened 2020-04-15
Move eo3 related code from app to lib

whatnick:
Enable better re-use of YAML post-processing code form eo3 metadata by moving it to library. Also add a proposed eo3 auto-detect block to be implemented and addressed in separate PR, with relevant change management for deprecation of `--eo3` CLI flag.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
+++ b/apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
@@ -5,43 +5,11 @@ from datacube.utils.changes import allow_any
 from odc.io.tar import tar_doc_stream, tar_mode
 from odc.io.timer import RateEstimator
 from odc.index import from_yaml_doc_stream
-from odc.index import eo3_grid_spatial
-
-# TODO: Move to library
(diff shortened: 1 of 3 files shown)


apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -143,6 +123,7 @@ def process_file(filename, index, fps, mode=None, n_failed=0, doc_transform=None
 
         return n_failed
 
+    # TODO: Release SQLAlchemy with context manager
```
Kirill888, line 126: release when? It's used until the last moment of this function and then deleted, so what's the point?

libs/index/odc/index/_eo3.py
```diff
@@ -102,3 +102,36 @@ def eo3_grid_spatial(doc, tol=None):
         lon=dict(begin=bb.left, end=bb.right))
 
     return oo
+
+def add_eo3_parts(doc, tol=None):
```
Kirill888, line 106: Please add doc comments, now that it's in the library

PR #44, opened 2020-04-20
Move thredds crawler YAML download code to library

whatnick:
Move code to library allowing future refactor to aiohttp async download of large collection of YAMLS.

### DIFF
--- a/libs/thredds/odc/thredds/__init__.py
+++ b/libs/thredds/odc/thredds/__init__.py
@@ -9,8 +9,8 @@ from typing import List, Tuple, Optional
 
 
 def thredds_find_glob(
-        base_catalog: str, skips: list, select: list, workers: int = 8
-    ) -> list:
+        base_catalog: str, skips: List[str], select: List[str], workers: int = 8
+    ) -> List[str]:
     """Glob YAML's from base Thredds Catalog recursively
     Arguments:
         base_catalog {str} -- Base of the catlog to crawl from
@@ -18,7 +18,7 @@ def thredds_find_glob(
         select {list} -- Paths to select (useful YAML's)
         workers {int} -- Number of workers to use for Thredds Crawling
     Returns:
-        list -- List of datasets to Index
+        list -- List of Thredds hosted dataset YAML url's to Index
     """
     user_skips = Crawl.SKIPS
     user_skips = user_skips.extend(skips)
--- a/libs/thredds/odc/thredds/__init__.py
+++ b/libs/thredds/odc/thredds/__init__.py
@@ -3,6 +3,7 @@
 from thredds_crawler.crawl import Crawl
 import requests
 from urllib.parse import urlparse
+from multiprocessing.dummy import Pool as ThreadPool
 
 
 def thredds_find_glob(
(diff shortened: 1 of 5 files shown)


libs/thredds/odc/thredds/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    
+    Returns:
+        list -- List of datasets to Index
+    """
+    user_skips = Crawl.SKIPS
+    user_skips = user_skips.extend(skips)
+
+    results = Crawl(
+        base_catalog + "/catalog.xml", select=select, skip=user_skips, workers=workers
+    ).datasets
```
Kirill888, line 28: Is `results` a list of object or a stream? If its is a list then fine, but if it can be "lazy" then we should also return generator and not a list.

libs/thredds/odc/thredds/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+        service["url"]
+        for dataset in results
+        for service in dataset.services
+        if service["service"].lower() == "httpserver"
+    ]
+
+    return urls
+
+
+def download_yamls(yaml_urls: list, workers : int = 8) -> list:
```
Kirill888, line 40: `: list` is not a good mypy annotation, should be `: List[str]` and `-> List[WhatEverThisReturns]`

libs/thredds/odc/thredds/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    # use a threadpool to download from thredds
+    pool = ThreadPool(workers)
+    yamls = pool.map(_download, yaml_urls)
+    pool.close()
+    pool.join()
+
+    return yamls
+
+def _download(url: str) -> tuple:
```
Kirill888, line 58: mypy again, ~-> tuple~ `-> Tuple[Optional[bytes], str, Optional[str]`

libs/thredds/odc/thredds/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+        tuple -- URL content, target file and placeholder for error
+    """
+    parsed_uri = urlparse(url)
+    target_filename = url[len(parsed_uri.scheme + "://") :]
+    try:
+        resp = requests.get(url)
+        if resp.status_code == 200:
+            return (resp.content, target_filename, None)
+        else:
+            return(None, None, "Yaml not found")
```
Kirill888, line 74: Should always return `target_filename` for error logging,

libs/thredds/odc/thredds/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    """Glob YAML's from base Thredds Catalog recursively
+    
+    Arguments:
+        base_catalog {str} -- Base of the catlog to crawl from
+        user_skips {list} -- Paths to skip in addition to NCI specific defaults
+        select {list} -- Paths to select (useful YAML's)
+        workers {int} -- Number of workers to use for Thredds Crawling
+    
+    Returns:
+        list -- List of datasets to Index
```
Kirill888, line 21: needs more info about what exactly is a "dataset" in this context

libs/thredds/odc/thredds/__init__.py
whatnick: It is a list , I would like to make is lazy and return iterators in the future. This is a refactor than a rewrite. Internally crawl uses multiprocessing.

libs/thredds/odc/thredds/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+    Arguments:
+        url {str} -- URL on thredds to download YAML for
+    Returns:
+        tuple -- URL content, target file and placeholder for error
+    """
+    parsed_uri = urlparse(url)
+    target_filename = url[len(parsed_uri.scheme + "://") :]
+    try:
+        resp = requests.get(url)
+        if resp.status_code == 200:
```
Kirill888, line 70: `200` is not the only success code.

Issue #41, opened 2020-04-15
Include eo3 auto-detection


whatnick:
Move eo3 from CLI flag which is often missed by users to internal-heuristic which examines certain signature keys in YAML being indexed to auto-apply eo3 related post-processing.

Also retain current flag and issue deprecation warnings for backward compatibility.

whatnick:
return 'grids' in doc and 'default' in doc['grids'] and 'transform' in doc['grids']['default']

Kirill888:
no longer relevant, code moved to datacube core, and does include auto-detection

PR #45, opened 2020-04-22
#41 Implemement 1 liner for eo3 detection

whatnick:


libs/index/odc/index/_eo3.py
```diff
@@ -138,4 +138,4 @@ def detect_eo3(doc : dict) -> bool:
     Returns:
         bool -- is this metadata eo3 ?
     """
-    return False
\ No newline at end of file
+    return ('grids' in doc) and ('default' in doc['grids']) and ('transform' in doc['grids']['default'])
```
uchchwhash, line 141: Maybe a newline would be nicer? :smirk:

PR #46, opened 2020-04-29
Read AWS region from AWS_REGION environment variable.

SpacemanPaul:


SpacemanPaul:
Oh damn, that's right.

I was using AWS_REGION and DEFAULT_AWS_REGION.  Whoops.

PR #48, opened 2020-05-21
First pass at STAC indexing

alexgleith:
Reasonably generic function that will index STAC 1.0 metadata.

It does have a specific Sentinel-2 pathway, but should be able to accommodate other platforms ok. There's an easy way to do it too.

Still needs:
- More testing
- New product definition to test on
- A way to pass in the default grid, maybe it should be part of the platform-specific section

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
+++ b/apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
@@ -79,21 +79,17 @@ def _get_stac_bands(item, default_grid='g10.0m'):
 
 
 def geographic_to_projected(geometry, target_srs):
-    wgs = CRS.from_epsg(4326)
-    projection = CRS.from_epsg(target_srs)
+    """ Transform from WGS84 to the target projection, assuming Lon, Lat order
+    """
+    transformer = Transformer.from_crs(4326, target_srs, always_xy=True)
+    geometry['coordinates'][0] = list(transformer.itransform(
+        geometry['coordinates'][0]))
 
-    transformer = Transformer.from_crs(wgs, projection)
-
-
-    new_geometry = deepcopy(geometry)
-    new_geometry['coordinates'][0] = [
-        transformer.transform(lon, lat) for lon, lat in new_geometry['coordinates'][0]]
-
-    return new_geometry
+    return geometry
 
 
 def stac_transform(input_stac):
-    """takes in a raw STAC 1.0 dictionary and returns an ODC dictionary
+    """ Takes in a raw STAC 1.0 dictionary and returns an ODC dictionary
     """
 
     product_label, product_name, region_code = _stac_product_lookup(input_stac)
@@ -107,14 +103,14 @@ def stac_transform(input_stac):
     stac_odc = {
         '$schema': 'https://schemas.opendatacube.org/dataset',
         'id': deterministic_uuid,
-        'crs': "epsg:{}".format(properties['proj:epsg']),
+        'crs': f"epsg:{properties['proj:epsg']}",
         'geometry': geographic_to_projected(
             input_stac['geometry'],
             properties['proj:epsg']
         ),
         'grids': grids,
         'product': {
-            'name': product_name
+            'name': product_name.lower()
         },
         'label': product_label,
         'properties': {
--- a/apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
+++ b/apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
@@ -1,13 +1,143 @@
 """ Index datasets from tar arachive
 """
 
+import os
 import sys
+from copy import deepcopy
+
 import click
 import datacube
 from datacube.utils.changes import allow_any
+from odc.index import from_yaml_doc_stream, odc_uuid
 from odc.io.tar import tar_doc_stream, tar_mode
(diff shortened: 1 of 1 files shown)


apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -1,13 +1,143 @@
...
+def _stac_product_lookup(item):
+    product_label = item['id']
+    product_name = item['properties']['platform']
+    region_code = None
+
+    constellation = item['properties']['constellation']
+
+    if constellation in KNOWN_CONSTELLATIONS:
+        if constellation == 'sentinel-2':
+            product_label = item['properties']["sentinel:product_id"]
```
Kirill888, line 31: I usually prefer using `toolz.get_in` for nested access (`https://kite.com/python/docs/toolz.get_in`), since we do not do any schema validation we can't guarantee presence of various fields.

Mixing `'` and `"` on the same line is also not ideal.

apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -1,13 +1,143 @@
...
+
+        bands[asset_name] = band_info
+
+    grids['default'] = grids[default_grid]
+    del grids[default_grid]
+
+    return bands, grids
+
+
+def geographic_to_projected(geometry, target_srs):
```
Kirill888, line 79: Is this funtion in use? Which GDAL this is supposed to work with? lon/lat order will be wrong for one of them

apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -1,13 +1,143 @@
...
+    # Make a proper deterministic UUID
+    deterministic_uuid = str(odc_uuid("sentinel2_stac_process", "1.0.0", [product_label]))
+
+    # Get grids and bands
+    bands, grids = _get_stac_bands(input_stac)
+
+    stac_odc = {
+        '$schema': 'https://schemas.opendatacube.org/dataset',
+        'id': deterministic_uuid,
+        'crs': "epsg:{}".format(input_stac['properties']['proj:epsg']),
```
Kirill888, line 109: there are so many `input_stac['properties']` in here, I would

```python
props = input_stac['properties']
...
use(props['proj:epsg'])
```

This will read better, be safer and more performant too.

apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -51,6 +181,8 @@ def untar(tarfname, mk_uri):
 @click.option('--xz', is_flag=True, help='Input is compressed with xz (needed when reading from stdin)')
 @click.option('--protocol', type=str, default='s3', show_default=True,
               help='Override the protocol for working with data in other environments, i.e gs')
+@click.option('--stac', is_flag=True, default=False, show_default=True,
+              help='Convert STAC 1.0 metadata to ODC EO3 metadata')
```
Kirill888, line 185: Maybe re-word this a bit? "Expect STAC 1.0 metadata on input" or something along these lines.

apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -1,13 +1,135 @@
...
+    del grids[default_grid]
+
+    return bands, grids
+
+
+def geographic_to_projected(geometry, target_srs):
+    """ Transform from WGS84 to the target projection, assuming Lon, Lat order
+    """
+    transformer = Transformer.from_crs(4326, target_srs, always_xy=True)
+    geometry['coordinates'][0] = list(transformer.itransform(
```
Kirill888, line 85: should't this be done for every polygon ring, not just first?

apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -1,13 +1,135 @@
...
+    return geometry
+
+
+def stac_transform(input_stac):
+    """ Takes in a raw STAC 1.0 dictionary and returns an ODC dictionary
+    """
+
+    product_label, product_name, region_code = _stac_product_lookup(input_stac)
+
+    deterministic_uuid = str(odc_uuid("sentinel-2_stac_process", "1.0.0", [product_label]))
```
Kirill888, line 97: so `product_label` is sufficiently unique identifier for a dataset?

apps/dc_tools/odc/apps/dc_tools/index_from_tar.py
```diff
@@ -1,13 +1,135 @@
...
+            'name': product_name.lower()
+        },
+        'label': product_label,
+        'properties': {
+            'datetime': properties['datetime'].replace("000+00:00", "Z"),
+            'odc:processing_datetime': properties['datetime'].replace("000+00:00", "Z"),
+            'eo:cloud_cover': properties['eo:cloud_cover'],
+            'eo:gsd': properties['gsd'],
+            'eo:instrument': properties['instruments'][0],
+            'eo:platform': properties['platform'],
```
Kirill888, line 122: how come we prefix `eo:` for `gsd` and `platform`, has this been moved from extension to stac-proper? if yes should we update eo3 metadata then?

PR #55, opened 2020-06-15
Dask reproject

Kirill888:
MEM -> MEM reproject that should work on bigger than available RAM Dask graphs.

Accepts `xr.Data{Array|Set}` and `GeoBox` on input and produces `xr.Data{Array|Set)` on output.
Time dimension is supported, but RGBA inputs are not yet supported for the `reproject` some of the lower-level components are fully N-dimensional `crop_2d_dense`.

When working with Dask inputs the dependency graph is computed to be "minimal", i.e. only those blocks that are needed to compute output block are declared as a dependency see (`crop_2d_dense`). 

Future Improvements not included here

- Support RGBA inputs
- Share GeoSpatial computation across bands (might be handy for really large, in the number of blocks, inputs)

The assumption is that reproject doesn't perform a huge reduction is scale, as this might require to fully realize (load into memory) too big of a window of the input for one single output block.

### DIFF
--- a/libs/algo/odc/algo/__init__.py
+++ b/libs/algo/odc/algo/__init__.py
@@ -33,6 +33,10 @@ from ._rgba import (
     to_rgba_np,
 )
 
+from ._warp import (
+    dask_reproject
+)
+
 __all__ = (
     "keep_good_np",
     "keep_good_only",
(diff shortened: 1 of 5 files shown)


uchchwhash (approved):
Not sure if I understood everything, but 👍

PR #56, opened 2020-06-16
add custom gridspec to dscache

uchchwhash:


libs/dscache/odc/dscache/apps/dstiler.py
```diff
@@ -14,8 +14,12 @@
 @click.option('--native', is_flag=True, help='Use Landsat Path/Row as grouping')
 @click.option('--native-albers', is_flag=True, help='When datasets are in Albers grid already')
 @click.option('--web', type=int, help='Use web map tiling regime at supplied zoom level')
+@click.option('--crs', type=str, help="Custom gridspec: crs")
+@click.option('--resolution', type=str, help="Custom gridspec: resolution_y,resolution_x")
+@click.option('--shape', type=str, help="Custom gridspec: shape_y,shape_x")
```
Kirill888: Can we change help string to just say `Y,X` rather than `{resolution|shape}_{y,x}`, I feel like that reads clearer than repeating same word three times on one line of text.

"Custom gridspec: resolution in CRS units per pixel in y,x order"
"Custom gridspec: shape of tile in pixels ny,nx"

libs/dscache/odc/dscache/apps/dstiler.py
```diff
@@ -36,6 +40,17 @@ def cli(native, native_albers, web, dbfile):
         gs = web_gs(web)
         group_key_fmt = 'web_' + str(web) + '/{:03d}_{:03d}'
         binner = lambda dss: bin_dataset_stream(gs, dss)
+    elif crs is not None or resolution is not None or shape is not None or bin_format is not None:
```
Kirill888, line 42: this condition looks odd, but I guess it's right, I guess `any(x is not None for x in [crs, resolution, shape, bin_format])` is not that much better

libs/dscache/odc/dscache/apps/dstiler.py
```diff
@@ -36,6 +40,17 @@ def cli(native, native_albers, web, dbfile):
...
+        settings = dict(crs=crs, resolution=resolution, shape=shape, bin_format=bin_format)
+        missing = [key for key, value in settings.items() if value is None]
+        if missing:
+            raise ValueError('Missing options in custom grid spec: {}'.format(' '.join(missing)))
+        group_key_fmt = bin_format
+        resolution = [float(comp) for comp in resolution.split(',')]
+        shape = [float(comp) for comp in shape.split(',')]
+        tile_size = [abs(res * shp) for res, shp in zip(resolution, shape)]
+        gs = GridSpec(crs=geom.CRS(crs), tile_size=tile_size, resolution=resolution)
```
Kirill888: most things now accept crs as a string on input, if `GridSpec` doesn't we should fix that in core

Kirill888 (approved):

PR #60, opened 2020-06-22
Fix grid naming for STAC grids

alexgleith:
Removes the .0 from grid names, chancing them from things like g10.0m to g10m.

### DIFF
--- a/libs/index/odc/index/stac.py
+++ b/libs/index/odc/index/stac.py
@@ -115,6 +115,6 @@ def stac_transform(input_stac):
     }
 
     if region_code:
-        stac_odc['properties']['odc:region_code']: region_code
+        stac_odc['properties']['odc:region_code'] = region_code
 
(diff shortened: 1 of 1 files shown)


libs/index/odc/index/stac.py
```diff
@@ -45,7 +45,7 @@ def _get_stac_bands(item, default_grid='g10.0m'):
             continue
 
         transform = asset['proj:transform']
-        grid = f'g{transform[0]}m'
+        grid = f'res:g{transform[0]}m'
```
Kirill888: What I meant on Slack was `f"{whatever_variable_name_you_use_for_resolution:g}` ,so in this case 
```python
grid = f'{transform[0]:g}m'
```
so rather than converting to a string in a default fashion convert with "human readable floating point that might be an int" policy.

Kirill888 (approved):

PR #65, opened 2020-06-24
add binning to slurpy

uchchwhash:
This seems to work.

The performance did not improve much, but at least the API is there now. I probably am converting datasets to dicts back and forth more than I need to.

The true bottleneck is `gridspec.tiles_from_polygon` according to the profiler.

Q: for named gridspecs should we use the name as `group_prefix`?


Kirill888 (approved):
excellent, thanks!

PR #69, opened 2020-07-02
Migrate to newer `aiobotocore` and add support for unsigned and requestor pays

### LINKED ISSUE #38: Breaking API change for [aiobotocore==1.0.0](https://github.com/aio-libs/aiobotocore/releases/tag/1.0.0) , works with 0.12.0
@Kirill888 this is a breaking API change for [aiobotocore==1.0.0](https://github.com/aio-libs/aiobotocore/releases/tag/1.0.0) , works with 0.12.0 , is worth pinning to that version in deployments. Will create a separate issue and fix for breaking API change.

_Originally posted by @whatnick in https://github.com/opendatacube/odc-tools/pull/37#issuecomment-607033572_

Kirill888:
1. Bring code changes from `datacube-core` back into `odc-tools` (this includes support for unsigned listing)
2. Support newer `aibotocore` lib
3. Add support for unsigned requests to `odc.aio`
4. Add options to pass through arbitrary boto keywords into various s3 helper methods (needed for Request Payer)
5. Add CLI options `--no-sign-request` and `--request-payer` to `s3-find`, `s3-to-tar` and `s3-inventory-dump` tools, option names are copied from `aws` cli app.

- [x] Closes #38

### DIFF
--- a/apps/cloud/odc/apps/cloud/s3_find.py
+++ b/apps/cloud/odc/apps/cloud/s3_find.py
@@ -1,13 +1,15 @@
 import click
 import sys
-from odc.aio import s3_find_glob
+from odc.aio import s3_find_glob, S3Fetcher
 
 
 @click.command('s3-find')
 @click.option('--skip-check', is_flag=True,
               help='Assume file exists when listing exact file rather than wildcard.')
+@click.option('--aws-unsigned', is_flag=True,
+              help='Do not sign AWS S3 requests')
 @click.argument('uri', type=str, nargs=1)
-def cli(uri, skip_check):
+def cli(uri, skip_check, aws_unsigned=None):
     """ List files on S3 bucket.
 
     Example:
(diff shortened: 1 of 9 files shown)


apps/cloud/odc/apps/cloud/s3_find.py
```diff
@@ -1,13 +1,17 @@
...
-from odc.aio import s3_find_glob
+from odc.aio import s3_find_glob, S3Fetcher
 
 
 @click.command('s3-find')
 @click.option('--skip-check', is_flag=True,
               help='Assume file exists when listing exact file rather than wildcard.')
+@click.option('--no-sign-request', is_flag=True,
+              help='Do not sign AWS S3 requests')
+@click.option('--request-payer', is_flag=True,
```
alexgleith, line 11: The AWS CLI uses `--request-payer requester` as the flag, which sucks. This way is better. Just saying ;-)

alexgleith (approved):
Looks super simple. Nice work!

apps/cloud/odc/apps/cloud/s3_find.py
Kirill888: I guess that makes sense, oh well let's just leave it as is, unless you can suggest a better name to not be "almost like aws but not really". Maybe `--requestor-pays`? But then it's too similar but different.

Kirill888: I guess with the current option name you can turn it into sentence as "this command is `--request-payer`"

PR #73, opened 2020-07-27
Tools to faciliate metadata generation for stats

uchchwhash:
- Add dataclasses representing stats `OutputProduct` and `Task`
- `OutputProduct` contains roughly the common metadata across a product, a `Task` contains specifics for a particular task to run
- Add a function `task_to_metadata_dict` that converts a task to a metadata document dictionary in the `eo3` format
- A tentative `clear_pixel_count_product` output product capturing information that could be from some external source
- The `odc.stats.metadata` module is currently a standalone script that enhances a dataset cache with metadata docs for this product over 2020
- It really should not do that, it should save the task objects there, but it is easier to inspect the end result this way for now
- The primary purpose of this standalone script currently is to demonstrate usage pattern for this tool

### DIFF
--- a/libs/stats/odc/stats/model.py
+++ b/libs/stats/odc/stats/model.py
@@ -1,0 +1,30 @@
+from typing import Dict, Tuple, Sequence
+from uuid import UUID
+from datetime import datetime
+
+from dataclasses import dataclass
+
+from datacube.model import GridSpec
+
+
+@dataclass
+class OutputProduct:
+    product_name: str
+    product_version: str
+    product_info: Dict[str, str]
+    product_short_name: str
+    algo_info: Dict[str, str]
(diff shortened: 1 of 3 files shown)


libs/stats/odc/stats/model.py
```diff
@@ -0,0 +1,30 @@
...
+
+from datacube.model import GridSpec
+
+
+@dataclass
+class OutputProduct:
+    product_name: str
+    product_version: str
+    product_info: Dict[str, str]
+    product_short_name: str
```
Kirill888, line 15: Class name is `Product` no need to prefix with `product_`

Kirill888 (approved):
I will merge and make changes in develop.

PR #79, opened 2020-08-19
Refactor mask loading into a generic load with native transform

Kirill888:
Adding `odc.algo.io.load_with_native_transform` method, and then re-implementing `odc.algo.io.load_enum_mask` in terms of `load_with_native_transform`.

Also adds `odc.algo.choose_first_valid` which is essentially datacube default fuser  as a Dask aware operation.


uchchwhash (approved):
Looks reasonable. :+1:

PR #80, opened 2020-08-20
Fixes for making transform compatible to any STAC doc

sachitrajbhandari:


libs/index/odc/index/stac.py
```diff
@@ -86,9 +150,7 @@ def stac_transform(input_stac):
     """ Takes in a raw STAC 1.0 dictionary and returns an ODC dictionary
     """
 
-    product_label, product_name, region_code = _stac_product_lookup(input_stac)
-
-    deterministic_uuid = str(odc_uuid("sentinel-2_stac_process", "1.0.0", [product_label]))
+    product_id, product_label, product_name, region_code = _stac_product_lookup(input_stac)
```
Kirill888: Looks like `product_id` contains UUID of a dataset, it's not a good name for that.

libs/index/odc/index/stac.py
```diff
@@ -29,8 +33,23 @@ def _stac_product_lookup(item):
                 properties['sentinel:latitude_band'],
                 properties['sentinel:grid_square']
             )
+            product_id = str(odc_uuid("sentinel-2_stac_process", "1.0.0", [product_label]))
+    elif properties.get('platform') in LANDSAT_PLATFORMS:
+        product_id = item['id']
+        product_label = _product_label(item)
+        product_name = properties.get('odc:product')
+        region_code = properties.get('odc:region_code')
```
Kirill888, line 57: Needs an `else:` with reasonable something in it.

libs/index/odc/index/stac.py
```diff
@@ -100,7 +162,7 @@ def stac_transform(input_stac):
 
     stac_odc = {
         '$schema': 'https://schemas.opendatacube.org/dataset',
-        'id': deterministic_uuid,
+        'id': product_id,
```
Kirill888: If my interpretation is wrong and `product_id` is not a UUID of a Dataset then it should not be here.

libs/index/odc/index/stac.py
sachitrajbhandari: Added `else:` statement

libs/index/odc/index/stac.py
sachitrajbhandari: Removed ambiguous use of `product_id`

libs/index/odc/index/stac.py
sachitrajbhandari: Replaced variable name `product_id` with `deterministic_uuid `

libs/index/odc/index/stac.py
```diff
@@ -86,11 +108,16 @@ def stac_transform(input_stac):
...
-    product_label, product_name, region_code = _stac_product_lookup(input_stac)
+    product_label, product_name, region_code, default_grid = _stac_product_lookup(input_stac)
 
-    deterministic_uuid = str(odc_uuid("sentinel-2_stac_process", "1.0.0", [product_label]))
+    # Generating UUID for products not having UUID.
+    # TODO: Check is based on hardcoded Product Name, find generic way
+    if product_name in ["s2_l2a"]:
+        deterministic_uuid = str(odc_uuid("sentinel-2_stac_process", "1.0.0", [product_label]))
+    else:
+        deterministic_uuid = input_stac["id"]
```
Kirill888, line 199: is `id` keys guaranteed by STAC api?
what happens if it's not there?

libs/index/odc/index/stac.py
```diff
@@ -29,42 +29,64 @@ def _stac_product_lookup(item):
...
             )
+            default_grid = "g10m"
+    elif properties.get('platform') in LANDSAT_PLATFORMS:
+        product_label = _product_label(item)
+        product_name = properties.get('odc:product')
+        region_code = properties.get('odc:region_code')
+        default_grid = "g30m"
+    else:
+        product_label = item['id']
+        product_name = properties['platform']
```
Kirill888, line 61: are these items guaranteed to be present?

Kirill888:
I think this needs tests, This code does mostly dict->dict to transform, no IO no DB so tests should be easy to write.

sachitrajbhandari:
> I think this needs tests, This code does mostly dict->dict to transform, no IO no DB so tests should be easy to write.

Ok adding tests for transforms

libs/index/odc/index/stac.py
sachitrajbhandari: The `id` is REQUIRED field and will exists in any valid STAC doc.
https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields

Kirill888: ok, but it's not UUID, code elsewhere assumes UUID. Need to generate deterministic UUID somehow.

libs/index/odc/index/stac.py
```diff
@@ -86,11 +108,16 @@ def stac_transform(input_stac):
...
     """
 
-    product_label, product_name, region_code = _stac_product_lookup(input_stac)
+    product_label, product_name, region_code, default_grid = _stac_product_lookup(input_stac)
 
-    deterministic_uuid = str(odc_uuid("sentinel-2_stac_process", "1.0.0", [product_label]))
+    # Generating UUID for products not having UUID.
+    # TODO: Check is based on hardcoded Product Name, find generic way
+    if product_name in ["s2_l2a"]:
+        deterministic_uuid = str(odc_uuid("sentinel-2_stac_process", "1.0.0", [product_label]))
```
Kirill888: how does product_label looks like? Is it unique per dataset, `deterministic_uuid` needs to be unique and product_label is the only non-const parameter.

sachitrajbhandari: Yes `product_label` is unique for `product_name="s2_l2a"`

libs/index/odc/index/stac.py
sachitrajbhandari: The `id` guaranteed to be present as it is REQUIRED field but `platform` is not REQUIRED field
https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields

alexgleith (approved):

libs/index/odc/index/stac.py
```diff
@@ -9,16 +11,35 @@
     'sentinel-2'
 ]
 
+LANDSAT_PLATFORMS = [
+    'landsat-5', 'landsat-7', 'landsat-8'
+]
 
-def _stac_product_lookup(item):
+# Mapping between EO3 field names and STAC properties object field names
+MAPPING_STAC_TO_EO3 = {
```
Kirill888, line 19: @alexgleith So what's up with these mappings. Have there been update to STAC and those fields are now standard? Cause then we should probably update eo3 instead. This will be fine for now, but not in a long term.

Kirill888 (approved):
I'm still not too sure about uuid generation, but we can iterate later

PR #83, opened 2020-08-28
Added support for temporal slicing of datasets

tebadi:


libs/stats/odc/stats/model.py
```diff
@@ -41,17 +42,20 @@ def __init__(self, start: Union[str, datetime],
         DateTimeRange(datetime(2019, 3, 1), '3M')
 
         """
+        anchor = None
```
Kirill888: I don't understand this anchor thing.

anchor is only meaningful when you don't have an exact period, where you say something like "starting from the first period for which data exists give me 3month periods starting in March". Then anchor would be "March".

this class encodes exact period, so no anchor. anchor is a wild-card for generating periods not a property of a period.

tebadi: I had used the DateTimeRange class to accommodate anchor.  I've now separated 'period' from 'tempoal_range'.  Note that temporal_range is equivalent of the period in the initial implementation.

libs/stats/odc/stats/model.py
```diff
@@ -41,17 +42,21 @@ def __init__(self, start: Union[str, datetime],
         DateTimeRange(datetime(2019, 3, 1), '3M')
 
         """
+
+        if isinstance(start, str) and start.startswith("%Y"):
```
Kirill888: this does not belong here, constructor should only accept valid time range in ISO format.

libs/stats/odc/stats/cli.py
```diff
@@ -224,7 +283,7 @@ def sanitize(v):
     csv_path = out_path(".csv")
     print(f"Writing summary to {csv_path}")
     with open(csv_path, 'wt') as f:
-        f.write('"Period","X","Y","datasets","days"\n')
+        f.write('"temporal_range","X","Y","datasets","days"\n')
```
Kirill888: how about just `T` here.

libs/stats/odc/stats/cli.py
```diff
@@ -3,9 +3,12 @@
 from tqdm.auto import tqdm
 import sys
 import pickle
+import pandas as pd
+from typing import Dict, Tuple, List
 from datetime import datetime
 from collections import namedtuple
 from odc.io.text import click_range2d
+from datacube.utils.dates import normalise_dt
```
Kirill888: can you move this into function, `datacube` imports are slow and really slow down cli, same for pandas

libs/stats/odc/stats/cli.py
```diff
@@ -30,10 +32,13 @@ def is_tile_in(tidx, tiles):
...
               type=int,
-              help="Only extract datasets for a given year")
+              help="Only extract datasets for a given year. This is a shortcut for --temporal-range=<int>--P1Y")
+@click.option('--temporal_range',
+              type=str,
+              help="Only extract datasets for a given time temporal_range, Example '2020-05--P1M' month of May 2020")
 @click.option('--period',
               type=str,
-              help="Only extract datasets for a given time period, Example '2020-05--P1M' month of May 2020")
+              help="Extract datasets for a given period specified, Example 'all', 'annual', '2020-03--P2M' ")
```
Kirill888: so how does seasonal thing works?
Also is `period` a good name for this, maybe "frequency"?

libs/stats/odc/stats/cli.py
```diff
@@ -30,10 +32,13 @@ def is_tile_in(tidx, tiles):
               default=None)
 @click.option('--year',
               type=int,
-              help="Only extract datasets for a given year")
+              help="Only extract datasets for a given year. This is a shortcut for --temporal-range=<int>--P1Y")
+@click.option('--temporal_range',
+              type=str,
+              help="Only extract datasets for a given time temporal_range, Example '2020-05--P1M' month of May 2020")
```
Kirill888: "time temporal_range"?

libs/stats/odc/stats/cli.py
```diff
@@ -82,26 +87,86 @@ def save_tasks(grid, year, period,
...
 
     if output == '':
-        if period is not None:
-            output = f'{product}_{period.short}.db'
+        if temporal_range is not None:
+            output = f'{product}_{temporal_range.short}.db'
         else:
             output = f'{product}_all.db'
 
+    def bin_data(dss: Dict[Tuple, list]) -> Dict[Tuple, List]:
```
Kirill888: That's not a valid type annotation

```python
Dict[Tuple[int,int], List[CompressedDataset]]
```

one has to specify what's inside the Tuple/List

Kirill888: same goes for output type annotation

libs/stats/odc/stats/cli.py
tebadi: I've added an example to help string.  It'd be something like %Y-03--P2M. 

I also think frequency is more intuitive, so changed this accordingly.

libs/stats/odc/stats/cli.py
tebadi: Moved binning of data to a separate file.

libs/stats/odc/stats/cli.py
tebadi: It's 'time range'.

libs/stats/odc/stats/cli.py
tebadi: Specified type of data elements in the annotation.

libs/stats/odc/stats/cli.py
tebadi: Sounds better, done.

libs/stats/odc/stats/model.py
tebadi: Removed this block from the construction.

PR #85, opened 2020-09-06
Adding tests and generalizing temporal binning

Kirill888:
Two sets of changes here

1. Tweaks to dataset cache and adding tests
   - create file rather than directory by default
   - do not lock by default (assume single process)
   - move some methods in "class" namespace (static methods instead of module methods)
2. Changes to temporal binning approach


tebadi (approved):
Looks good!

PR #86, opened 2020-09-23
stats: save tasks refactor + more

Kirill888:
- Make a class for it, not just cli
- Started binning refactor, but hasn't finished it yet
- Added support for `file://` output locations to `S3COGSink`
- Added option to set public ACL when writing to S3
- Added k8s job generator notebook


alexgleith (approved):
Looks good to me.

PR #89, opened 2020-09-27
Refactor load tasks and Geomedian work

Kirill888:
### Refactor run_pq

- Take metadata management code into TaskReader class
- Some cleanup in OutputProduct

### Geomedian Work

- Added `wk_rows` parameter to process only that many output rows at any given time, this is to reduce temporary memory used by the computation to assist Dask with scheduling (it didn't help, more below)
- Added `chunked_persist_ds` to persist geomedian one chunk at a time (this didn't work either see below)
- Added cli tool to run S2 Geomedian

Even with the reduced extra memory allocations during the computation, Dask schedules too many concurrent runs of geomedian, particularly closer to the end and runs out of memory, kills the worker and restarts. With only one single worker, like we have in stats, this approach doesn't work and basically, if you let it do it's thing it will keep failing like that for a while. Things to try to overcome this problem:

- Increase chunk size to be large enough that Dask won't attempt to schedule too many Geomedian computations at a time
- Tweak memory management settings of Dask: raise kill threshold, give Dask less memory than what's available (increase reserved memory)
- See if there is a way to tell Dask about temp memory usage for any given task, maybe this is now supported

When trying to manually control scheduling with `ds_chunked_persist` things do proceed but in very sub-optimal way. When computing geomedian all bands are processed together, we then split them into a Dataset. Dask decides to combine "geomedian+extract bandX" steps into one and run all of them concurrently (since there are only 10 tasks running), repeating geomedian computation for each band (10x) for S2 case. This obviously takes much more time, and RAM, than running single geomedian at once and then splitting the result. Things to try to overcome this:

- Increase chunk size to be much larger (> 0.5 all RAM), to force Dask into not optimizing that way
- Disable graph optimization in Dask to force execution of the graph as it is
- Some other way to convince Dask not to repeat Geomedian computation (mark as IO or something)
- Call persist before splitting single data array into a dataset (requires significant refactor)

Ideally I would like to avoid forcing Dask scheduling with chunked_persist, it would be less complex without that. So if a solution exists that avoids that, that would be preferable.

### Other things

- various cleanups
- added transparency pixel coloring option to jpeg output
- notebooks (using jupytetxt, they should work in sanbox, use open with Notebook option)


uchchwhash (approved):
Can't say I grok it a 100% but looks good to me! I need to learn jupytext.

alexgleith (approved):

PR #91, opened 2020-10-07
Chunked processing of Geomedian

Kirill888:
Force Dask to process 1/6 of an image at a time when computing Geomedian.

Settings as hard-coded in the CLI are for 9600x9600 Sentinel2 with 10 bands running on 256G instance with 32 cores.

Run with `--threads=40` to increase compute density (it's mostly waiting for IO)

### DIFF
--- a/libs/stats/odc/stats/_cli_run_gm.py
+++ b/libs/stats/odc/stats/_cli_run_gm.py
@@ -40,6 +40,12 @@ def run_gm(cache_file, tasks, dryrun, verbose, threads, overwrite, public, locat
     from .tasks import TaskReader
     from datacube.utils.dask import start_local_dask
     from datacube.utils.rio import configure_s3_access
+    import dask
+
+    dask.config.set({'distributed.worker.memory.target': False})
+    dask.config.set({'distributed.worker.memory.spill': False})
+    dask.config.set({'distributed.worker.memory.pause': False})
+    dask.config.set({'distributed.worker.memory.terminate': False})
 
     # config
     resampling = 'nearest'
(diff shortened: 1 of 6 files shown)


Kirill888:
Testred on XXXL sandbox:

```
# ~75 daily obs ~ 12 minutes
odc-stats run-gm s2_2019.db 2019--P1Y/4/13 --location file:///tmp/ --verbose --threads 40
# ~150 daily obs ~ 20 minutes
odc-stats run-gm s2_2019.db 2019--P1Y/5/13 --location file:///tmp/ --verbose --threads 40
```

libs/stats/odc/stats/_cli_run_gm.py
```diff
@@ -56,8 +65,15 @@ def run_gm(cache_file, tasks, dryrun, verbose, threads, overwrite, public, locat
         print(repr(rdr))
 
     def _proc(task):
-        ds_in = gm_input_data(task, resampling=resampling)
-        ds = gm_reduce(ds_in)
+        NY, NX = task.geobox.shape
+
+        ds_in = gm_input_data(task, resampling=resampling, chunk=(NY//y_chunks, NX))
+        tdim = list(ds_in.dims)[0]
+        ds_in = ds_in.chunk({tdim: -1, 'x': NX//x_chunks})
```
tebadi, line 72: I'm not sure what tdim: -1 achieves.

tebadi (approved):
Looks good!

alexgleith (approved):

libs/stats/odc/stats/_cli_run_gm.py
Kirill888: combines all time slices into one column, `-1` means make chunk as big as the whole dimension.

Kirill888: since geomedian needs to look at the whole time dimension at once and we use simple `map_blocks` to achieve it we need rechunk that way, `int_geomedian` can do that internally, but it might be best to rechunk once rather many times, rechunking is comparatively expensive operation even when operating within one single worker, and much worse on a larger cluster.

PR #92, opened 2020-10-08
Use .tif

Kirill888:
- Use `.tif` instead of `.tiff`
- Expose some more parameters on cli for GM run


alexgleith (approved):

Kirill888:
@alexgleith I'd rather you didn't squash all these different commits into one

Kirill888:
I had forced pushed those changes as separate commits now, you might have to `git reset` when pulling

PR #93, opened 2020-10-08
Change metadata for stats to STAC

alexgleith:
Probably should have a real test rather than just checking we can convert it.

Happy to add more to this if we need it.


Kirill888:
How confident are you that output generated by this will index with our tools, and which tools are these?

Kirill888:
@alexgleith you'll need to rebase and force push, I have since unsquashed previously squashed PR for tif-tiff.

libs/stats/odc/stats/model.py
```diff
@@ -234,37 +236,63 @@ def metadata_path(self, relative_to: str = 'dataset', ext: str = 'yaml') -> str:
...
+                "href": self.metadata_path('absolute', ext='json')
+            },
+            {
+                "rel": "product_overview",
+                "type": "application/json",
+                "href": product.href
+            }
+        ]
+
+        geobox_wgs84_json = geobox.extent.to_crs('epsg:4326', resolution=math.inf).json
```
Kirill888: this will break for dateline crossing datasets.

libs/stats/odc/stats/model.py
```diff
@@ -234,37 +236,63 @@ def metadata_path(self, relative_to: str = 'dataset', ext: str = 'yaml') -> str:
...
         return {
-            '$schema': 'https://schemas.opendatacube.org/dataset',
+            "type": "Feature",
+            "stac_version": "1.0.0-beta.2",
             'id': str(self.uuid),
-            'product': dict(name=product.name,
-                            href=product.href),
-            'location': self.metadata_path('absolute', ext='yaml'),
-
+            "bbox": [geobox_wgs84_json['coordinates'][0][0] + geobox_wgs84_json['coordinates'][0][2]],
```
Kirill888: this does not make a valid bounding box, we have tools for computing bounding boxes from geometry, should use those.

libs/stats/tests/test_utils.py
```diff
@@ -26,6 +27,17 @@ def gen_compressed_dss(n,
         dt = dt + step
 
 
+def test_stac():
+    product = gm_product(location='/tmp/')
+    reader = TaskReader(str(TEST_DIR / 'test_tiles.db'), product)
+    task = reader.load_task(reader.all_tiles[0])
+
+    odc_meta = stac_transform(task.render_metadata())
```
Kirill888: need to verify at least something about the content of `odc_meta`, for example lineage is kinda important, but also geometry and timestamps and location of files.

libs/stats/odc/stats/model.py
alexgleith: What's the tool?

Kirill888: `.boundingbox` or something, like that

Kirill888: you'll have to delay converting to json to use it.

libs/stats/tests/test_utils.py
alexgleith: Example output: https://gist.github.com/alexgleith/154959394a1ff44061b5d34a55301b54

libs/stats/odc/stats/model.py
```diff
@@ -234,37 +236,64 @@ def metadata_path(self, relative_to: str = 'dataset', ext: str = 'yaml') -> str:
...
+            "type": "Feature",
+            "stac_version": "1.0.0-beta.2",
             'id': str(self.uuid),
-            'product': dict(name=product.name,
-                            href=product.href),
-            'location': self.metadata_path('absolute', ext='yaml'),
-
+            "bbox": [bbox.left, bbox.bottom, bbox.right, bbox.top],
+            "geometry": geobox_wgs84.json,
             'crs': str(geobox.crs),
```
Kirill888: is `crs` a valid STAC key?

libs/stats/odc/stats/model.py
tebadi: Is it Ok that CRS value is hard-coded?  I'm thinking of future proofing the code.

tebadi (approved):

libs/stats/odc/stats/model.py
alexgleith: Nope, good catch ;-)

libs/stats/odc/stats/model.py
alexgleith: Yeah, this should always be lat/lon WGS84 (epsg:4326)

libs/stats/tests/test_utils.py
Kirill888: @alexgleith right and bottom are trimmed by fraction of a pixel

https://gist.github.com/alexgleith/154959394a1ff44061b5d34a55301b54#file-odc-from-stac-json-L304

I believe geometry should not be generated when it covers most of the image anyway, it's meant for defining valid pixel region and should not be present for 99% "full" images.

libs/index/odc/index/stac.py
```diff
@@ -130,8 +130,13 @@ def _geographic_to_projected(geometry, crs):
     geom = Geometry(geometry, 'EPSG:4326')
     geom = geom.to_crs(crs, resolution=math.inf)
 
+    # This is a somewhat dangerous assumption, if data is not in meters or you
+    # are working with very precise data!
+    def round_coords(c1, c2):
+        return [round(coord, 2) for coord in [c1, c2]]
```
Kirill888: this will fail miserably for `crs==epsg:4326`.

These things are not that straightforward. One should really use pixel size as a yard stick for rounding.

alexgleith: >One should really use pixel size as a yard stick for rounding.

That's true. I assumed that `projected` would be projected into a CRS that uses meters, but you're right that data could be in lat/lon natively... I'll change it.

alexgleith: Ok, I have a fix that should be solid.

libs/index/odc/index/stac.py
```diff
@@ -59,7 +59,7 @@ def _stac_product_lookup(item: Document) -> Tuple[str, str, Optional[str], str]:
         product_name = properties.get('odc:product')
         region_code = properties.get('odc:region_code')
         default_grid = "g30m"
-
+
```
Kirill888: spaces?

alexgleith: Mm. I'll fix it and squash the commits.

libs/index/odc/index/stac.py
```diff
@@ -207,7 +209,11 @@ def stac_transform(input_stac: Document, relative: bool = True) -> Document:
     epsg = properties['proj:epsg']
     native_crs = f"epsg:{epsg}"
 
-    geometry = _geographic_to_projected(input_stac['geometry'], native_crs)
+    # Transform geometry to the native CRS at an appropriate precision
+    pixel_size = get_in(['default', 'transform', 0], grids)
+    geometry = geom = Geometry(input_stac['geometry'], 'epsg:4326')
+    if native_crs != 'epsg:4326':
+        geometry = _geographic_to_projected(geometry, native_crs, int(1 // pixel_size))
```
Kirill888: this will always set precision to `0`, I don't know what logic should be, but it's not that.

libs/stats/odc/stats/model.py
```diff
@@ -5,6 +5,8 @@
 from uuid import UUID
 import pandas as pd
 
+import math
```
Kirill888: why is this import special, all on it's own?

libs/stats/odc/stats/model.py
```diff
@@ -234,37 +236,63 @@ def metadata_path(self, relative_to: str = 'dataset', ext: str = 'yaml') -> str:
...
+
         properties = deepcopy(product.properties)
 
         properties.update(self.time_range.to_stac())
         properties['odc:processing_datetime'] = format_datetime(processing_dt, timespec='seconds')
         properties['odc:region_code'] = region_code
+        properties['odc:lineage'] = dict(inputs=inputs)
+        properties['odc:product'] = product.name
+        properties['proj:epsg'] = geobox.crs.epsg
+        properties['platform'] = "sentinel-2"
```
Kirill888: don't like that at all, is landsat WOFS going to be platform `sentinel-2`?

libs/index/odc/index/stac.py
alexgleith: I don't think that's true. If the pixel_size is 0.01 it'll make the precision very high.

Kirill888: `//` is integer division, why are we using integer division for floating point values?

Kirill888: So for `0.01` pixel size do you really need 99 points of precision?

alexgleith: You're right though, it's not a good method... let me think about it a bit.

libs/stats/odc/stats/model.py
alexgleith: Good point.

Kirill888 (approved):

PR #96, opened 2020-10-09
Improvements to stac-to-eo3 conversion

jeremyh:
See this PR on eodatasets3 eo3-to-stac converter for some context: https://github.com/GeoscienceAustralia/eo-datasets/pull/98

- Use stac 'created' field as the odc:prodocessed_datetime if available instead of duplicating the dataset datetime.
- Support the stac 'title' field as the output dataset label (but don't use it for the stable uuid generation)
- Avoid putting raw UUIDs in the dataset label (the field is optional and shouldn't be set if just a uuid)
- Fix landsat 8 OLI_TIRS datasets being recorded as only OLI (it was always throwing away all but the first instrument in the list)

(the 'created' and 'title' fields were requested by DEA Access.)

I've tested this on our Landsat products by doing a round trip conversion:

```bash
$ eo3-to-stac ga_ls5t_ard_3-1-20200605_113081_1988-03-30_final.odc-metadata.yaml
$ python stac_to_eo3.py ga_ls5t*.stac-item.json
$ eo3-to-stac ga_ls8c_ard_3-1-0_088080_2020-05-25_final.odc-metadata.yaml
$ python stac_to_eo3.py ga_ls8c*.stac-item.json
```

Further test cases are welcome -- I'm not sure what some of it is doing.

---

# Known Issues

The round-trip-diff shows a few outstanding things that aren't converted by this tool, but they're unrelated to getting the above PR merged so I haven't addressed them here.

Here's what the diff looks like on a Landsat 8 document after these changes:

(left is the original eo3, right is after conversion and back again)

Screenshot_2020-10-09 Diff Checker

- I'd like to look at those hard-coded product exceptions in this file at some point -- I deliberately avoided changing their behaviour in this PR.

### DIFF
--- a/libs/index/odc/index/stac.py
+++ b/libs/index/odc/index/stac.py
@@ -1,7 +1,7 @@
 import math
 import re
 from pathlib import Path
-from typing import Dict, Tuple, Any, Optional
+from typing import Dict, Tuple, Any, Optional, Iterable
 from uuid import UUID
 
 from datacube.utils.geometry import Geometry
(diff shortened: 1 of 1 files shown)


jeremyh:
Replaced by other PRs, such as #235

PR #98, opened 2020-10-12
Added necessary files to create an image for stats.

tebadi:


docker/libs/stats/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ENV LC_ALL=C.UTF-8
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Do the apt install process, including more recent Postgres/PostGIS
+RUN apt-get update && apt-get install -y wget gnupg \
+    && rm -rf /var/lib/apt/lists/*
+RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | \
+    apt-key add - \
+    && echo "deb http://apt.postgresql.org/pub/repos/apt/ bionic-pgdg main" \
+    >> /etc/apt/sources.list.d/postgresql.list
```
alexgleith: Can we please test without this step as I don't think it's needed.

docker/libs/stats/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+FROM opendatacube/geobase:runner
+COPY --from=env_builder /env /env
+
+ENV PATH="/env/bin:${PATH}"
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+# Copy Datacube bootstrapping script
+# COPY /assets/bootstrap-odc.sh /code/bootstrap-odc.sh
```
alexgleith: Let's get rid of these, please.

docker/libs/stats/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+services:
+  # Start docker container for PostgreSQL to mock RDS
+  dc-db:
+    image: odc-stats-v01
+    ports:
+      - "5434:5432"
+  maildev:
+    image: maildev/maildev
+    ports:
+      - "1080:80"
```
alexgleith: This looks like it came from Datacube Alchemist.

Maildev is for pretending to be a mailserver that you can send emails to.

Also, the `dc-db` should be postgres.

Please copy this: https://github.com/opendatacube/cube-in-a-box and test it before committing.

docker/libs/stats/requirements.txt
```diff
@@ -0,0 +1,21 @@
+--pre
```
alexgleith: I think we can remove `--pre`

docker/libs/stats/docker-compose.yml
alexgleith: Basically, we should have two services:

- postgres: this is the postgres DB
- statistician: this is the Docker image we're building (so use `build: .` and not an image)

docker/libs/stats/requirements.txt
```diff
@@ -0,0 +1,21 @@
...
+--extra-index-url https://packages.dea.ga.gov.au/
+datacube[performance,s3]==1.8.3
+aiobotocore[boto3,awscli]==1.1.1
+aiohttp==3.6.2
+pandas==1.0.5
+tqdm==4.48.2
+digitalearthau
+thredds-crawler
```
alexgleith: Nope

docker/libs/stats/requirements.txt
```diff
@@ -0,0 +1,21 @@
+--pre
+--extra-index-url https://packages.dea.ga.gov.au/
+datacube[performance,s3]==1.8.3
+aiobotocore[boto3,awscli]==1.1.1
+aiohttp==3.6.2
+pandas==1.0.5
+tqdm==4.48.2
+digitalearthau
```
alexgleith: Don't think we need this.

docker/libs/stats/requirements.txt
```diff
@@ -0,0 +1,21 @@
+--pre
+--extra-index-url https://packages.dea.ga.gov.au/
+datacube[performance,s3]==1.8.3
+aiobotocore[boto3,awscli]==1.1.1
+aiohttp==3.6.2
+pandas==1.0.5
+tqdm==4.48.2
```
alexgleith: Not sure about these two?

docker/libs/stats/requirements.txt
```diff
@@ -0,0 +1,21 @@
...
+digitalearthau
+thredds-crawler
+wget
+requests
+odc-aws
+odc-aio
+odc-stats
+odc-thredds
```
alexgleith: Nope

alexgleith (commented):
This needs to be in requirements: `git+https://github.com/opendatacube/datacube-index.git`

See other comments.

docker/libs/stats/Dockerfile
tebadi: It worked without these.  I've deleted these.

docker/libs/stats/Dockerfile
tebadi: Done.

docker/libs/stats/docker-compose.yml
tebadi: I've addressed these.

docker/libs/stats/requirements.txt
tebadi: Deleted.

docker/libs/stats/requirements.txt
tebadi: Deleted pandas, digitalearthau, thredds-crawler, digitalearthau.

docker/libs/stats/requirements.txt
tebadi: Done.

docker/libs/stats/requirements.txt
tebadi: tqdm is required by the indexing.

docker/libs/stats/requirements.txt
tebadi: Removed odc-thredds.

Kirill888:
I'd like to have access to `tiffinfo` command inside this docker,
It's in this apt package `libtiff-tools`.

docker/stats/requirements.txt
```diff
@@ -6,6 +6,7 @@ wget
 requests
 odc-aws
 odc-aio
+odc-algo
```
Kirill888, line 9: The issue is not missing `odc-algo` issue is missing `hdstats` which `odc-algo` is using without declaring dependency on it.

PR #101, opened 2020-10-14
Add GitHub Action to run all tests

alexgleith:
This adds a GitHub action that installs all the apps and libs and runs all the tests.

I have disabled some tests, which @sachitrajbhandari may be able to help with?

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -15,7 +15,6 @@ RUN mkdir -p /conf
 COPY ./tests/requirements.txt /conf/
 RUN env-build-tool new /conf/requirements.txt ${py_env_path} /wheels
 
-
 # Install the tools
 ADD ./ /code
 WORKDIR /code
@@ -36,18 +35,6 @@ RUN /env/bin/pip install --extra-index-url https://packages.dea.ga.gov.au/ \
     --editable libs/thredds \
     --editable libs/ui
 
-FROM opendatacube/geobase:runner
-
-# Copy the environment over
-COPY --from=env_builder /env /env
-
-ENV PATH="/env/bin:${PATH}"
-ENV LC_ALL=C.UTF-8
-ENV LANG=C.UTF-8
-
-# Copy the live code in
-ADD ./ /code
-WORKDIR /code
-
 # Copy in the test runner script
 COPY ./tests/run_tests.sh /usr/local/bin/run_tests.sh
+ENV PATH="/env/bin:${PATH}"
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,0 +1,53 @@
+FROM opendatacube/geobase:wheels as env_builder
+
+ARG py_env_path=/env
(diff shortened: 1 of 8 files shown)


tests/requirements-apt.txt
```diff
@@ -0,0 +1,4 @@
+libyaml-dev
```
Kirill888, line 1: do we really need this indirection via file?
at least rename it to `apt.txt` it has nothing to do with python requirements.txt

tests/requirements.txt
```diff
@@ -0,0 +1,19 @@
+--pre
```
Kirill888: no, never use `--pre` for global installs only for specific libraries

tebadi (approved):

tests/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ENV PATH="/env/bin:${PATH}"
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+# Copy the live code in
+ADD ./ /code
+WORKDIR /code
+
+# Copy in the test runner script
+COPY ./tests/run_tests.sh /usr/local/bin/run_tests.sh
```
Kirill888, line 39: This Docker doesn't make sense to me. Code is copied twice but installed once.

I recommend following approach:
1. build wheels outside of Docker, see other action that does that
2. Create environment using those wheels
3. Copy to runner
4. Run tests in runner

That way we test packaging as well as code, so any mistakes/missing files in packaging are more likely to be caught.

Does it even need to be 2 stage docker?

tests/requirements.txt
```diff
@@ -0,0 +1,18 @@
+--extra-index-url https://packages.dea.ga.gov.au/
+setuptools
+datacube[performance,s3]==1.8.3
+aiobotocore[boto3,awscli]==1.1.1
+aiohttp==3.6.2
+pandas==1.0.5
+digitalearthau
```
Kirill888: what is using `digitalearthau`?

tests/requirements.txt
```diff
@@ -0,0 +1,18 @@
+--extra-index-url https://packages.dea.ga.gov.au/
+setuptools
+datacube[performance,s3]==1.8.3
+aiobotocore[boto3,awscli]==1.1.1
+aiohttp==3.6.2
+pandas==1.0.5
```
Kirill888: No need to pin pandas to older version anymore, let's test with the same setup we have in sandbox currently,

```
pandas==1.1.2
xarray==0.16.1
```

tests/requirements-apt.txt
alexgleith: This is pretty standard in our other repositories.

tests/requirements.txt
alexgleith: Not sure, actually

Kirill888 (approved):

PR #102, opened 2020-10-15
add get_messages  as a helper function.

tebadi:


alexgleith:
I think we should move this to the aws tools, not just as a utils function in the dc_tools folder. Perhaps we create an `queues.py` file in there?

tebadi:
Moved queue helper functions to odc.aws.

alexgleith:
Hey, can you please remove the _cli app runner thing from this PR @tebadi

libs/aws/odc/aws/queues.py
```diff
@@ -0,0 +1,30 @@
...
+            WaitTimeSeconds=10,
+            MessageAttributeNames=["All"],
+        )
+
+        if len(messages) == 0 or (limit and count >= limit):
+            break
+        else:
+            for message in messages:
+                count += 1
+                yield message
```
alexgleith: I think this should have a newline

libs/aws/odc/aws/queues.py
```diff
@@ -0,0 +1,30 @@
...
+    sqs = boto3.resource('sqs')
+    queue = sqs.get_queue_by_name(QueueName=queue_name)
+    return queue
+
+def publish_message(queue, msg):
+    resp = queue.send_message(
+        QueueUrl=queue.url,
+        MessageBody=msg
+    )
+    assert(resp['ResponseMetadata']['HTTPStatusCode'] == 200)
```
alexgleith: Careful with parenthesis here. And this should probably be describing what happened and not just failing.
```suggestion
    assert resp['ResponseMetadata']['HTTPStatusCode'] == 200, "Failed to publish the message"
```

libs/aws/odc/aws/queues.py
tebadi: Added a newline.

libs/aws/odc/aws/queues.py
tebadi: Done.

tebadi:
cli_app_runner changes should be excluded now.

alexgleith (approved):

apps/dc_tools/setup.py
```diff
@@ -25,6 +25,8 @@
         'datacube',
         'odc_index',
         'odc_io',
+        'odc_aws'
+        'odc_aio'
```
Kirill888, line 29: is `odc_aio` legitimately needed? It pulls in `aiobotocore` which can be a pain.

libs/aws/odc/aws/queues.py
```diff
@@ -0,0 +1,31 @@
...
+        QueueUrl=queue.url,
+        MessageBody=msg
+    )
+
+    assert resp['ResponseMetadata']['HTTPStatusCode'] == 200, "Failed to publish the message"
+
+def get_messages(queue, limit):
+    count = 0
+    while True:
+        messages = queue.receive_messages(
```
Kirill888: What happens when there is a Timeout, i.e. queue is empty? Should we catch that and translate into end of iteration.

libs/aws/odc/aws/queues.py
```diff
@@ -0,0 +1,31 @@
...
+            VisibilityTimeout=60,
```
Kirill888: What is `VisibilityTimeout=`. Is it not something we want to expose as a configurable parameter?
Same goes for other timeout.

libs/aws/odc/aws/queues.py
```diff
@@ -0,0 +1,31 @@
+import boto3
+
+def get_queue(queue_name):
```
Kirill888: code that creates s3 client supports configuring profile, region code, and accepting credentials, would be nice if this also had these features, there is code that creates session that takes care of all of that, one can then create sqs resource from that session.

alexgleith: I think we can wait until we need it to deal with this.

libs/aws/odc/aws/queues.py
alexgleith: The timeout is how long a message takes to reappear on the queue.

PR #111, opened 2020-10-26
Fix name of stac-to-dc CLI command

### LINKED ISSUE #108: stac_api_to_dc.py calls itself sqs-to-dc in the cli
I think this might be a copy and paste error:
https://github.com/opendatacube/odc-tools/blob/f313310f1ab4711f09efffe5ef6bbddca9f4b88a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py#L154

alexgleith:
Closes #108


whatnick (approved):
Affects 2 separate issues but LGTM 👍

PR #115, opened 2020-11-08
change http s3 to s3 uri

pindge:
sqs indexing to use s3 path


alexgleith:
I don't think this needs to be a function. I think just change line 120 to:

`uri = f"s3://{bucket_name}/{key}"`

pindge:
> I don't think this needs to be a function. I think just change line 120 to:
> 
> `uri = f"s3://{bucket_name}/{key}"`

Thanks

alexgleith (approved):

PR #116, opened 2020-11-17
Stats work

Kirill888:
- Persist Dask collection into RAM incrementally allowing larger than 1/2 RAM sized datasets to be loaded for processing
- Version of above that also reshapes into YXBT order needed by Geomedian
- Another version of reshape to YXBT approach that is pure Dask (allows to  decouple load chunk size from processing chunk size)
- Adding cloud buffer to geomedian plugin
- Some other minor cleanups in odc.algo

### DIFF
--- a/libs/algo/odc/algo/_dask.py
+++ b/libs/algo/odc/algo/_dask.py
@@ -2,7 +2,7 @@
 Generic dask helpers
 """
 
-from typing import Tuple, Union, cast
+from typing import Tuple, Union, cast, Iterator, List, Any
 from random import randint
 from bisect import bisect_right, bisect_left
 import numpy as np
@@ -10,6 +10,8 @@ import xarray as xr
 from dask.distributed import wait as dask_wait
 import dask.array as da
 from dask.highlevelgraph import HighLevelGraph
+import functools
+import toolz
 from toolz import partition_all
 from ._tools import ROI, roi_shape, slice_in_out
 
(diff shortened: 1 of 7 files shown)


libs/algo/odc/algo/_dask.py
```diff
@@ -108,6 +119,81 @@ def unpack_chunksize(chunk: int, N: int) -> Tuple[int, ...]:
...
+    Return a list of tuples describing output chunks and their relation to input chunks.
+
+    Output: [(dst_idx: int, src_idx: int, src_roi: slice]
+
+    Note that every output chunk has only one chunk on input,
+    so chunking might be irregular on output. This is by design, to avoid
+    creating cross chunk dependencies.
+    """
+    dst_idx = 0
+    for src_idx, src_sz in enumerate(chunks):
```
alexgleith, line 142: Sometimes these abbreviations make it harder to read. I think this one is ok, `src_sz` says 'source size' to me. Just wanted to comment, though.

alexgleith (approved):

libs/algo/odc/algo/_dask.py
Kirill888: I disagree, I find long variable names are much much harder to pattern match visually.
I find that understanding what variable IS in a given context is a "once off operation" human brain does, while seeing where it is being used is done over and over again. If you had `source_index` and `source_size` and then `destination_index` and `destination_size`  then

(a) things do not align vertically
```
source_size
destination_size
# vs
dst_sz
src_sz
```
(b) Lines become much longer making them harder to grasp at a glance, and you are also more likely to end up splitting them vertically, making comprehension even harder still.

That's why it is idiomatic in some domains (like computer vision) to shorten prefixes for source and destination variables to `src_{whatever}, dst_{whatever}`.

PR #117, opened 2020-11-24
Add Azure indexing methods

GypsyBojangles:
Add methods used to index datasets from Azure blobstore.
Indexing uses Azure blob and container clients to search an Azure blobstore, uses a prefix and suffix as filtering for json/yaml needs to be done client side.
https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.containerclient?view=azure-python#list-blobs-name-starts-with-none--include-none----kwargs-
Files are indexed with HTTPS protocol, so this is unsuitable for non-public blobs. Secured blobs will require `az` protocol added to RaserIO and ODC in line with GDAL vsi* protocol support.

Also includes azure-to-tar, while we are deprecating this as a pattern, it was convenient to use for the EY challenge so that everyone could index once without needing special Azure credentials.

### DIFF
--- a/apps/cloud/setup.py
+++ b/apps/cloud/setup.py
@@ -13,16 +13,18 @@ setup(
     license="Apache License 2.0",
     tests_require=["pytest"],
     install_requires=[
-        "odc_aws",
-        "odc_io",
-        "odc_aio",
-        "odc_ppt",
-        "odc_thredds",
+        'odc_aws',
+        'odc_io',
+        'odc_aio',
+        'odc_ppt',
+        'odc_thredds',
+        'odc_azure',
         "click",
     ],
     extras_require={
-        "GCP": ["google-cloud-storage"],
-        "THREDDS": ["thredds_crawler", "requests"],
+        'GCP': ['google-cloud-storage'],
+        'THREDDS': ['thredds_crawler', 'requests'],
+        'AZURE': ['azure-storage-blob']
     },
     entry_points={
         "console_scripts": [
--- a/apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
@@ -14,16 +14,17 @@ from typing import List, Tuple
 
 
 def dump_list_to_odc(
-        account_url,
-        container_name,
-        yaml_content_list: List[Tuple[bytes, str, str]],
-        dc: Datacube,
-        products: List[str],
-        **kwargs,
+    account_url,
+    container_name,
+    yaml_content_list: List[Tuple[bytes, str, str]],
+    dc: Datacube,
+    products: List[str],
+    **kwargs,
 ):
     expand_stream = (
-        (account_url + "/" + container_name + "/" + d[1][:d[1].rfind("/") + 1], d[0]) for d in yaml_content_list if
-        d[0] is not None
+        (account_url + "/" + container_name + "/" + d[1][: d[1].rfind("/") + 1], d[0])
+        for d in yaml_content_list
+        if d[0] is not None
     )
 
     ds_stream = from_yaml_doc_stream(
@@ -61,8 +62,8 @@ def dump_list_to_odc(
     is_flag=True,
     default=True,
     help=(
-            "Default is to fail if lineage documents not present in the database. "
-            "Set auto add to try to index lineage documents."
+        "Default is to fail if lineage documents not present in the database. "
+        "Set auto add to try to index lineage documents."
     ),
 )
 @click.option(
@@ -71,25 +72,31 @@ def dump_list_to_odc(
     default=False,
     help="Default is no verification. Set to verify parent dataset definitions.",
 )
-@click.option('--product', '-p', 'product_names',
-              help=('Only match against products specified with this option, '
-                    'you can supply several by repeating this option with a new product name'),
-              multiple=True)
+@click.option(
+    "--product",
+    "-p",
+    "product_names",
+    help=(
+        "Only match against products specified with this option, "
+        "you can supply several by repeating this option with a new product name"
+    ),
+    multiple=True,
+)
 @click.argument("account_url", type=str, nargs=1)
 @click.argument("containter_name", type=str, nargs=1)
 @click.argument("credential", type=str, nargs=1)
 @click.argument("prefix", type=str, nargs=1)
 @click.argument("suffix", type=str, nargs=1)
 def cli(
-        skip_lineage: bool,
-        fail_on_missing_lineage: bool,
-        verify_lineage: bool,
-        account_url: str,
-        container_name: str,
-        credential: str,
-        product_names: List[str],
-        prefix: str,
-        suffix: str,
+    skip_lineage: bool,
+    fail_on_missing_lineage: bool,
+    verify_lineage: bool,
+    account_url: str,
+    container_name: str,
+    credential: str,
+    product_names: List[str],
+    prefix: str,
+    suffix: str,
 ):
     print(f"Opening AZ Container {container_name} on {account_url}")
     print(f"Searching on prefix '{prefix}' for files matching suffix '{suffix}'")
@@ -109,7 +116,7 @@ def cli(
         product_names,
         skip_lineage=skip_lineage,
         fail_on_missing_lineage=fail_on_missing_lineage,
-        verify_lineage=verify_lineage
+        verify_lineage=verify_lineage,
     )
 
     print(f"Added {added} Datasets, Failed to add {failed} Datasets")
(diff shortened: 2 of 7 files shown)


apps/cloud/setup.py
```diff
@@ -22,17 +22,20 @@
...
         'odc_thredds',
+        'odc_azure',
         "click",
     ],
 
     extras_require={
         'GCP': ['google-cloud-storage'],
-        'THREDDS': ['thredds_crawler', 'requests']
+        'THREDDS': ['thredds_crawler', 'requests'],
+        'AZURE': ['azure-storage-blob']
```
Kirill888, line 27: doesn't really matter, but for the future: Python is ok with trailing commas and it helps with future diffs.

apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
```diff
@@ -0,0 +1,115 @@
+"""Crawl Thredds for prefixes and fetch YAML's for indexing
```
Kirill888, line 1: doc cut and paste?

libs/azure/odc/azure/__init__.py
```diff
@@ -0,0 +1,56 @@
+"""Azure blob storage crawling and YAML fetching utilities
+"""
+from azure.storage.blob import ContainerClient, BlobClient
+from typing import List, Tuple, Optional
+from multiprocessing.dummy import Pool as ThreadPool
+from functools import partial
+
+def find_blobs(account_url: str, container_name: str, credential: str, prefix: str, suffix: str):
```
Kirill888: I would make this into iterator, so `yield blob_name` instead of `blob_list.append(..)`.
You can still make list out of it with `list(find_blobs(...))` if that's what's needed later on.

Also missing return type ` -> List[str]` with the current code, or `-> Iterator[str]` if changing to iterator.

libs/azure/odc/azure/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+def _download(account_url: str, container_name: str, credential: str,
+              blob_name: str) -> Tuple[Optional[bytes], str, Optional[str]]:
+    """Internal method to download YAML's from Azure via BlobClient
+    Arguments:
+        account_url {str} -- Azure account url
+        container_name {str} -- Azure container name
+        credential {str} -- Azure credential token
+        blob_name {str} -- Blob name to download
+    Returns:
+        tuple -- URL content, target file and placeholder for error
```
Kirill888, line 69: so like 
`(url:str, content: bytes, None)` on success
 `(url:str, None, ErrorMessage: str)` on error?

apps/dc_tools/setup.py
```diff
@@ -26,7 +26,8 @@
         'odc_index',
         'odc_io',
         'odc_aws',
-        'odc_aio'
+        'odc_aio',
+        'odc_azure'
```
Kirill888: should we hide that behind feature flag? how big is `azure-storage-blob` lib?

libs/azure/odc/azure/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    container = ContainerClient(account_url=account_url, container_name=container_name, credential=credential)
+    for blob_record in container.list_blobs(name_starts_with=prefix):
+        blob_name = blob_record['name']
+
+        if blob_name.endswith(suffix):
+            blob_list.append(blob_name)
+    return blob_list
+
+
+def download_yamls(account_url: str, container_name: str, credential: str, yaml_urls: List[str],
```
Kirill888: Ideally should be iterator, but still with concurrency.

Kirill888: not being iterator limits how much data one can return at any given time, because RAM

libs/azure/tests/test_azure.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_find_blobs():
+    """Find blobs in a sample Azure account, this will fail if the blob store changes or is removed
+    """
+
+    account_name = "geoau"
+    account_url = "https://" + account_name + ".blob.core.windows.net"
+    container_name = "ey-gsa"
+    credential = "sv=2019-10-10&si=ey-gsa-ro&sr=c&sig=4T2mncGZ2v%2FqVsjjyHp%2Fv7BZytih8b251pSW0QelT98%3D"
```
Kirill888: is that a secret secret?

libs/azure/odc/azure/__init__.py
Kirill888: I would prefer `url` first, as that thing is always returned, and then the rest is kind of Result|Failure tuple.

apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
```diff
@@ -0,0 +1,115 @@
...
+from odc.index import from_yaml_doc_stream
+from datacube import Datacube
+
+from typing import List, Tuple
+
+
+def dump_list_to_odc(
+        account_url,
+        container_name,
+        yaml_content_list: List[Tuple[bytes, str, str]],
```
Kirill888: Missing Optional for bytes/error data type

libs/azure/odc/azure/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    Arguments:
+        account_url {str} -- Azure account url
+        container_name {str} -- Azure container name
+        credential {str} -- Azure credential token
+        blob_name {str} -- Blob name to download
+    Returns:
+        tuple -- URL content, target file and placeholder for error
+    """
+
+    blob = BlobClient(account_url=account_url, container_name=container_name, credential=credential,
```
Kirill888: Should this be guarded with `try: except`? Function signature suggests error as return value.

Kirill888: by "this" I mean including `return ..` line

libs/azure/tests/test_azure.py
GypsyBojangles: i guess it is - it is read only, and we were planning on distributing it with the EY datascience challenge, but the issue is we dont currently have a good location that we control. I was tempted to use my personal Azure, but that isn't going to available once my trail runs out and I am not aware of any other location in Azure where yamls are stored.

Kirill888: Best to read from `os.environ` and either default to that one, or skip test if not set.

libs/azure/odc/azure/__init__.py
GypsyBojangles: This was some bad copy-pasta from the THREDDs lib. That error placeholder was for capturing HTTP errors. I have removed this.

libs/azure/odc/azure/__init__.py
GypsyBojangles: This one I wasn't sure how to approach. Are you aware of a example I could model this off?

libs/azure/odc/azure/__init__.py
GypsyBojangles: This was some bad copy-pasta from the THREDDs lib. That error placeholder was for capturing HTTP errors. I have removed this. With the updated the function signature, is this still required?

libs/azure/odc/azure/__init__.py
alexgleith: There's a lazy example here: https://github.com/opendatacube/odc-tools/blob/develop/libs/aws/odc/aws/queue.py#L25

The idea is to use the Python 'generator' model, so instead of returning, you `yield` a single item. Just be aware, `yield` is not the same as return, so you need to keep the flow of code happening after.

Homework: https://wiki.python.org/moin/Generators

Kirill888: It's easy enough to have `Iterator -> Iterator` for a single-threaded fetch, and it's easy enough to have `List->List` with multi-threading. But having `Iterator->Iterator` with multi-threading is much more involved. So let's just keep it as is, I assume faster fetch for data that fits in memory is more important than dealing with so much data it doesn't fit to RAM.

libs/azure/odc/azure/__init__.py
Kirill888: This signature still suggests "error as value" result with `(None, url)` indicating failure for a given url. And I think that's what you want, given that you are using `pool.map` and probably don't want the entire lot to fail when one blob fails to fetch. 

If this just raises exception on error then signature should be just `url:str -> bytes` or exception if anything goes wrong.

I would also put `blob_name` as the first argument, you can call `pool.map` with lambda, in case this order is motivated by the use of `partial` in code that calls this, or use named arguments with `partial`

```python
pool.map(partial(_download, 
        account_url=account_url, 
        conatiner_name=container_name, 
        credential=credential), yaml_urls)
```

Kirill888: Also is there a standard way to represent `container_name+blob_name` as a url? Is `account_url` + `credentials` related, or is it more like `account_url + container_name + blob_name`?

PR #121, opened 2020-12-02
Fail fast if candidate products do not exist.

tebadi:


apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -152,6 +152,12 @@ def cli(
 
     # Consume generator and fetch YAML's
     dc = Datacube()
+    odc_products = dc.list_products().name.values
+
+    assert set(candidate_products).issubset(
+        set(odc_products)
+    ), f"{candidate_products} is not a subset of {odc_products}"
+
```
omad:
```suggestion
    odc_products = set(odc_products)
    if not set(candidate_products).issubset(odc_products):                                
        missing_products = list(set(candidate_products) - odc_products)                   
        print(f"Error: Requested Product/s {', '.join(missing_products)} {'is' if len(missing_products) == 1 else 'are'} "
               "not present in the ODC Database", file=sys.stderr)
        sys.exit(1)

```

omad (changes requested):
Looks like a useful feature, but lets make it a bit prettier.

It is possible to disable assertions in python (not that anyone ever really does), so using them for error handling if frowned upon. Printing a stack trace to the screen when an error occurs is ugly too.

I haven't tested my suggestion (and probably got carried away, sorry).

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
alexgleith: For some context, I read about assertions. And they should ONLY be used for conditions that are not EVER expected to be met! 

For things that we expect happen, exceptions are better.

alexgleith:
Can you please finish this off @tebadi :-)

alexgleith (approved):

PR #123, opened 2020-12-10
Stats orchestration

tebadi:


libs/aws/odc/aws/queue.py
```diff
@@ -16,13 +17,16 @@ def publish_message(queue, message: str, message_attributes: Mapping[str, Any] =
...
 
 
-def get_messages(queue, limit: bool = None, visibility_timeout: int = 60, message_attributes: Iterable[str] = ["All"]):
+def get_messages(
+    queue,
+    limit: bool = None,
+    visibility_timeout: int = 60,
+    message_attributes: Iterable[str] = ["All"],
+):
     """
```
Kirill888, line 31: needs docs, particularly units for timeout

libs/stats/odc/stats/_cli_publish_tasks.py
```diff
@@ -0,0 +1,28 @@
...
+        print(f"Found {len(tasks):,d} tasks in the file")
+        return tasks
+
+    queue = get_queue(queue)
+    tasks = get_tasks(db)
+    for ta in tasks:
+        task = (ta[0], str(ta[1]), str(ta[2]))
+        task = ",".join(task)
+        # ToDo: add a check that ensures a message has been delivered
+        publish_message(queue, task)
```
alexgleith: When we're pushing `4,000` tasks this will be too slow.

I think pushing 10 at a time is preferable https://github.com/opendatacube/datacube-alchemist/blob/main/datacube_alchemist/worker.py#L189

Also, you don't need to check something has been delivered, because you'll get an Exception if they're not.

libs/stats/odc/stats/cli.py
```diff
@@ -2,3 +2,5 @@
 from . import _cli_save_tasks
 from . import _cli_run_pq
 from . import _cli_run_gm
+from . import _cli_pq_queue
+from . import _cli_publish_tasks
```
alexgleith: newline?

libs/stats/odc/stats/proc.py
```diff
@@ -89,17 +92,17 @@ def prep_stage(tasks: Iterable[Task],
...
+        cog = client.compute(sink.dump(task, ds), fifo_timeout="1ms")
         rr = dask_wait(ds)
         assert len(rr.not_done) == 0
         del ds, rr
         in_flight_cogs.add(cog)
 
     done, _ = drain(in_flight_cogs)
     for r in done:
-        yield r
+        print(r)
```
alexgleith, line 107: Is this mean to be here?

libs/stats/odc/stats/tasks.py
```diff
@@ -55,16 +58,35 @@ def sanitize(v):
...
         return v
+
     return transform_object_tree(sanitize, query)
 
 
+def download_from_s3(cache):
+    s3 = boto3.resource("s3")
+    path_parts = cache.replace("s3://", "").split("/")
+    bucket = path_parts.pop(0)
+    key = "/".join(path_parts)
```
alexgleith: There are common functions all over the place for getting a bucket and key from a s3 URI...

Like: https://github.com/opendatacube/odc-tools/blob/develop/libs/aws/odc/aws/__init__.py#L35

libs/stats/odc/stats/tasks.py
```diff
@@ -55,16 +58,35 @@ def sanitize(v):
...
+    try:
+        filename = Path(key).name
```
alexgleith: Why are we making it a Path and getting the name?

alexgleith: Oh, I see, that's the local file... It's getting plonked in a path relative to wherever this thing is run.

We should either place it in a known location, like `/tmp/{key}` or in a magic tempfile (tempfile.givemeone or whatever).

Also, we should perhaps test if it exists locally first, rather than overwriting it every time we look for it.

libs/stats/odc/stats/tasks.py
```diff
@@ -238,15 +263,19 @@ def msg_default(msg):
...
-    def __init__(self,
-                 cache: Union[str, DatasetCache],
-                 product: Optional[OutputProduct] = None):
+    def __init__(
+        self, cache: Union[str, DatasetCache], product: Optional[OutputProduct] = None
+    ):
+        self.cache_path = None
         if isinstance(cache, str):
+            if "s3://" in cache:
+                self.cache_path = download_from_s3(cache)
```
Kirill888: Should be the other way, you tell `download_from_s3` where to download data to.
Use https://docs.python.org/3.6/library/tempfile.html#tempfile.mkstemp to generate temp file name

libs/stats/odc/stats/tasks.py
```diff
@@ -256,9 +285,13 @@ def __init__(self,
         self._gridspec = gridspec
         self._all_tiles = sorted(idx for idx, _ in cache.tiles(grid))
 
+    def _delete_local_cache(self):
```
Kirill888: Is this supposed to be called from destructor? Does it need a function of it's own?

```python
    def __del__(self):
        # co cleanup
```

libs/stats/odc/stats/tasks.py
Kirill888: Please move to

`odc/aws/__init__.py` and use this function for implementation that is compatible with other s3 methods

https://github.com/opendatacube/odc-tools/blob/86dc6d37e388d6d4e1aad4603fe07a3a4ae61dcf/libs/aws/odc/aws/__init__.py#L295-L305

libs/stats/odc/stats/tasks.py
```diff
@@ -55,16 +58,35 @@ def sanitize(v):
         if isinstance(v, datetime):
             return v.isoformat()
         return v
+
     return transform_object_tree(sanitize, query)
 
 
+def download_from_s3(cache):
```
Kirill888: Must support output to specified location, only default to object name if user hasn't supplied desired output location

Kirill888: also why is parameter called `cache` this function download any s3 object to filesystem.

libs/stats/odc/stats/tasks.py
```diff
@@ -238,15 +263,19 @@ def msg_default(msg):
...
 class TaskReader:
-    def __init__(self,
-                 cache: Union[str, DatasetCache],
-                 product: Optional[OutputProduct] = None):
+    def __init__(
+        self, cache: Union[str, DatasetCache], product: Optional[OutputProduct] = None
+    ):
+        self.cache_path = None
         if isinstance(cache, str):
+            if "s3://" in cache:
```
Kirill888: Use `cache.startswith("s3://")` instead.

libs/stats/odc/stats/tasks.py
```diff
@@ -238,15 +263,19 @@ def msg_default(msg):
 
 
 class TaskReader:
-    def __init__(self,
-                 cache: Union[str, DatasetCache],
-                 product: Optional[OutputProduct] = None):
+    def __init__(
+        self, cache: Union[str, DatasetCache], product: Optional[OutputProduct] = None
+    ):
+        self.cache_path = None
```
Kirill888: should be `self._cache_path` this is private property.

libs/stats/odc/stats/_cli_pq_queue.py
```diff
@@ -0,0 +1,130 @@
...
+    if verbose:
+        print("Starting local Dask cluster")
+
+    client = start_local_dask(threads_per_worker=threads, mem_safety_margin="1G")
+
+    for message in get_messages(queue, limit):
+        try:
+            task_def = get_task_from_message(message)
+            _task = rdr.stream([task_def], product)
+            configure_s3_access(aws_unsigned=True, cloud_defaults=True, client=client)
```
Kirill888: > configure_s3_access(aws_unsigned=True, cloud_defaults=True, client=client)

this should not be inside the loop

libs/stats/odc/stats/_cli_pq_queue.py
```diff
@@ -0,0 +1,130 @@
...
+            if verbose:
+                print(client)
+
+            results = process_tasks(
```
Kirill888, line 107: so this is always one task in this list?

libs/stats/odc/stats/tasks.py
alexgleith: Also, is this cleaning up the dscache file after every tile? It really should cache the cache forever.

libs/aws/odc/aws/queue.py
tebadi: Added documentation.

libs/stats/odc/stats/tasks.py
alexgleith: Agree, making this a generic method to download a file will be better. Or even better, using a method from somewhere else...

The boto download file function is already very simple, so wrapping it up is not necessary, I think: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-example-download-file.html

``` python
import boto3

s3 = boto3.client('s3')
s3.download_file('BUCKET_NAME', 'OBJECT_NAME', 'FILE_NAME')
```

libs/stats/odc/stats/_cli_publish_tasks.py
tebadi: Added support to publish 10 messages at a time.  

There is no guarantee messages are delivered.  We can check failures by checking the response response.get('Failed') - see https://boto3.amazonaws.com/v1/documentation/api/latest/guide/sqs.html

libs/stats/odc/stats/proc.py
tebadi: Yeah, for now - I'm using it for debugging.

libs/stats/odc/stats/tasks.py
tebadi: I've moved this to odc.aws and rely on an existing function 's3_url_parse'.

libs/stats/odc/stats/tasks.py
tebadi: I've moved this to odc/aws/__init__.py.  Reading from s3 is slow so I didn't use this function.  please check the new implementation.

libs/stats/odc/stats/tasks.py
Kirill888: @alexgleith yes it is necessary to support common features across all s3 access methods

- reading without credentials
- reading with specific set of credentials
- any other configuration that differs from default

Those features are not there because I have nothing better to do, they were added in response to real user requirements.

The way `odc.aws` library works is that you setup (or extract from the cache) properly configured `s3` client (that step takes care of various credential needs). Then, you use that to perform various s3 operations, whether that means listing buckets or fetching small objects into memory or fetching large objects to disk. This is library code we need to put some extra effort in, to make it usable in more contexts than just this specific use-case.

tebadi: have changed this to url.

libs/stats/odc/stats/tasks.py
tebadi: Fixed.

libs/stats/odc/stats/tasks.py
tebadi: Added support for this.

libs/stats/odc/stats/_cli_pq_queue.py
tebadi: Moved out of the loop.

libs/stats/odc/stats/_cli_pq_queue.py
tebadi: Yes, that's right.  The name is plural for compatibility with your existing tools (otherwise this has to change in proc.py and will break your commands).

libs/stats/odc/stats/tasks.py
tebadi: @alexgleith are you suggesting we leave the file and it gets destroyed when the pod gets destroyed?

libs/stats/odc/stats/tasks.py
tebadi: This has been refactored.

libs/aws/odc/aws/__init__.py
```diff
@@ -314,6 +316,20 @@ def s3_open(url: str,
     oo = s3.get_object(Bucket=bucket, Key=key, **kwargs)
     return oo['Body']
 
+def cache_s3_object(url,
```
Kirill888: This does not actually cache an object, it will perform new download each time every time to a new path location.

Also is this really functionality we need in this library? I think simple `s3_download` is all that is needed. Let me implement that in the way this library works.

libs/stats/odc/stats/tasks.py
Kirill888: @alexgleith download is per app invocation, not per tile.

alexgleith: Ok good, so long as we're not just caching it for one job!

alexgleith (approved):
Let's give it a go!

Kirill888 (approved):

PR #124, opened 2020-12-14
Use the dev-install.sh script in test environment

alexgleith:
A bunch of tests don't work still when installing using that script.

It's an import error:
```
ImportError while importing test module '/code/libs/stats/tests/test_utils.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.6/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
libs/stats/tests/test_utils.py:8: in <module>
    from odc.stats._gm import gm_product
E   ModuleNotFoundError: No module named 'odc.stats'
=========================== short test summary info ============================
ERROR apps/dc_tools/tests/test_sqs_to_dc.py
ERROR libs/algo/odc/algo/test_dask.py
ERROR libs/algo/odc/algo/test_masking.py
ERROR libs/dscache/odc/dscache/tools/test_tiling.py
ERROR libs/dscache/tests/test_dscache.py
ERROR libs/dscache/tests/test_jsoncache.py
ERROR libs/index/tests/test_stac.py
ERROR libs/stats/tests/test_stats_model.py
ERROR libs/stats/tests/test_utils.py
!!!!!!!!!!!!!!!!!!! Interrupted: 9 errors during collection !!!!!!!!!!!!!!!!!!!!
============================== 9 errors in 1.90s ===============================
```

I need to look into it more.

### DIFF
--- a/scripts/dev-install.sh
+++ b/scripts/dev-install.sh
@@ -16,7 +16,7 @@ install_all_in_edit_mode () {
     #  1. without deps -- install odc-* but none of their dependents
     #  2. with deps -- now that all `odc-*` are here, reinstall getting all other deps
     python3 -m pip install --no-deps -r "${reqs}"
-    python3 -m pip install -r "${reqs}"
+    python3 -m pip install --extra-index-url https://packages.dea.ga.gov.au/ -r "${reqs}"
 
     rm "${reqs}"
 }
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -18,22 +18,9 @@ RUN env-build-tool new /conf/requirements.txt ${py_env_path} /wheels
 # Install the tools
 ADD ./ /code
 WORKDIR /code
-# TODO: work out why this needs `--editable`
-RUN /env/bin/pip install --extra-index-url https://packages.dea.ga.gov.au/ \
-    --editable apps/dc_tools[tests] \
-    --editable apps/cloud \
-    --editable apps/dnsup \
-    --editable libs/algo \
-    --editable libs/aio \
-    --editable libs/aws \
-    --editable libs/dscache \
-    --editable libs/geom \
-    --editable libs/index \
-    --editable libs/io \
-    --editable libs/ppt \
-    --editable libs/stats \
-    --editable libs/thredds \
-    --editable libs/ui
+
+# This installs things twice because of circular dependencies
+RUN /code/scripts/dev-install.sh
 
 # Copy in the test runner script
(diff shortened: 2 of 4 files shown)


scripts/dev-install.sh
```diff
@@ -16,7 +16,7 @@ install_all_in_edit_mode () {
     #  1. without deps -- install odc-* but none of their dependents
     #  2. with deps -- now that all `odc-*` are here, reinstall getting all other deps
     python3 -m pip install --no-deps -r "${reqs}"
-    python3 -m pip install -r "${reqs}"
+    python3 -m pip install --extra-index-url https://packages.dea.ga.gov.au/ -r "${reqs}"
```
Kirill888, line 19: You should not need that `--extra-index-url`.

tests/Dockerfile
```diff
@@ -18,22 +18,9 @@ RUN env-build-tool new /conf/requirements.txt ${py_env_path} /wheels
...
-    --editable libs/dscache \
-    --editable libs/geom \
-    --editable libs/index \
-    --editable libs/io \
-    --editable libs/ppt \
-    --editable libs/stats \
-    --editable libs/thredds \
-    --editable libs/ui
+
+# This installs things twice because of circular dependencies
```
Kirill888, line 22: there no circular dependencies, the reason for doing it twice is to ensure installation installs dev versions of code

If installing just two libraries A and B with "A needs B" dependency, pip will install A (and B from pypi, because A needs it), then it won't install local B, because there is B installed already.

A->B
B

scripts/dev-install.sh
alexgleith: Without it it can't find hdstats.

Kirill888:
Looking at install log it still grabs stuff from packages repo and hence failures
https://github.com/opendatacube/odc-tools/runs/1547905603?check_suite_focus=true#step:3:269

alexgleith:
> Looking at install log it still grabs stuff from packages repo and hence failures
> https://github.com/opendatacube/odc-tools/runs/1547905603?check_suite_focus=true#step:3:269

Maybe doing it the other way round will work better? I.e., install everything including dependencies, and then install the local versions with `--upgrade`

Kirill888:
Not really, as this then requires fetching libs from non-standard locations, this will break when adding new library as it won't be on packages repo yet.

Why did you need to add `--extra--packages-url` to dev install script? The point of the script it to be able to install from local fs only.

Kirill888:
@alexgleith It's not working because odc-libs are already pulled in by requirements.txt

PR #125, opened 2020-12-14
Rework test docker

Kirill888:
- Pass through extra arguments in dev-install to pip as we now have hdstats
  dependency which is not on pip
- Install all odc-stats in edit mode using dev-install.sh
- Use constraints.txt instead of jamming version pins into requirements.txt
- Don't use env-build-tool anymore
- Split docker build int separate step so that timing is apparent
- Ensure that all dependencies are present before installing odc-tools
- Switch to python 3.8


tests/requirements.txt
```diff
@@ -1,18 +1,18 @@
...
-odc-aws
-odc-aio
-odc-thredds
-odc-apps-cloud
-odc-index
+deepdiff
+dask-image
+scikit-image
 # Pinned for Alex to index STAC
 sat-search==0.3.0
```
alexgleith, line 17: Why handle this differently?

tests/constraints.txt
```diff
@@ -0,0 +1,140 @@
...
+urllib3==1.25.11
+wcwidth==0.2.5
+webencodings==0.5.1
+wget==3.2
+widgetsnbextension==3.5.1
+wrapt==1.12.1
+xarray==0.16.2
+yarl==1.6.3
+zict==2.0.0
+zstandard==0.14.1
```
alexgleith, line 140: This is going to be so tedious to maintain...

tests/Dockerfile
```diff
@@ -1,40 +1,31 @@
...
-    --editable libs/geom \
-    --editable libs/index \
-    --editable libs/io \
-    --editable libs/ppt \
-    --editable libs/stats \
-    --editable libs/thredds \
-    --editable libs/ui
+
+ENV PATH="/env/bin:${PATH}"
+RUN ./scripts/dev-install.sh --extra-index-url=https://packages.dea.ga.gov.au/ -c /conf/constraints.txt
```
alexgleith, line 28: Neat solution :-)

tests/requirements.txt
Kirill888: it looked special
`constraints.txt` is "semi-auto" generated, while `requirements.txt` is for intentional pins.

tests/constraints.txt
Kirill888: Not really there are `pip-tools` that help with that, and they can be automated, but "it's not a priority", because infrastructure is never a priority around here. And you know what else is nightmarish to maintain: debugging docker build errors because some lib got new version released just at the wrong time when you PR is failing.

woodcockr: @Kirill888 your description here matches my world today...sigh.

PR #126, opened 2020-12-15
Added a limit for number of published messages.

tebadi:


libs/stats/odc/stats/_cli_publish_tasks.py
```diff
@@ -32,8 +34,9 @@ def get_tasks(cache_file):
             }
         messages.append(message)
         counter += 1
-        # ToDo: add a check that ensures a message has been delivered
+        print(counter)
```
Kirill888: debug output that should not be here?

tebadi: Yeah, it slipped through.  It's removed now.

libs/stats/odc/stats/_cli_publish_tasks.py
```diff
@@ -23,7 +24,8 @@ def get_tasks(cache_file):
 
     messages = []
     counter = 0
-    for ta in tasks:
+
+    for ta in tasks[:limit]:
```
alexgleith, line 28: What happens if limit is `None`?

tebadi: It reads everything.

PR #127, opened 2020-12-15
Stats

Kirill888:
bunch of stuff

### DIFF
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -334,7 +334,8 @@ def enum_to_bool(mask: xr.DataArray,
                  flag_name: str = '',
                  value_true: int = 1,
                  value_false: int = 0,
-                 dtype: Union[str, np.dtype] = 'bool') -> xr.DataArray:
+                 dtype: Union[str, np.dtype] = 'bool',
+                 name: str = "enum_to_bool") -> xr.DataArray:
     """
     This method works for fmask and other "enumerated" masks
 
@@ -366,7 +367,7 @@ def enum_to_bool(mask: xr.DataArray,
 
     if dask.is_dask_collection(mask.data):
         data = da.map_blocks(op, mask.data,
-                             name=randomize("enum_to_bool"),
+                             name=randomize(name),
                              dtype=dtype)
     else:
         data = op(mask)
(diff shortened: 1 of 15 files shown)


libs/stats/odc/stats/_pq.py
```diff
@@ -42,40 +52,89 @@ def pq_product(location: Optional[str] = None) -> OutputProduct:
...
+    .valid            -> .total
+    .erased           -> .clear
+    .erased{postfix}  -> .clear{postfix} For All {postfix}
     """
-    pq = xr.Dataset(dict(clear=xx.clear.sum(axis=0, dtype='uint16'),
-                         total=xx.valid.sum(axis=0, dtype='uint16')))
+    valid = xx.valid
+    erased_bands = [str(n) for n in xx.data_vars if str(n).startswith("erased")]
+    total = valid.sum(axis=0, dtype='uint16')
+    pq = xr.Dataset(dict(total=total))
```
tebadi, line 122: Shouldn't this just be Dict(total)?

libs/algo/odc/algo/_masking.py
```diff
@@ -458,17 +579,87 @@ def _first_valid_np(*aa: np.ndarray, nodata: Union[float, int, None] = None) ->
...
+
+
+def _fuse_and_np(*aa: np.ndarray) -> np.ndarray:
+    """
+    Element wise bit and
+    """
+    assert len(aa) > 0
+    out = aa[0].copy()
+    for a in aa[1:]:
+        out &= a
```
tebadi, line 618: Not an issue, but can rely on numpy.bitwise_and.

libs/stats/odc/stats/_pq.py
```diff
@@ -1,34 +1,44 @@
...
 
 
-bad_pixel_classes = (0, 'saturated or defective')
 cloud_classes = ('cloud shadows',
                  'cloud medium probability',
                  'cloud high probability',
                  'thin cirrus')
 
+default_filters = [(2, 5),
+                   (0, 5)]
```
tebadi, line 29: Please document these number.

tebadi (approved):

libs/algo/odc/algo/_masking.py
Kirill888: I'm not sure if the equivalent function based implementation is much clearer than an operator based one

```python
    out = aa[0].copy()
    for a in aa[1:]:
        np.bitwise_and(out, a, out=out)
```

PR #134, opened 2020-12-21
downgrade message from exception to warning

pindge:
This is a safe message handling not an exception.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
@@ -123,8 +123,8 @@ def handle_bucket_notification_message(
                 [PurePath(key).match(p) for p in record_path]
             ):
                 message.delete()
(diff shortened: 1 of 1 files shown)


apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
```diff
@@ -123,8 +123,8 @@ def handle_bucket_notification_message(
                 [PurePath(key).match(p) for p in record_path]
             ):
                 message.delete()
-                raise SQStoDCException(
-                    f"Key: {key} not in specified list of record_paths, ignoring it."
+                logging.warning(
+                    f"Key: {key} not in specified list of record_paths, delete message from the queue."
```
alexgleith:
```suggestion
                    f"Key: {key} not in specified list of record_paths, deleting message from the queue."
```

alexgleith (approved):
:-)

PR #138, opened 2021-01-07
Change metadata creation to use PySTAC

alexgleith:
Do metadata creation better by using an upstream library.


tebadi (approved):
Nice work!

PR #140, opened 2021-01-15
Reworking Geomedian Plugin

Kirill888:
- Different masking strategy: cloud in any view for this pixel -- no input pixel
- Adds cloud preprocessing: shrink away small clouds (helps with persistent misclassifications), then pad out remaining clouds. This computation happens in the native resolution of the mask band.
- Compute TMADs as part of the output
- Use floating point version of `hdstats.geomedian`, as TMADs for uint version do not work correctly
   - This adds some memory overhead, but too much and it can be controlled by choosing smaller/bigger work chunk size
   - Floating point version of geomedian is more robust and actually runs faster even though it needs to read twice as much data from RAM.
- Use more memory stable approach for loading data

image

Storage is pre-allocated at the start of load operation and every Band-time-slice goes into the right spot (that includes transposition into YXBT order). Only after all the load have completed geomedian computation starts. Compared to a properly dask native approach this can only work on single worker cluster, but has much more predictable performance and lower peak memory usage, also typically faster.

### DIFF
--- a/libs/algo/odc/algo/_memsink.py
+++ b/libs/algo/odc/algo/_memsink.py
@@ -1,93 +1,130 @@
-from typing import Tuple, Union, Optional, Dict
+from typing import Any, Dict, Optional, Tuple, Union
 import numpy as np
+import dask
 import dask.array as da
+from dask.highlevelgraph import HighLevelGraph
 from distributed import Client
 import uuid
+from ._dask import randomize, unpack_chunks, _roi_from_chunks, with_deps
+
 
 ShapeLike = Union[int, Tuple[int, ...]]
 DtypeLike = Union[str, np.dtype]
 ROI = Union[slice, Tuple[slice, ...]]
 MaybeROI = Optional[ROI]
+Delayed = Any
+CacheKey = Union["Token", str]
 
 _cache: Dict[str, np.ndarray] = {}
 
 
+class Token:
+    __slots__ = ["_k"]
+
+    def __init__(self, k: str):
+        # print(f"Token.init(<{k}>)@0x{id(self):08X}")
+        self._k = k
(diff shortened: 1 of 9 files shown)


libs/algo/odc/algo/_memsink.py
```diff
@@ -104,40 +142,208 @@ def store_to_mem(
...
+        dsk[(name, *idx)] = (_chunk_extractor, token.key, _roi(idx))
+
+    dsk = HighLevelGraph.from_collections(name, dsk, dependencies=[token])
+
+    return da.Array(dsk, name, shape=shape, dtype=dtype, chunks=_chunks)
+
+
+def da_mem_sink(xx: da.Array, chunks: Tuple[int, ...], name="memsink") -> da.Array:
+    """
+    It's a kind of fancy rechunk for special needs.
```
omad, line 251: Compute a dask array to memory on a single worker, expose as a dask array with arbitrary chunking.

omad (approved):
Looks good I think. I don't 100% grok everything, but I've spend an hour or so trying to read through and understand.

I don't fully get all the Cache and CachedArray stuff, I can kind of see what it's doing and useful for, but more docs would be useful. The naming isn't obvious enough to me.

libs/algo/odc/algo/_memsink.py
Kirill888: Correct yes. Dask -> RAM -> Dask with different chunking.

All this cache stuff is just so that I can supply a simple string to Dask and then look up an `np.ndarray` from that, because otherwise Dask will try to pickle numpy array and will complain that it's huge. And all the Token stuff is so that cleanup of the Cache can happen.

PR #141, opened 2021-01-18
Expose configuration of odc-stats run via command line

Kirill888:
- Plugin configuration from command-line
- COG settings configuration from cli
- Support per-band COG configuration overrides
- Support tweaking output resolution from cli for quick test runs
- Support tweaking input resampling from cli


tebadi (approved):
Looks good!

PR #142, opened 2021-01-20
CLI command to redrive a queue.

tebadi:


alexgleith:
I like it!

Do you think we should actually make it a real command that's installable outside of Stats?

tebadi:
> u think we should actually make it a real command that's installable outside of Stats?

We need this to reprocess failed messages for different products, so yeah, it sounds like a good addition!  What do you think of moving it under apps/cloud?

alexgleith:
Sounds good to me @tebadi !

alexgleith (approved):

PR #150, opened 2021-02-10
Add update-if-exists flag to sqs-to-dc

alexgleith:
Change the logic in the `do_index_update_dataset` function, and add a new flag so that we can `update-if-exists` to update datasets that are already in the DB.


apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
```diff
@@ -171,25 +171,36 @@ def do_index_update_dataset(
...
+        # Now do something with the dataset
         if ds is not None:
-            if update:
-                updates = {}
-                if allow_unsafe:
-                    updates = {tuple(): changes.allow_any}
-                dc.index.datasets.update(ds, updates_allowed=updates)
-            else:
-                if dc.index.datasets.has(metadata.get("id")):
+            if dc.index.datasets.has(metadata.get("id")):
```
Kirill888, line 188: should read `id` from `ds` not from `metadata` dictionary. So `ds.id` rather than `metadata.get("id")`

apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
```diff
@@ -268,7 +280,7 @@ def queue_to_odc(
                         )
 
             # Index the dataset
-            do_index_update_dataset(metadata, uri, dc, doc2ds, update, allow_unsafe)
+            do_index_update_dataset(metadata, uri, dc, doc2ds, update, update_if_exists, allow_unsafe)
```
Kirill888: there are now quite a few flags here, better to use named parameters rather than rely on order being correct.
`update=update, update_if_exists=update_if_exists, ...`

tebadi (approved):
Looks good!

PR #151, opened 2021-02-10
Adding WofsSummary plugin

Kirill888:
Other:

- Define 10,20,30m grids for Australia
- Add "no resolution" grids `au` and `africa` (have each tile as 1x1 pixel image the size of the tile)
- Smarter formatting of grid index when dumping JSON (auto pick format width)
- More verbose exit from `run` command with explicit `sys.exit(0)` to maybe work around Dask hanging on exit issues
- Helper methods for running numexpr on xarray data (dask and non-dask), used in WofsSummary plugin

### DIFF
--- a/libs/algo/odc/algo/_numexpr.py
+++ b/libs/algo/odc/algo/_numexpr.py
@@ -48,7 +48,38 @@ def apply_numexpr(
     **params,
 ):
     """
-    Apply numexpr to variables within a Dataset
+    Apply numexpr to variables within a Dataset.
+
+    numexpr library offers a limited subset of types and operations supported
+    by numpy, but is much faster and memory efficient, particularly for complex
+    expressions. See numexpr documentation for a more detailed explanation of
+    performance advantages of using this library over numpy operations,
+    summary: single pass over input memory, no temporary arrays, cache
+    locality.
+
+    :param expr: Numexpr compatible string to evaluate
+    :param xx: Dataset object that contains arrays to be used in the ``expr`` (can be Dask)
+    :param dtype: specify output dtype
+    :param name: Used to name computation when input is Dask
+    :param casting: Passed to ``numexpr.evaluate``
+    :param order: Passed to ``numexpr.evaluate``
+    :param params: Any other constants you use in the expression
+    :raturns: xr.DataArray containing result of the equation (Dask is input is Dask)
+
+    Example:
+
+    .. code-block:: python
+
+       # Given a Dataset with bands `red` and `nir`
+       xx = dc.load(..., measurements=["red", "nir"], dask_chunks={})
+
+       # Compute NDVI (ignore nodata for simplicity of the example)
+       ndvi = apply_numexpr("(_1f*nir - red)/(_1f*nir + red)",
+                            xx,
+                            dtype='float32',   # Output is float32
+                            _1f=np.float32(1)  # Define constant `_1f` being a float32(1),
+                                               # used for casting to float32
+                           )
     """
 
     bands = {}
--- a/libs/algo/odc/algo/_numexpr.py
+++ b/libs/algo/odc/algo/_numexpr.py
@@ -129,6 +129,13 @@ def safe_div(x1: xr.DataArray, x2: xr.DataArray, dtype='float32') -> xr.DataArra
     """
     Compute ``x1.astype(dtype)/x2.astype(dtype)`` taking care of cases where x2==0.
 
+    For every element compute the following:
+
+    ::
+
+      x2 is 0 => NaN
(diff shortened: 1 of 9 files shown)


libs/stats/odc/stats/_wofs.py
```diff
@@ -0,0 +1,105 @@
...
+    ):
+        self.resampling = resampling
+
+    def product(self, location: Optional[str] = None, **kw) -> OutputProduct:
+        name = "ga_s2_wo_summary"
+        short_name = "ga_s2_wo_summary"
+        version = "0.0.0"
+
+        if location is None:
+            bucket = "deafrica-stats-processing"  # TODO: ??
```
tebadi, line 25: If the location in None, it should throw an exception.

libs/stats/odc/stats/_gjson.py
```diff
@@ -35,12 +35,21 @@ def timedelta_to_hours(td: timedelta) -> float:
...
     """
+    if title_width == 0:
+        nmax = max([max(abs(ix), abs(iy)) for ix, iy in cells])
+        if nmax < 100:
+            title_width = 3
+        elif nmax < 1000:
+            title_width = 4
+        else:
+            title_width = 5
+
```
tebadi, line 48: How do we decide on these numbers?  Maybe I don't have enough context.

libs/stats/odc/stats/_wofs.py
Kirill888: yeah I just copied it across from previous one, what we really need is support a user supplied/default template .

libs/stats/odc/stats/_gjson.py
Kirill888: number of digits in the 10-base representation of a number +1 for a leading `+` or `-` character.

libs/stats/odc/stats/_gjson.py
```diff
@@ -35,12 +35,21 @@ def timedelta_to_hours(td: timedelta) -> float:
...
     Where ``cells`` is produced by ``odc.index.bin_dataset_stream``
     """
+    if title_width == 0:
+        nmax = max([max(abs(ix), abs(iy)) for ix, iy in cells])
+        if nmax < 100:
+            title_width = 3
+        elif nmax < 1000:
+            title_width = 4
+        else:
+            title_width = 5
```
omad:
```suggestion
        # title_width is the number of digits in the cell index, +1 for positive or negative sign
        title_width = len(str(nmax)) + 1
```

libs/algo/odc/algo/_numexpr.py
```diff
@@ -0,0 +1,110 @@
...
+        coords=sample_band.coords,
+        name=name,
+    )
+
+
+def safe_div(x1: xr.DataArray, x2: xr.DataArray, dtype='float32') -> xr.DataArray:
+    """
+    Compute ``x1.astype(dtype)/x2.astype(dtype)`` taking care of cases where x2==0.
+
+    TODO: currently doesn't treat nodata values in any special way.
```
omad, line 139: Unless it's assumed for everything in `algo`, it's worth mentioning that this is Dask friendly.

Also worth mentioning that it assumes a float output, so "taking care of" means, setting to NaN.

libs/algo/odc/algo/_numexpr.py
```diff
@@ -0,0 +1,110 @@
...
+    xx: xr.Dataset,
+    dtype=None,
+    name="result",
+    casting="safe",
+    order="K",
+    **params,
+):
+    """
+    Apply numexpr to variables within a Dataset
+    """
```
omad, line 83: As a public function I think it needs better docs.
- Dask friendly
- What are all the other args? Are they same as `xr.apply()` or `da.apply()`?
- Why would I want to use this? Is it just faster, or more RAM efficient too?

libs/stats/odc/stats/_wofs.py
```diff
@@ -0,0 +1,105 @@
...
+        wet = apply_numexpr("wet & (~dry) & (~cloud)", xx, dtype="bool")
+        dry = apply_numexpr("dry & (~wet) & (~cloud)", xx, dtype="bool")
+
+        return xr.Dataset(dict(wet=wet, dry=dry, cloud=xx.cloud))
+
+    def input_data(self, task: Task) -> xr.Dataset:
+        chunks = {"y": -1, "x": -1}
+        groupby = "solar_day"
+
+        xx = load_with_native_transform(
```
omad, line 78: With this many args, we should use keywords.

omad (commented):
Looks good, would like some more docs, but happy if you comment in the PR and add them later.

Thanks!

PR #153, opened 2021-02-24
Adding watchdog and SQS renew

Kirill888:
- Update docker image with moto
- Add some moto based unit test
- Add some dummy plugins for testing
- Adding `WorkTokenInterface` class and implementation for that for SQS, responsible to SQS valid period extension
- Redefine `cfg.max_processing_time` to mean "abort computation if taking longer than that amount"
- Add `cfg.job_queue_max_lease` to be maximum timeout on SQS, what `cfg.max_processing_time` used to mean


omad (approved):
Looks great Kirill! I hope it works as well as it looks :)

and, we'll have to write some more docs at some point. :)

PR #156, opened 2021-03-01
Fix SQS to DC by removing attempts to index after archiving

alexgleith:
Perhaps this script is trying to do too much, but this function certainly is. Don't attempt indexing after you're archiving. That's just silly.


omad (approved):
Groovy!

PR #158, opened 2021-03-01
Add Azure Indexing methods

GypsyBojangles:
Add methods used to index datasets from Azure blobstore.
Indexing uses Azure blob and container clients to search an Azure blobstore, uses a prefix and suffix as filtering for json/yaml needs to be done client side.
https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.containerclient?view=azure-python#list-blobs-name-starts-with-none--include-none----kwargs-
Files are indexed with HTTPS protocol, so this is unsuitable for non-public blobs. Secured blobs will require az protocol added to RaserIO and ODC in line with GDAL vsi* protocol support.

Also includes azure-to-tar, while we are deprecating this as a pattern, it was convenient to use for the EY challenge so that everyone could index once without needing special Azure credentials.


Kirill888 (approved):
looks fine, we will need to turn lists into streams at some stage though, loading all into RAM and only then dumping that into db/tar is not ideal.

PR #159, opened 2021-03-10
SQS absolute paths

kieranricardo:
This PR adds an `--absolute` flag to `sqs-to-dc` that triggers `stac_transform` to use absolute paths. 

Prior to this the `stac_transform` would use relative paths which works under the assumption that each dataset has a flat directory structure. For datasets like L2 ARD the metadata document is located at `/dataset/metadata.yaml` and measurements are located at `/dataset/subfolder/measurement.TIF`. `stac_transform(metadata, relative=True)` stores the incorrect relative path of `measurement.TIF. ` instead of `subfolder/measurement.TIF` causing errors when data is loaded.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
@@ -385,6 +385,12 @@ def queue_to_odc(
     default=None,
     help="A path to a list (one item per line, in txt or gzip format) of valide region_codes to include",
 )
+@click.option(
+    "--absolute",
+    is_flag=True,
+    default=False,
+    help="Use absolute paths when converting from stac",
+)
 @click.argument("queue_name", type=str, nargs=1)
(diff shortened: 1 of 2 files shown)


Kirill888 (approved):
looks good,
please use "squash" option when merging, you have 3 commits for what logically should be just 1.

apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
```diff
@@ -400,14 +406,15 @@ def cli(
...
+    absolute,
     queue_name,
     product,
 ):
     """ Iterate through messages on an SQS queue and add them to datacube"""
 
     transform = None
     if stac:
-        transform = stac_transform
+        transform = lambda stat_doc: stac_transform(stat_doc, not absolute)
```
Kirill888: actually, maybe explicitly naming second parameter to `stac_transform` would be clearer.

PR #160, opened 2021-03-10
Add support for "band" for stac conversion.

danlipsa:


danlipsa:
@alexgleith Please review.

libs/index/odc/index/stac.py
```diff
@@ -102,10 +102,13 @@ def _get_stac_bands(
             }
 
         path = asset["href"]
+        band_index = asset["band"] if "band" in asset else None
```
alexgleith:
```suggestion
        band_index = asset.get("band", None)
```
Keep it simple!

alexgleith (changes requested):
If we can simplify that one line then I'm happy 🙂

libs/index/odc/index/stac.py
danlipsa: Thanks for the review. Yes, this makes sense. See the updated code.

alexgleith (approved):

danlipsa:
Any estimate of when the pypi odc-index package will be updated? Thanks!

PR #161, opened 2021-03-11
Fix band_index == 0 case.

danlipsa:


danlipsa:
@Kirill888 Please review.

PR #162, opened 2021-03-11
Configuration for deafrica dev, PQ runs.

tebadi:


tebadi (commented):
I'm not sure whether we need to merge this to ode-tools but the aim is to review the pq_cfg.yaml

Kirill888:
I have branch (PR soon) with example config already, and I do not think we should include specific configs here, it should be in bitbucket next to job definition if not inside it.

Kirill888:
So please don't merge.
Things look fine in there, but since it's not an example config, I would remove comments, and remove some of the settings that just mirror defaults and were only included for documentation purposes. I'd say aim for minimum config and keep defaults you don't have a need to change unchanged. That way if we find better defaults in the future we won't be stuck with worse config.

libs/stats/docs/pq-cfg.yaml
```diff
@@ -0,0 +1,69 @@
...
+  properties: {}
+
+# Plugin Specific Configuration
+# defaults for PQ are currently this
+#
+# plugin_config:
+#   filters: [[2, 5], [0, 5]]
+#   resampling: nearest
+
+# Other common settings
```
Kirill888: delete things that were documentation sake.

libs/stats/docs/pq-cfg.yaml
```diff
@@ -0,0 +1,69 @@
...
+# as along as it takes
+max_processing_time: 1200
+
+# SQS specific settings
+#  300s, i.e. 5 minutes is max lease at any given time
+#  extended when getting within 30s of expiry
+#  polling every 5 seconds
+job_queue_max_lease: 300
+renew_safety_margin: 30
+future_poll_interval: 5
```
Kirill888: delete and rely on default settings for that

libs/stats/docs/pq-cfg.yaml
```diff
@@ -0,0 +1,69 @@
...
+  ovr_blocksize: 512
+  overview_resampling: average
+  overview_levels: [2, 4, 8, 16, 32]
+
+  # Bands that require special treatment
+  # Applies to Geomedian Product
+  overrides:
+    rgba:
+      compression: webp
+      webp_level: 90
```
Kirill888: also delete and stick with defaults

libs/stats/docs/pq-cfg.yaml
tebadi: Done.

tebadi: Done.

tebadi: Done.

libs/stats/docs/pq-cfg.yaml
tebadi: I've removed all the redundant configs.  If you're happy, just leave a comment and I'll use that as the final.

libs/stats/docs/pq-cfg.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+# Generic product attributes
+product:
+  name: pc_s2_annual
+  short_name: pc_s2_annual
+  version: 1.0.0
+  product_family: pixel_quality_statistics
+
+  collections_site: explorer.digitalearth.africa
+  producer: digitalearthafrica.org
```
Kirill888, line 14: With index changes we now need

```yaml
  region_code_format: "x{x:03d}y{y:03d}"
```
added to `.product`, I'd put it after `producer` it's "fixed for Africa configs" kinda setting.

libs/stats/docs/pq-cfg.yaml
```diff
@@ -0,0 +1,15 @@
...
+  region_code_format: "x{x:03d}y{y:03d}"
```
Kirill888, line 15: Need more non-default settings

```yaml
s3_public: true
max_processing_time: 600
```

Kirill888: I really think this whole file should be inside k8s job definition.

tebadi: Absolutely, I'm about to raise a PR for the job template and embed this in there.

PR #163, opened 2021-03-12
Change Grids to be positive index only

Kirill888:
- Offset AU and Global/Africa grids to never have negative index values
- New grids still have the same footprint as before, just different label
- Change default region_code formatting to be something like `x03y23`
- Allow region code formatting customization from config (python format string)
- Record product version in every dataset document in `properties["odc:dataset_version"]`, not my choice of naming, this is what is used by `eodataset3` lib, it was originally intended to be "dataset version" with the last digit changing whenever Dataset is re-processed.


tebadi (approved):
Looks good!

PR #168, opened 2021-03-17
Make hdstats optional

kieranricardo:
This PR makes HD stats optional. What do you think @Kirill888?


Kirill888 (approved):
thanks, that was needed for a while now

PR #169, opened 2021-03-17
Make combination of annual + temporal-range work in save tasks

Kirill888:


libs/stats/odc/stats/tasks.py
```diff
@@ -209,6 +209,8 @@ def msg_default(msg):
             tasks = bin_full_history(cells, start=dt_range.start, end=dt_range.end)
         elif self._frequency == "seasonal":
             tasks = bin_seasonal(cells, months=3, anchor=12)
+        elif self._frequency == "annual":
+            tasks = bin_annual(cells)
```
Kirill888, line 213: This ends up with extra year worth of tasks when `temporal_range` was supplied, need to check inside `bin_annual` and `bin_seasonal` for the limits requested by the user as incoming query was padded to deal with UTC vs local day boundary.

alexgleith (approved):

PR #172, opened 2021-03-19
Fix to make stac to dc options optional

### LINKED ISSUE #171: KeyError when bbox is missing in stac_api_to_odc
I am working on passing a proper polygon to the earth-search API using the `intersects` attribute (because a simple bbox for Chile is not workable), but bbox gets in the way. From what I can tell, bbox is not a mandatory field in the earth-search API. 

**[L130](https://github.com/opendatacube/odc-tools/blob/3117d2a2a4e98739bcc0bb48cc6fab02f730fb18/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py#L130)** in stac_api_to_dc.py results in a KeyError if bbox is not supplied. It could perhaps be:
```if "bbox" in config:```
which would prevent the KeyError issue and will then continue passing the rest of the query config.

When supplying both bbox and intersect, the API responds with
`'Expected bbox OR intersects, not both'`

I am ensuring that I am passing valid geojson to the query before calling stac_api_to_odc, but if stac_api_to_odc is checking for a valid bbox, it should probably also check for a valid geom, e.g.:
```
if "intersects" in config:
    try:
        intersects = geojson.loads(json.dumps(config["intersects"]))
        assert(
            intersects.is_valid
        )
    except (TypeError, AttributeError) as e:
        logger.error(f"{type(e).__name__}: Intersect polygon is not valid geojson")
    except AssertionError as e:
        logger.error(f"{type(e).__name__}: Intersects polygon is geojson but validation failed. The error was: '{intersects.errors()}'")
```
as well as managing (probably before above) whether both bbox and intersects are supplied:
```
bbox_exists = "bbox" in config
intersects_exists = "intersects" in config
if not any([bbox_exists,intersects_exists]):
    logger.error("No bbox or intersects polygon exists in search config")
    return 0, 0

if all([bbox_exists,boundary_exists]):
    logger.error("Can't have both bbox and intersects polygon in search config") 
    return 0, 0 
```
(or preferable raise a proper error rather than returning 0, 0 but either is manageable

### PR DESCRIPTION
**Pull request by alexgleith**
Make the CLI search parameters optional, and make a safe check for included BBOX

Closes #171

Thanks for reporting, @JonDHo

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -127,7 +127,7 @@ def stac_api_to_odc(
     **kwargs,
 ) -> Tuple[int, int]:
     # QA the BBOX
-    if config["bbox"]:
+    if config.get("bbox"):
         assert (
             len(config["bbox"]) == 4
         ), "Bounding box must be of the form lon-min,lat-min,lon-max,lat-max"
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -211,18 +211,17 @@ def cli(
 
     candidate_products = product.split()
 
-    config = {
-        "datetime": datetime,
-        "bbox": bbox,
-        "collections": collections,
-    }
+    config = {}
 
     # Format the search terms
     if bbox:
         config["bbox"] = list(map(float, config["bbox"].split(",")))
 
-    if config["collections"]:
+    if collections:
         config["collections"] = config["collections"].split(",")
+
+    if datetime:
+        config["datetime"] = datetime
 
     # Do the thing
     dc = Datacube()
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -127,10 +127,8 @@ def stac_api_to_odc(
     **kwargs,
 ) -> Tuple[int, int]:
     # QA the BBOX
-    if config.get("bbox"):
-        assert (
-            len(config["bbox"]) == 4
-        ), "Bounding box must be of the form lon-min,lat-min,lon-max,lat-max"
+    if config.get("bbox") and len(config["bbox"]) != 4:
+        raise ValueError("Bounding box must be of the form lon-min,lat-min,lon-max,lat-max")
 
     # QA the search
     srch = Search().search(**config)

### CODE REVIEW

File: apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py [hunk 1]
```diff
@@ -127,10 +127,8 @@ def stac_api_to_odc(
     **kwargs,
 ) -> Tuple[int, int]:
     # QA the BBOX
-    if config["bbox"]:
-        assert (
-            len(config["bbox"]) == 4
-        ), "Bounding box must be of the form lon-min,lat-min,lon-max,lat-max"
+    if config.get("bbox") and len(config["bbox"]) != 4:
```
Kirill888, line 130: Rather than extracting "bbox" key twice, one with `.get` and then again with `[]` it's generally better practice to do it once, this also makes it thread-safe unlike current code (I know it does not matter in this code at this point in time)

```python
bbox = config.get("bbox", None)
if bbox is not None:
   if len(bbox) != 4:
       raise ValueError("Bounding box must be of the form lon-min,lat-min,lon-max,lat-max")
```
Also this will still fail if `bbox` is not a list, but some other construct, so should probably check for that. Or use some library that performs that kind of validation internally.

JonDHo:
I don't use this from command line, but I note that the cli interface only allows a limited number of parameters. Perhaps not all STAC interaces support the "intersects" object, but being able to pass a polygon is super-useful for strangely-shaped areas/countries. The current code works perfectly if a geojson object is passed in config['intersects'] for the element84 Earth Search API (SAT-API). See [https://sat-utils.github.io/sat-api/#tocsintersectsfilter](https://sat-utils.github.io/sat-api/#tocsintersectsfilter) and I suspect it would work with others.

This might be a nice feature to allow extra config parameters in the CLI.

JonDHo:
And I can confirm that the new code (including the latest comment) works... just successfully indexed the S2 archive for all of Chile.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
Kirill888: hold on, is `config` our own construct or something coming from STAC API? or is it going to STAC API?

If it's our own then I would recommend using @dataclasses to explicitly define configuration set, then all the checking can happen there.

Kirill888 (approved):
not familiar enough with the code so made assumptions based on function names alone, so if `config` with or without `bbox` key comes from our own code and is passed on to stac api, then this is fine.

Still, I'm a bit uncomfortable with the amount of undocumented dictionaries flynig around stac related code.

alexgleith:
Yeah, I think relying on the external STAC Search library is a bit undesirable. The problem is that not all STAC API's are the same.

They are standardising in a few months, so it'll be a good idea to wait for that and use a client implementation from that.

PR #174, opened 2021-03-25
Upgrade redrive to queue

alexgleith:
* Cleans up some code
* Adds a redrive_to_queue function to be shared in other apps
* Includes tests

### DIFF
--- a/libs/aws/odc/aws/queue.py
+++ b/libs/aws/odc/aws/queue.py
@@ -6,8 +6,8 @@ from typing import Mapping, Any, Iterable, Optional
 def redrive_queue(
     queue_name: str,
     to_queue_name: Optional[str] = None,
-    limit: Optional[int] = None,
-    dryrun: Optional[bool] = False,
+    limit: Optional[int] = 0,
+    dryrun: bool = False,
 ):
     """
     Redrive messages from one queue to another. Default usage is to define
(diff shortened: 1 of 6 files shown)


libs/aws/odc/aws/queue.py
```diff
@@ -3,6 +3,55 @@
 from typing import Mapping, Any, Iterable, Optional
 
 
+def redrive_queue(
+    queue_name: str, to_queue_name: str = None, limit: int = None, dryrun: bool = False
```
Kirill888: if it can be `None` it can not be `str` or `int`, needs to be `Optional[str]` and `Optional[int]`

libs/aws/odc/aws/queue.py
```diff
@@ -3,6 +3,55 @@
...
+):
+    """
+    Redrive messages from one queue to another. Default usage is to define
+    a "deadletter" queue, and pick its "alive" counterpart, and redrive
+    messages to that queue.
+    """
+    dead_queue = get_queue(queue_name)
+    alive_queue = None
+
+    if to_queue_name:
```
Kirill888: prefer explicit check for `None`, `if to_queue_name is None:`

libs/aws/odc/aws/queue.py
Kirill888: For `limit` changing default to `0` might be better than making `Optional[int]`

libs/aws/odc/aws/queue.py
```diff
@@ -3,6 +3,58 @@
...
+            alive_queue = q
+            count += 1
+            if count > 1:
+                raise Exception(
+                    "Deadletter queue has more than one source, please specify the target queue name."
+                )
+        if count == 0:
+            raise Exception(
+                "No alive queue found for the deadletter queue, please check your configuration."
+            )
```
Kirill888: This is better done without iterations
- get a list of source queue candidate
- list has one element -- great use that
- else report error with appropriate message for 0 and >1

```python
qq = list(dead_queue.dead_letter_source_queues.all())
if len(qq) == 0:
   raise ...
elif len(qq) > 1:
   raise ...
assert len(qq) == 1
active_queue = qq[0]
```
```

File: libs/aws/odc/aws/queue.py [hunk 4]
```diff
@@ -3,6 +3,58 @@
 from typing import Mapping, Any, Iterable, Optional
 
 
+def redrive_queue(
+    queue_name: str,
+    to_queue_name: Optional[str] = None,
+    limit: Optional[int] = None,
+    dryrun: Optional[bool] = False,
```
Kirill888: this one does not need to be optional, since default value is also `bool`. `Optional[some_type]` means can be one of `None|value of some_type` it has nothing to do with parameter having a default value and hence "being optional".

alexgleith: Aha, I see. Thanks!

Kirill888 (approved):

PR #178, opened 2021-04-08
rough rust implementation of geomedian

kieranricardo:
Here's is a very draft PR that I whipped up of what an implementation of `geomedian` in rust within the `odc.algo` library might look like.

With `rust` installed this should compile and run by just running `python setup.py install`.

This might be better suited to being placed in a seperate library or even written in another language/python extension like `Numba`, `cpython`, or `c++`.

Some rough timing on my local machine seem to indicate that this is about twice as fast as the optimised (old) `hdstats` library compiled and available on `dea-packages`.

What do you reckon @Kirill888?


libs/algo/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+        for &t in &valid_time_steps {
+            
+            let mut dist: f32 = 0.0;
+            for band in 0..bands {
+                dist += (out_array[[row, column, band]] - in_array[[row, column, band, t]]).powi(2); 
+            }
+            
+            dist = dist.sqrt();
+            
+            let mut inv_dist = 1.0 / dist;
```
Kirill888, line 65: potential divide by zero here

Kirill888: maybe better start off by setting to 0 and only changing if dist within some tolerance.

Kirill888:
this is great, but I think it's better off being a separate lib with `hdstats` compatible api.

libs/algo/odc/algo/_geomedian.py
```diff
@@ -9,6 +9,7 @@
 from ._dask import randomize, reshape_yxbt
 from ._masking import to_float_np, from_float_np
 from ._memsink import yxbt_sink
+from .backend import rust_geomedian
```
Kirill888, line 13: backend will need to be selectable, at least at first to allow easier comparison both in terms of equivalence and performance.

libs/algo/src/lib.rs
```diff
@@ -0,0 +1,105 @@
+use pyo3::prelude::{pymodule, PyModule, PyResult, Python};
+use numpy::{PyArray4, PyArray3, ToPyArray};
+use ndarray::{array, Array, ArrayBase, ArrayView, ArrayViewMut, Axis, Ix1, Ix2, Ix3, Ix4};
+
+
+fn geomedian<'a>(in_array: ArrayView<'a, f32, Ix4>, max_iter: usize, eps: f32) -> Array<f32, Ix3> {
```
Kirill888, line 6: What's the situation with GIL? Is it released, when is it released, when is it re-acquired?

libs/algo/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+    column: usize,
+    max_iter: usize, 
+    eps: f32
+) { 
+
+    let bands = in_array.shape()[2];
+    let time_steps = in_array.shape()[3];
+    
+    let mut valid_time_steps: Vec<usize> = Vec::with_capacity(time_steps);
+    for t in 0..time_steps {
```
Kirill888, line 38: Rather than keeping indexes of valid time slices, it's probably better to extract actual data into temp array (re-used between pixels columns)
- better data cache performance independent of source data memory layout
- allows for arbitrary pre-processing steps (scale normalization or uint16 -> float32 conversion)

kieranricardo: I wasn't sure whether this would be worth the trade-off of copying the data around. If most of the measurements are valid the cache performance would be pretty good anyway, but then again I have no idea what this data is like in practice! Roughly how much of the data is valid?

libs/algo/src/lib.rs
kieranricardo: atm the GIL is just held the entire time, but I think I'll end up releasing it for the call to `geomedian` - what do you reckon?

libs/algo/odc/algo/_geomedian.py
kieranricardo: ah yep, good idea

kieranricardo:
Should this be in a new `lib` in `odc_tools` or in a seperate repo?

Kirill888:
either one, probably completely separate is easie, this thing is quite independent, doesn't need any of the other gunk within `odc.libs`

libs/algo/src/lib.rs
Kirill888: About 50% I'd say, but it also matters how it is stored

libs/algo/src/lib.rs
Kirill888: Has to be once per top level call to be fast enough

PR #179, opened 2021-04-09
Add sha1 output

Kirill888:
Write `.sha1` file with hashes of written files.

- Generate STAC file content and compute SHA1 of that
- Write all COGs (parallel as completed order) and compute SHA1 hashes of each
- Write `.sha1` file (including hash of STAC file that was not written yet)
- Finally write STAC json

### Other Changes

- Change stac output file name to end with `.stac-item.json`, rather than just `.json`
- Update example config with `plugin_config` section.


tebadi (approved):
Looks good!

PR #180, opened 2021-04-13
GQA filtering in save-tasks and GM refactor and fmask fix

Kirill888:
- Add `--gqa` option to `save-tasks`, allows filtering based on `odc:gqa_iterative_mean_xy` property (assumed to be present on every dataset of the source product)
- Add `--frequency=annual-fy` for financial year binning
- Split GM plugin into Generic/GM/S2 version, where GM and S2 just supply different configuration defaults
- Add extra (optional) native pixel transform to GM plugin, it erases data pixels based on the `nodata` value(s) of the mask. This is needed for Landsat GM since edge pixels do not get valid `fmask` classification and so need to be discarded. Cloud masking happens separately from this as we need to handle fusing for those differently.

### DIFF
--- a/libs/stats/odc/stats/_cli_save_tasks.py
+++ b/libs/stats/odc/stats/_cli_save_tasks.py
@@ -31,7 +31,7 @@ from ._cli_common import main
 @click.option(
     "--frequency",
     type=str,
-    help="Specify temporal binning: annual|semiannual|seasonal|all",
+    help="Specify temporal binning: annual|annual-fy|semiannual|seasonal|all",
 )
 @click.option("--env", "-E", type=str, help="Datacube environment name")
 @click.option(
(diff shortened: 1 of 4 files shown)


libs/stats/odc/stats/_cli_save_tasks.py
```diff
@@ -29,7 +29,9 @@
     help="Only extract datasets for a given time range, Example '2020-05--P1M' month of May 2020",
 )
 @click.option(
-    "--frequency", type=str, help="Specify temporal binning: annual|semiannual|seasonal|all"
+    "--frequency",
+    type=str,
+    help="Specify temporal binning: annual|semiannual|seasonal|all",
```
alexgleith: Did you say there was a new option for this?
```suggestion
    help="Specify temporal binning: annual|annual-fy|semiannual|seasonal|all",
```

alexgleith (approved):

PR #181, opened 2021-04-14
Added a cli tool to stats for generating mosaic overviews.

tebadi:


libs/stats/odc/stats/_cli_generate_mosaic.py
```diff
@@ -0,0 +1,125 @@
...
+    dc = Datacube()
+    dss = list(cache.get_all())
+    xx = dc.load(
+        datasets=dss,
+        dask_chunks={"x": 3200, "y": 3200},
+        resolution=(-120, 120),
+        measurements=["red", "green", "blue"],
+    )
+
+    save(xx, location, product.name, verbose)
```
Kirill888, line 121: We don't need this part in here.

Let's just make one tool for capturing db file, and then separate one for mosaic generation which will have to be way more configurable than this.

For this PR, just db saving please.

Kirill888 (changes requested):
just db saving please, mosaics separately and later when we have robust code for huge images.

Kirill888 (approved):
:+1:

PR #187, opened 2021-04-20
Fix broken assumption an only check limit if it's not None

alexgleith:


apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -54,7 +54,7 @@ def get_items(
     count = 0
     for metadata in items.geojson()["features"]:
         count += 1
-        if count > limit:
+        if (limit is not None) and (count > limit):
```
Kirill888, line 56: If `limit` can be None, then it should be marked as `limit: Optional[int]` rather than current `:int`

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -54,7 +54,7 @@ def get_items(
     count = 0
     for metadata in items.geojson()["features"]:
```
Kirill888: rather than incrementing count manually it's best to use `enumerate` instead

Kirill888 (approved):
:+1:

PR #188, opened 2021-04-22
Add heartbeats to stats.

tebadi:


tebadi:
Sorry about all the format changes.  The only addition is the `register_hearbeat()` in `model.py` and a call to the function in `proc.py`.

libs/stats/odc/stats/model.py
```diff
@@ -234,6 +238,17 @@ def extend_if_needed(self, seconds, buffer_seconds: int = 30) -> bool:
             return self.extend(seconds)
         return True
 
+    def register_heartbeat(self):
```
Kirill888: This is not the right place for this. It should be TaskRunner concern, also location of the file should be configurable, we might be running several workers on the same machine.

Kirill888 (changes requested):
Please do formatting change as a separate commit before any other work.

- Move heartbeat into TaskRunner
- Make it configurable with default being switched off
- Add cli option to `odc-stats run` to set heartbeat file location

libs/stats/odc/stats/model.py
```diff
@@ -555,6 +555,9 @@ def default_cog_settings():
     )
     overwrite: bool = False
 
+    # Heartbeat filepath
+    heartbeat_filepath: str = None
```
Kirill888: If it can be None, it should `Optional[str]`

libs/stats/odc/stats/proc.py
```diff
@@ -181,7 +191,7 @@ def _run(self, tasks: Iterable[Task]) -> Iterator[TaskResult]:
                 t0 = tk.start_time
             else:
                 t0 = datetime.utcnow()
-
+
```
Kirill888: trailing whitespace

libs/stats/odc/stats/proc.py
```diff
@@ -166,6 +166,16 @@ def _safe_result(self, f: Future, task: Task) -> TaskResult:
...
 
+    def _register_heartbeat(self, hearbeat_filepath: str):
+        """
+        Records the timestamp at which a hearbeat was detected
+
+        """
+        t_now = self.now()
+        file_obj = open(f"{hearbeat_filepath}/hearbeat.tmp", "w")
+        file_obj.write(t_now.strftime("%Y-%m-%d %H:%M:%S"))
+        file_obj.close()
```
Kirill888:
```suggestion
        with open(hearbeat_filepath, "w") as file_obj:
            file_obj.write(t_now.strftime("%Y-%m-%d %H:%M:%S"))
```

libs/stats/odc/stats/_cli_run.py
```diff
@@ -107,11 +113,12 @@ def run(
             filedb=filedb,
             plugin=plugin,
             threads=threads,
-            memory_limit=memory_limit,
+            memory_limit=memory_limit,
```
Kirill888: whitespace?

Kirill888:
@tebadi can you please rebase into single commit?

Something like


- `git reset --soft HEAD~3` -- this removes last 3 commits, but keeps code changes on disk
- Make further cleanup changes
- `git commit` commit current state as one single change
- `git push --force-with-lease` push to github, needs extra arg as you changed history

libs/stats/odc/stats/proc.py
Kirill888: Hold on, we want actual filepath supplied, not directory for it.

tebadi: Initially I was thinking of having a file with a known path that I can read in the liveness prob section of the job template.  We can use a variable inside the template and pass it to the cli as well as using it in the liveness check I suppose.

Kirill888: yes, share path via ENV var between launch command and liveness probe script.

Kirill888 (approved):
:+1:

PR #192, opened 2021-04-26
Feature/ml prediction pipeline

fivejjs:
The code is to create the crop-mask from semiannual geomedia-mad .


libs/stats/odc/stats/_gm_ml_pred.py
```diff
@@ -0,0 +1,545 @@
...
+        return self.bands
+
+    def input_data(self, task: Task) -> xr.Dataset:
+        """
+        assemble the input data and do prediction here.
+        This method work as pipeline
+        """
+        dc = Datacube(app=self.target_product)
+
+        geobox = self.africa_N.tile_geobox(task.tile_index)
```
Kirill888, line 508: you don't need this, task object already includes `.geobox` property

libs/stats/odc/stats/_gm_ml_pred.py
```diff
@@ -0,0 +1,545 @@
...
+    VERSION = "0.1.0"
+    PRODUCT_FAMILY = "geomedian"
+
+    source_product = 'gm_s2_semiannual'
+    target_product = 'crop_mask_eastern'
+
+    def __init__(self):
+        # target band to be saved
+        self.bands = ('mask', 'prob')  # skip, 'filtered')
+        self.africa_N = GRIDS[f'africa_{PredConf.resolution[1]}']
```
Kirill888, line 495: no, grid comes from Task.

libs/stats/odc/stats/_gm_ml_pred.py
```diff
@@ -0,0 +1,545 @@
...
+        s3_bucket, s3_key = path.split('/')[2], path.split('/')[3:]
+        s3_key = '/'.join(s3_key)
+        with BytesIO() as f:
+            boto3.client("s3", config=Config(signature_version=UNSIGNED)).download_fileobj(Bucket=s3_bucket, Key=s3_key,
+                                                                                           Fileobj=f)
+            f.seek(0)
+            model = joblib.load(f)
+
+    # Path is a local directory 
+    else:
```
Kirill888, line 219: Please use `s3_download` or `s3_open` from here

https://github.com/opendatacube/odc-tools/blob/3e9c2a42a33cf5bf5b82553230685fee74061e70/libs/aws/odc/aws/__init__.py#L308-L362

libs/stats/odc/stats/_gm_ml_pred.py
```diff
@@ -0,0 +1,545 @@
...
+    """
+    down scale band not in exclude list.
+    :param ds:
+    :param exclude:
+    :param scale:
+    :return:
+    """
+    for band in ds.data_vars:
+        if band not in exclude:
+            ds[band] = ds[band] / scale
```
Kirill888, line 339: not sure it's a good idea to modify `ds` object in place.

libs/stats/odc/stats/_gm_ml_pred.py
fivejjs: nice.

libs/stats/odc/stats/_gm_ml_pred.py
fivejjs: I will discuss it with Chad

Kirill888:
@fivejjs please cleanup git history, there are too many "fixup" commits currently. It's ok to "force push" to your own branch.

PR #194, opened 2021-04-27
Support custom external plugins

Kirill888:
Basically

```yaml
plugin: somelib.MyPlugin
```

where `somelib.MyPlugin` must derive from `odc.stats.model.StatsPluginInterface`

### DIFF
--- a/libs/stats/odc/stats/_plugins.py
+++ b/libs/stats/odc/stats/_plugins.py
@@ -1,3 +1,4 @@
+import pydoc
 from typing import Callable, Dict
 from functools import partial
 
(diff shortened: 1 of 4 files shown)


Kirill888:
@fivejjs let's switch your plugins to use this method rather than keeping code in this repo.

alexgleith (approved):
Looks good to me! This is how Alchemist does it, actually all of them are a path to an Object... could do that instead of having internal plugins a special case.

I like it how it is now, anyhow.

libs/stats/odc/stats/_plugins.py
```diff
@@ -14,6 +15,14 @@ def _new(plugin_class, *args, **kwargs) -> StatsPluginInterface:
 
 def resolve(name: str) -> PluginFactory:
     maker = _plugins.get(name)
+    if maker is None:
+        maker = pydoc.locate(name)
+        if maker is not None:
+            mro = getattr(maker, "__mro__", tuple())
```
omad: What's wrong with `issubclass(maker, StatsPluginInterface)`?

omad (approved):

libs/stats/odc/stats/_plugins.py
Kirill888: nothing, could not google it, is all

PR #196, opened 2021-04-30
Saving Larger than RAM Dask image to COG

Kirill888:
Main entry point is `odc.algo.save_cog` it expects georegistered Dask image on input.

Internally `dask.array.store` is used to persist Dask graph into compressed in-memory (or on disk) images. Overviews are computed on the fly block by block. We ensure that chunks are 2048x2048 pixels wide which allows generating 7 levels of overviews without chunk boundary effect issues. Once all the chunks were saved a second pass re-encodes all overviews and image into a single COG. It supports compressing to memory or to disk. Multi-band input are also supported.

Other incidental changes: moved some test files around


libs/algo/odc/algo/_tiff.py
```diff
@@ -0,0 +1,514 @@
...
+
+def save_cog(
+    xx: xr.DataArray,
+    dst: str,
+    blocksize: Optional[int] = None,
+    ovr_blocksize: Optional[int] = None,
+    bigtiff: Union[bool, str] = "auto",
+    temp_folder: Optional[str] = None,
+    overview_resampling: str = "average",
+    use_final_blocksizes: bool = False,
```
omad, line 417: This argument isn't being passed through to the `COGSink` constructor and should be.

Kirill888: good catch fixed that and also first pass compression options

libs/algo/odc/algo/_tiff.py
```diff
@@ -0,0 +1,518 @@
...
+        rio_opts_first_pass=rio_opts_first_pass,
+        use_final_blocksizes=use_final_blocksizes,
+        **extra_rio_opts,
+    )
+
+    rr = da.store(data, sink, lock=False, compute=False)
+
+    # wait for all stores to complete
+    dsk = dict(rr.dask)
+    deps = dsk.pop(rr.key)
```
kieranricardo, line 515: a general dask question, why do you remove `rr.key` from the dask graph here?

libs/algo/odc/algo/_tiff.py
```diff
@@ -0,0 +1,518 @@
...
+        transfer_config = TransferConfig(multipart_threshold=5 * GB)
+        bucket, key = s3_url_parse(url)
+        s3 = boto3.client("s3")
+
+        return s3.upload_fileobj(
+            self._mem, bucket, key, ExtraArgs=kw, Config=transfer_config
+        )
+
+    @staticmethod
+    def dask_finalise(sink: Delayed, extract=False, strict=False, *deps) -> Delayed:
```
kieranricardo, line 387: why is this a static method? is it something to do with needed to wrap the `COGSink` object with `Delayed`?

libs/algo/odc/algo/_tiff.py
```diff
@@ -0,0 +1,518 @@
...
+        **extra_rio_opts,
+    )
+
+    rr = da.store(data, sink, lock=False, compute=False)
+
+    # wait for all stores to complete
+    dsk = dict(rr.dask)
+    deps = dsk.pop(rr.key)
+    name = "cog_finish-" + tk
+    dsk[name] = (lambda sink, *deps: sink, sink.key, *deps)
```
kieranricardo: what's going on here with the `lambda sink`?

libs/algo/odc/algo/_tiff.py
```diff
@@ -0,0 +1,518 @@
...
+        opts = dict(
+            driver="GTiff",
+            bigtiff=bigtiff,
+            tiled=True,
+            blockxsize=_adjust_blocksize(blocksize, info.width),
+            blockysize=_adjust_blocksize(blocksize, info.height),
+            compress="DEFLATE",
+            zlevel=6,
+            predictor=2,
+            num_threads="ALL_CPUS",
```
kieranricardo, line 247: probably another naive question, but does this use the all the cpus on the machine? or all the cpus available to the dask worker?

libs/algo/odc/algo/_tiff.py
Kirill888: It's not operating on `self`, `self` wasn't even created yet at this point.

libs/algo/odc/algo/_tiff.py
Kirill888: I'm basically editing graph returned by `da.store` and replacing it's top level result with my own that performs further computation when all chunks were writen

Kirill888: I could not just depend on the result of `da.store` because it gets somehow optimized away by Dask, not sure if it's a bug or intended behavior on their part, it was not happening in the older version of Dask, but does now.

libs/algo/odc/algo/_tiff.py
Kirill888: This is really just "wait for all deps to finish even though I'm not using None values that were returned from there"

libs/algo/odc/algo/_tiff.py
Kirill888: That I'm not sure, it is probably using all the cores that are visible, so it could be problematic inside k8s where process is allocated fewer cores via resource limits.

libs/algo/odc/algo/_tiff.py
```diff
@@ -0,0 +1,518 @@
...
+        if temp_folder:
+            t_name = temp
+        else:
+            t_dir, t_name = temp[:8], temp[9:]
+
+        ext = ".tif"
+        ii = info
+        bsz = 2048
+        for idx in range(7 + 1):
+            if temp_folder:
```
kieranricardo, line 273: I think this can be pulled out of the loop

Kirill888: agree that this can be structured better, but not sure creating factory method is worth it here.

It's creating this:
```
  /path/file.tif
  /path/file.tif.ovr
  /path/file/tif.ovr.ovr
....
```
and so on. Some times it's just a path other times it's `MemoryFile`

libs/algo/odc/algo/_tiff.py
```diff
@@ -0,0 +1,518 @@
...
+            num_threads="ALL_CPUS",
+        )
+        opts.update(info.gdal_opts())
+        opts.update(extra_rio_opts)
+
+        mem: Optional[MemoryFile] = None
+        self._mem_mine: Optional[MemoryFile] = None
+
+        if isinstance(dst, str):
+            if dst == ":mem:":
```
kieranricardo, line 132: I think `dst` is either set to `str(Path(temp_folder) / f"{t_name}{ext}")` or `MemoryFile(dirname=t_dir, filename=t_name + ext)` so I think this ` dst == ":mem:"` is never taken

libs/algo/odc/algo/_tiff.py
kieranricardo: oh wait it's creating a new `MemoryFile` with the same `dirname` and `filename` each loop - are these distinct objects even though they have the same name and location?

libs/algo/odc/algo/_tiff.py
Kirill888: When coming from `save_cog` -- yes, but we might expose this class in other ways, one doesn't always need COG, normal TIFF is sometimes enough.

libs/algo/odc/algo/_tiff.py
kieranricardo: makes sense, another naive question but why isn't this throwing a syntax error with `lambda sink`?  I thought lambdas always had to have `: ... ` following and can't manage to get `lambda x` to work myself... is this short hand for `lambda sink: return None`?

libs/algo/odc/algo/_tiff.py
kieranricardo: weird, I think its worth putting in a comment explaining this weirdness

libs/algo/odc/algo/_tiff.py
Kirill888: `lambda` is not limited to single argument, this one takes `sink` and a variable number of arguments `, *deps`. So it's
```python
((lambda sink, *deps: sink), sink.key, *deps)
```
which probably reads better

kieranricardo: ohhhhh,  yep I was reading that wrong - can you chuck those brackets in? that reads a lot better for me

libs/algo/odc/algo/_tiff.py
kieranricardo: oh yep I see that now, still wrapping my head around dask! but why do you use this function as opposed to doing something like `dask.delayed(lambda sink, *deps: sink.finalise())`?

kieranricardo (approved):
@Kirill888 this is looking good to me!

PR #198, opened 2021-05-06
Dask Auto Persist

kieranricardo:
This PR adds functionality to automatically parse a dask graph and persist certain nodes to minimize peak memory usage. It currently makes a couple of assumptions:

- the dask graph is connected
- there is only one output node

I don't know whether these assumptions are realistic but they can be easily changed!

The graph parsing works by finding and removing the nodes that are the common descents of all  root nodes. This yields a disconnected graph with multiple connected components. The connected components can then be computed independently from one another, and are grouped by a user specified concurrency limit and computed.

Here's a toy example of a wofs graph with 5 scenes:

Original graph:

image

Processed graph:

image

### DIFF
--- a/libs/algo/odc/algo/_dask.py
+++ b/libs/algo/odc/algo/_dask.py
@@ -9,6 +9,7 @@ from bisect import bisect_right, bisect_left
 import numpy as np
 import xarray as xr
 import dask
+from dask.core import get_dependencies, ishashable, istask
 from dask.distributed import wait as dask_wait, TimeoutError
 import dask.array as da
 from dask.highlevelgraph import HighLevelGraph
@@ -16,6 +17,7 @@ from dask import is_dask_collection
 import functools
 import toolz
 from toolz import partition_all
+import networkx as nx
 from ._tools import ROI, roi_shape, slice_in_out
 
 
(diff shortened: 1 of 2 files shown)


Kirill888:
My main concern with this approach is that `dc.load` graphs often have first layer that is

1. cheap (in RAM and compute, it's just Dataset object really) 
2. shared with further layers

So I think it won't partition neatly like that. Basically a source Dataset will usually contribute to several output tiles.

kieranricardo:
@Kirill888 can you share a code snippet that creates a dask graph with that structure?

Kirill888:
Actually this SHOULD work, for separating timeslices, as those typically have zero cross-dependencies, it just won't work for single timeslice but large YX span case.

woodcockr:
Hi Folks,

This is an interesting idea, look forward to seeing the outcome.

I came across some additional arguments in lower level Dask that are pertinent to the idea of reducing peak memory usage. of note was `dask.array.reductions split_every` argument which alters peak memory usage by tweaking the graph structure when constructing the reduction so that Dask releases the interim results rather than holding them all in memory. Made me wonder if there was more under the hood like this that might be relevant to this problem space. Unfortunately whilst the `xarray` documentation does state that `kwargs` will be passed all the way through, and they are, it would seem `split_every` isn't readily passed through due to some change in `duck typing` (that's my guess based on the error message that `nanmean` can't be called with it even though it can in dask itself - anyway that's an aside). Just wanted to highlight an interest in your approach and wondering about `split_every` and other possible tweaks we might be able to leverage in ODC.

Final thought, I was watching some recent `dask` scheduler related videos and the primary scheduler authors noted there were significant changes in the scheduler underway to improve performance which should be released about now. If you aren't already aware of these changes might be worth checking in case it impacts your graph modifications approach.

PR #199, opened 2021-05-06
Cleanup dependencies in cloud apps and dc tools

Kirill888:
Making sure that AZURE/GCP/THREDDS are only pulled in if requested, adding some
missing dependencies (sat-search, toolz, pyyaml)


Kirill888:
@whatnick just for your info, cloud apps will now need extra feature flags for AZURE and THREDDS, so if you have dockers that need those take note.

PR #200, opened 2021-05-10
Dataset and product fusing

kieranricardo:
This PR adds strict dataset and product fusing. Currently this sets the lineage of the fused datasets to simply be the parent datasets but this can be changed to accomodate fusing two virtual datasets.

### DIFF
--- a/libs/stats/odc/stats/utils.py
+++ b/libs/stats/odc/stats/utils.py
@@ -1,8 +1,15 @@
 import toolz
-from typing import Dict, Tuple, List, Any, Callable
+from typing import Dict, Tuple, List, Any, Callable, Optional
 from collections import namedtuple
 from datetime import datetime
 from .model import DateTimeRange
+from datacube import Datacube
+from datacube.model import DatasetType, Dataset
+from odc.index import odc_uuid
+from datacube.storage import measurement_paths
+from datacube.model import Dataset, DatasetType
+from datacube.index.eo3 import prep_eo3
+
 
 CompressedDataset = namedtuple("CompressedDataset", ["id", "time"])
 Cell = Any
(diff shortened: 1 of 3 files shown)


libs/stats/odc/stats/utils.py
```diff
@@ -170,3 +177,87 @@ def dedup_s2_datasets(dss):
...
+    fused_def["name"] = name
+    fused_def["metadata"] = {"product": {"name": name}, "properties": {"odc:file_format": file_format}}
+    fused_def["description"] = f"Fused products: {def_1['description']}, {def_2['description']}"
+    fused_def["measurements"] = def_1["measurements"] + def_2["measurements"] 
+    fused_def["metadata_type"] = def_1["metadata_type"]
+
+    return dc.index.products.from_doc(fused_def)
+
+
+def fuse_ds(ds_1: Dataset, ds_2: Dataset, product=None) -> Dataset:
```
Kirill888:
```suggestion
def fuse_ds(ds_1: Dataset, ds_2: Dataset, product: Optional[DatasetType] = None) -> Dataset:
```

libs/stats/odc/stats/utils.py
```diff
@@ -170,3 +177,87 @@ def dedup_s2_datasets(dss):
         out.append(chunk[-1])
         skipped.extend(chunk[:-1])
     return out, skipped
+
+
+def fuse_products(type_1: DatasetType, type_2: DatasetType, dc: Datacube):
```
Kirill888:
```suggestion
def fuse_products(type_1: DatasetType, type_2: DatasetType, dc: Datacube) -> DatasetType:
```

libs/stats/odc/stats/utils.py
```diff
@@ -170,3 +177,87 @@ def dedup_s2_datasets(dss):
...
+
+    name = f"fused__{def_1['name']}__{def_2['name']}"
+
+    fused_def["name"] = name
+    fused_def["metadata"] = {"product": {"name": name}, "properties": {"odc:file_format": file_format}}
+    fused_def["description"] = f"Fused products: {def_1['description']}, {def_2['description']}"
+    fused_def["measurements"] = def_1["measurements"] + def_2["measurements"] 
+    fused_def["metadata_type"] = def_1["metadata_type"]
+
+    return dc.index.products.from_doc(fused_def)
```
Kirill888: This requires connection to db, not ideal. 
It only needs this to lookup metadata_type document from db using `metadata_type` field from product spec, but we already have metadata_type, cause we have product objects. Can you please change this to construct `DatasetType` directly with whatever MetadataType object is supplied in the first product.

libs/stats/odc/stats/utils.py
Kirill888: Let's get rid of `dc` parameter, see comment below, but add optional `name: Optional[str] = None` specifying name for fused product, current default will do when it's `None`

Kirill888:
@kieranricardo to make docker compile again change `hdstats` pin to be `hdstats==0.2.1` instead of what is there currently

https://github.com/opendatacube/odc-tools/blob/4c171c4bda580b8ff3a243a4079a716f0563d3f9/tests/constraints.txt#L46

libs/stats/tests/test_utils.py
```diff
@@ -117,3 +124,180 @@ def test_bin_seasonal_mk_season_rules(months, anchor):
...
+        'npv': {'path': 'ga_ls_fc_3_091086_2020-04-04_final_npv.tif'}}
+    }
+    
+    uris = [m["path"] for m in wo_metadata["measurements"].values()]
+    wo_ds = Dataset(wo_product, prep_eo3(wo_metadata), uris=uris)
+    uris = [m["path"] for m in fc_metadata["measurements"].values()]
+    fc_ds = Dataset(fc_product, prep_eo3(fc_metadata), uris=uris)
+
+    fused_ds = fuse_ds(wo_ds, fc_ds, fused_product)
+    fused_ds = fuse_ds(wo_ds, fc_ds)
```
Kirill888, line 294: Can we add a test that verifies that measurement paths of the `fused_ds` is a union of measurement_paths of the two inputs.

libs/stats/tests/test_utils.py
```diff
@@ -117,3 +124,180 @@ def test_bin_seasonal_mk_season_rules(months, anchor):
...
+        'landsat:collection_number': 1,
+        'landsat:landsat_product_id': 'LE07_L1TP_091086_20200404_20200501_01_T1',
+        'landsat:collection_category': 'T1'},
+        'measurements': {'bs': {'path': 'ga_ls_fc_3_091086_2020-04-04_final_bs.tif'},
+        'pv': {'path': 'ga_ls_fc_3_091086_2020-04-04_final_pv.tif'},
+        'ue': {'path': 'ga_ls_fc_3_091086_2020-04-04_final_ue.tif'},
+        'npv': {'path': 'ga_ls_fc_3_091086_2020-04-04_final_npv.tif'}}
+    }
+    
+    uris = [m["path"] for m in wo_metadata["measurements"].values()]
```
Kirill888: These are not uris of measurements, this should be something like `["s3://bucket/fc/003/020/some.yml"]`

libs/stats/tests/test_utils.py
```diff
@@ -117,3 +124,180 @@ def test_bin_seasonal_mk_season_rules(months, anchor):
...
+    fused_ds = fuse_ds(wo_ds, fc_ds, fused_product)
+    fused_ds = fuse_ds(wo_ds, fc_ds)
+
+    with pytest.raises(AssertionError):
+        bad_metadata = deepcopy(fc_metadata)
+        bad_metadata["properties"]["datetime"] = '2020-04-03T23:33:10.644420Z'
+        bad_ds = Dataset(fc_product, prep_eo3(bad_metadata), uris=uris)
+        fused_ds = fuse_ds(wo_ds, bad_ds, fused_product)
+
+    with pytest.raises(AssertionError):
```
Kirill888, line 306: should only wrap statement that is expected to throw in this, move prep steps out of `with`

libs/stats/odc/stats/utils.py
```diff
@@ -170,3 +177,87 @@ def dedup_s2_datasets(dss):
...
+    fused_doc["properties"] = dict()
+    assert doc_1["properties"]["datetime"] == doc_2["properties"]["datetime"] # datetime is the only manditory property
+    
+    # copy over all identical properties
+    for key, val in doc_1["properties"].items():
+        if val == doc_2["properties"].get(key, None):
+            fused_doc["properties"][key] = val
+    
+    measurements = {**measurement_paths(ds_1), **measurement_paths(ds_2)}
+    fused_doc["measurements"] = dict((key, {"path": val}) for key, val in measurements.items())
```
Kirill888: there are other components of measurement, not just `path` they will be lost here. We really want 
- copy all the measurement fields
- make `path` absolute by resolving relative urls

libs/stats/odc/stats/utils.py
```diff
@@ -170,3 +177,87 @@ def dedup_s2_datasets(dss):
...
+    fused_ds = Dataset(product, prep_eo3(fused_doc), uris=list(measurements.values()))
```
Kirill888: In here `uris=[]` are all known location of the dataset itself, not uris of measurements. I think this could be an empty list in this case. And if that does not work try list with a single empty string in it.

libs/stats/odc/stats/utils.py
```diff
@@ -170,3 +177,87 @@ def dedup_s2_datasets(dss):
...
+    fused_doc = dict()
+
+    fused_doc["id"] = odc_uuid(product.name, "0.0.0", sources=[doc_1["id"], doc_2["id"]])
+    fused_doc["lineage"] = {"source_datasets": [doc_1["id"], doc_2["id"]]}
+
+    # check that all grids with the same name are identical
+    common_grids = set(doc_1["grids"].keys()).intersection(doc_2["grids"].keys())
+    assert all(doc_1["grids"][g] == doc_2["grids"][g] for g in common_grids)
+    
+    fused_doc["grids"] = {**doc_1["grids"], **doc_2["grids"]}
```
Kirill888, line 242: I think fusing grids "properly" will be quite involved, so let's move this into a separate function, for now it can be as simple as it is now with a `# TODO:` comment

libs/stats/tests/test_utils.py
```diff
@@ -117,3 +124,180 @@ def test_bin_seasonal_mk_season_rules(months, anchor):
...
+        'description': 'Geoscience Australia Landsat Water Observations Collection 3',
+        'measurements': [{'name': 'water', 'dtype': 'uint8', 'units': '1', 'nodata': 1}],
+        'metadata_type': 'eo3'
+    }
+    return definition
+
+
+@pytest.fixture
+def dc():
+    return Datacube()
```
Kirill888, line 164: So tests are failing because database is not initialized. How about we get rid of the dependency on the database for this test suite. For this to work I suggest we extract this code into 

https://github.com/opendatacube/odc-tools/blob/4c171c4bda580b8ff3a243a4079a716f0563d3f9/libs/index/odc/index/_index.py#L366-L378

into `str -> MetadataType` function that constructs and returns `MetadataType` object from a standard name.

Kirill888:
@kieranricardo this is looking good now, can you please cleanup commit history and force push, then we can merge.

Kirill888 (approved):
:+1:

PR #201, opened 2021-05-13
added dask + numpy percentile

kieranricardo:
This PR adds functionality in `odc.algo` to calculate percentiles of an `xr.Dataset`. Currently this is replicating the index rounding behaviour of [`argpercentile`](https://github.com/opendatacube/datacube-stats/blob/14054f8820ddf2b79691996af8ac7d41e200fe49/datacube_stats/stat_funcs.py#L108) which as fair as I could tell is what's being used to create the `fc_percentile*` products. Although linear interpolation of values may be more appropriate here instead.

One caveat of my implementation is that it assumes `nodata` is the largest possible value and will fail with `int` data that uses `-1` as no data values. This fits our use case of calculating percentiles of `ga_ls_fc_3` but we may want something more general in the future

 I'm planning on adding an accompanying `odc.stats` class in a seperate PR and for now I'm using the code like this:

```python
src = dc.load(datasets=fused_products, output_crs=crs, resolution=(-30, 30), dask_chunks={'x': -1, 'y': -1})

# create dataset without water band
data_vars = dict((band, data) for band, data in src.data_vars.items() if band != "water")
xx = xr.Dataset(data_vars=data_vars, coords=src.coords, attrs=src.attrs)

# mask out bits that aren't dry
is_ok = (src["water"] == 0)
xx = keep_good_only(xx, is_ok, nodata=255)

xr_percentiles = xr_uint_percentile(xx, [0.1, 0.5, 0.9], 255)
```

### DIFF
--- a/libs/algo/odc/algo/_percentile.py
+++ b/libs/algo/odc/algo/_percentile.py
@@ -1,0 +1,78 @@
+from typing import List, Sequence
+import dask.array as da
+import xarray as xr
+import numpy as np
+from ._masking import keep_good_np
+from dask.base import tokenize
+from functools import partial
+
+
+def np_percentile(xx, percentile, nodata):
+
+    if np.isnan(nodata):
+        high = True
+        mask = ~np.isnan(xx)
+    else:
+        high = nodata >= xx.max()
+        mask = xx != nodata
+
+    valid_counts = mask.sum(axis=0)
+
+    xx = np.sort(xx, axis=0)
+    
+    indices = np.round(percentile * (valid_counts - 1))
+    if not high:
+        indices += (xx.shape[0] - valid_counts)
+        indices[valid_counts == 0] = 0
+
+    indices = indices.astype(np.int64).flatten()
+    step = (xx.size // xx.shape[0])
+    indices = step * indices + np.arange(len(indices))
+
+    xx = xx.take(indices).reshape(xx.shape[1:])
+
+    return keep_good_np(xx, (valid_counts >= 3), nodata)
+
+
+def xr_percentile(
+    src: xr.Dataset,
+    percentiles: Sequence,
+    nodata,
+) -> xr.Dataset:
+
+    """
+    Calculates the percentiles of the input data along the time dimension.
+
+    This approach is approximately 700x faster than the `numpy` and `xarray` nanpercentile functions.
+
+    :param src: xr.Dataset, bands can be either
+        float or integer with `nodata` values to indicate gaps in data.
+        `nodata` must be the largest or smallest values in the dataset or NaN.
+
+    :param percentiles: A sequence of percentiles in the [0.0, 1.0] range
+
+    :param nodata: The `nodata` value
+    """
+
+    data_vars = {}
+    for band, xx in src.data_vars.items():
+       
+        xx_data = xx.data
+        if len(xx.chunks[0]) > 1:
+            xx_data = xx_data.rechunk({0: -1})
+        
+        tk = tokenize(xx_data, percentiles, nodata)
+        for percentile in percentiles:
+            name = f"{band}_pc_{int(100 * percentile)}"
+            yy = da.map_blocks(
+                partial(np_percentile, percentile=percentile, nodata=nodata), 
+                xx_data, 
+                drop_axis=0, 
+                meta=np.array([], dtype=xx.dtype),
+                name=f"{name}-{tk}",
+            )
+            
+            data_vars[name] = (xx.dims[1:], yy)
+
+    coords = dict((dim, src.coords[dim]) for dim in xx.dims[1:])
+    return xr.Dataset(data_vars=data_vars, coords=coords, attrs=src.attrs)
--- a/libs/algo/tests/test_percentile.py
+++ b/libs/algo/tests/test_percentile.py
@@ -1,0 +1,88 @@
+from odc.algo._percentile import np_percentile, xr_percentile
+import numpy as np
+import pytest
+import dask.array as da
+import xarray as xr
+
+
+def test_np_percentile():
+    arr = np.array(
+        [[0, 1, 4, 6, 8, 10, 15, 22, 25, 27], [3, 5, 6, 8, 9, 11, 15, 28, 31, 50]]
+    )
+
+    np.random.shuffle(arr[0, :])
+    np.random.shuffle(arr[1, :])
+    arr = arr.transpose()
+
+    assert (np_percentile(arr, 0.5, 255) == np.array([8, 9])).all()
+    assert (np_percentile(arr, 0.7, 255) == np.array([15, 15])).all()
+    assert (np_percentile(arr, 1.0, 255) == np.array([27, 50])).all()
+    assert (np_percentile(arr, 0.0, 255) == np.array([0, 3])).all()
+
+
+@pytest.mark.parametrize("nodata", [255, 200, np.nan, -1])
+def test_np_percentile_some_bad_data(nodata):
+    arr = np.array(
+        [[0, 1, 4, 6, 8, nodata, nodata, nodata, nodata, nodata], [3, 5, 6, 8, 9, 11, 15, 28, 31, 50]]
+    )
+
+    np.random.shuffle(arr[0, :])
+    np.random.shuffle(arr[1, :])
+    arr = arr.transpose()
+
+    assert (np_percentile(arr, 0.5, nodata) == np.array([4, 9])).all()
+    assert (np_percentile(arr, 0.7, nodata) == np.array([6, 15])).all()
+    assert (np_percentile(arr, 1.0, nodata) == np.array([8, 50])).all()
+    assert (np_percentile(arr, 0.0, nodata) == np.array([0, 3])).all()
+
+
+@pytest.mark.parametrize("nodata", [255, 200, np.nan])
+def test_np_percentile_bad_data(nodata):
+    arr = np.array(
+        [
+            [0, 1, nodata, nodata, nodata, nodata, nodata, nodata, nodata, nodata],
+            [3, 5, 6, 8, 9, 11, 15, 28, 31, 50],
+        ]
+    )
+
+    np.random.shuffle(arr[0, :])
+    np.random.shuffle(arr[1, :])
+    arr = arr.transpose()
+
+    np.testing.assert_equal(np_percentile(arr, 0.5, nodata), np.array([nodata, 9]))
+    np.testing.assert_equal(np_percentile(arr, 0.7, nodata), np.array([nodata, 15]))
+    np.testing.assert_equal(np_percentile(arr, 1.0, nodata), np.array([nodata, 50]))
+    np.testing.assert_equal(np_percentile(arr, 0.0, nodata), np.array([nodata, 3]))
+
+
+@pytest.mark.parametrize("nodata", [255, 200, np.nan, -1]) #should do -1
+def test_xr_percentile(nodata):
+    band_1 = np.random.randint(0, 100, size=(10, 100, 200)).astype(type(nodata))
+    band_2 = np.random.randint(0, 100, size=(10, 100, 200)).astype(type(nodata))
+
+    band_1[np.random.random(size=band_1.shape) > 0.5] = nodata
+    band_2[np.random.random(size=band_1.shape) > 0.5] = nodata
+
+    true_results = dict()
+    true_results["band_1_pc_20"] = np_percentile(band_1, 0.2, nodata)
+    true_results["band_2_pc_20"] = np_percentile(band_2, 0.2, nodata)
+    true_results["band_1_pc_60"] = np_percentile(band_1, 0.6, nodata)
+    true_results["band_2_pc_60"] = np_percentile(band_2, 0.6, nodata)
+
+    band_1 = da.from_array(band_1, chunks=(2, 20, 20))
+    band_2 = da.from_array(band_2, chunks=(2, 20, 20))
+
+    attrs = {"test": "attrs"}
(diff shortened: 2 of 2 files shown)


libs/algo/odc/algo/_percentile.py
```diff
@@ -0,0 +1,58 @@
...
+    percentiles: Sequence,
+    nodata,
+) -> xr.Dataset:
+
+    """
+    Calculates the percentiles of the input data along the time dimension.
+
+    :param src: xr.Dataset, bands can be either
+        float or integer with `nodata` values to indicate gaps in data.
+        `nodata` must be the largest values in the dataset or NaN.
```
Kirill888: that's not feasible, and also not necessary, you can just as well deal with smallest value, and even middle value is not that much harder to remove post sorting.

Kirill888: but I guess supporting floating point only is fine.

libs/algo/odc/algo/_percentile.py
```diff
@@ -0,0 +1,58 @@
...
+
+    :param compute_mads: A sequence of percentiles in the [0.0, 1.0] range
+
+    :param nodata: The `nodata` value
+    """
+
+    xx = reshape_for_geomedian(src)
```
Kirill888: I don't think this a good idea. This reshape is quite expensive to perform on large data and it introduces dependencies between bands, which is not needed, you can compute percentiles on every band separately, right?

I say we should keep the same axis order and process every band separately, this thing should operate on an Array not on a Dataset. It would be nice to still accept Dataset in Dataset out for convenience, but per band operations should be independent Dask graphs.

Expected input axis order is T,Y,X single band, output is Y,X,P or Y,X if only one percentile was requested.

So really you only need to make sure that T dimension is one single chunk, by rechunking if necessary.

libs/algo/odc/algo/_percentile.py
```diff
@@ -0,0 +1,58 @@
...
+import dask.array as da
+import xarray as xr
+import numpy as np
+from ._geomedian import reshape_for_geomedian
+from ._masking import keep_good_np
+
+
+def np_percentile(xx, percentile, nodata):
+
+    mask = (xx != nodata) & (~np.isnan(xx))
```
Kirill888: Let's support 2 modes here.
1. Int + Nodata
2. Float with nan and no "nodata"

For int case convert to `float32` with one of the existing methods call itself with that then convert back to original `dtype/nodata`.

For floating point inputs `assert nodata is None` and assume `NaN` sort higher and just keep using current method.

Alternatively only support `float32` on input and perform conversion `int->float32->int` as a Dask operation and hope those things will get fused.

libs/algo/odc/algo/_percentile.py
Kirill888: doing computation in masked int is too hard, let's just compute in float with conversion back, it's almost always `(u)int16` so no loss of precision when computation is done in `float32`, if you want to detect `(u)int32` and use `float64` for those then that's fine too.

options for conversion are
- inside `np_percentile` a bit more annoying, but smaller Dask graph
- as a Dask operation with already existing methods (more Dask nodes, but simpler `np_percentile`)

libs/algo/odc/algo/_percentile.py
Kirill888: @robbibt points out on Slack that xarray already has this operation in Dask form, but also, if that is trying to be too smart and do some non-trivial parallel sorting there is also `np.nanpercentile?` which is probably faster then all this, so we just need to write `int+nodata` wrapper for that.

libs/algo/odc/algo/_percentile.py
```diff
@@ -46,9 +46,11 @@ def xr_percentile(
         
         for percentile in percentiles:
             name = f"{band}_pc_{int(100 * percentile)}"
-            data_vars[name] = da.map_blocks(
-                lambda x: np_percentile(x, percentile, nodata), xx.data, drop_axis=3
+            yy = da.map_blocks(
+                lambda x: np_percentile(x, percentile, nodata), xx, drop_axis=0, meta=np.array([], dtype=xx.dtype)
             )
+            
+            data_vars[name] = (("y", "x"), yy)
```
Kirill888: unfortunately it's not always `y,x`. you are better off using `.dims` of input to keep names as they were.

libs/algo/odc/algo/_percentile.py
```diff
@@ -0,0 +1,56 @@
...
+) -> xr.Dataset:
+
+    """
+    Calculates the percentiles of the input data along the time dimension.
+
+    :param src: xr.Dataset, bands can be either
+        float or integer with `nodata` values to indicate gaps in data.
+        `nodata` must be the largest values in the dataset or NaN.
+
+    :param compute_mads: A sequence of percentiles in the [0.0, 1.0] range
```
Kirill888: `compute_mad` cut and paste

libs/algo/odc/algo/_percentile.py
kieranricardo: aw man, I was really hoping to avoiding duplicating everything to caste as `floats`. Are there use cases where the `nodata` isn't the largest or smallest value in the range?

libs/algo/odc/algo/_percentile.py
kieranricardo: I've timed both this and my original and interestingly they take about the same amount of time. I think the cache friendliness of my approach (~10x `np_percedntile`) compensates for the `rechunk...` cost. 

But your suggestion has a way nicer dask graph and the code is simpler too 👍

libs/algo/odc/algo/_percentile.py
kieranricardo: woops

libs/algo/odc/algo/_percentile.py
kieranricardo: (copying from slack to keep a record)

unfortunately the numpy and  xr_array (which I think just wraps numpy ) are ~700x slower than this implementation!!

libs/algo/odc/algo/_percentile.py
Kirill888: Dask is hard to benchmark, performance changes a lot depending on size and number of worker and everything. I can see how transpose can help with sorting performance, but there is absolutely no need to interleave bands and make them data dependent without reason.

libs/algo/odc/algo/_percentile.py
Kirill888: it's usually smallest, -9999 for otherwise positive data, other common case is `0` for `uint16`, so typically it's the smallest value, but I guess `0xFFFF` can happen too on unsigned data.

Kirill888: placing nodata in the middle of your valid range seems odd, haven't seen that in the wild.

libs/algo/odc/algo/_percentile.py
```diff
@@ -0,0 +1,71 @@
+from typing import List, Sequence, Tuple
+import dask.array as da
+import xarray as xr
+import numpy as np
+from ._geomedian import reshape_for_geomedian
+from ._masking import keep_good_np
+
+
+def np_percentile(xx, percentile, nodata, high=True):
```
Kirill888: default should be `high=False`, but also I don't like boolean flags like that. You have two modes that you want to support

- `nodata` is smallest possible value
- `nodata` is largest possible value

kieranricardo: do you reckon this would be better as two seperate functions?

Kirill888: no, I prefer an "enumeration" type for this option even if it only has 2 possible values to a boolean flag. Like say
`mode = 'nodata_lowest'` or something like that. Cause eventually you'll want `mode = "auto"` and with bool you'll have to do weird things like `None|True|False`, but with named mode it's just one more added to the list.

kieranricardo:
@Kirill888 I'm getting some funky results with `dask` - for some reason it seems to overwrite the  `f"{band}_pc_{int(100 * percentile)}"` arrays with the final one - so they all return the same result.

If I compute the arrays immediately as they're made I get the right result so the error isn't with passing the wrong `percentile` value around or in the `numpy` percentile code.

Any idea what's going on here?

https://github.com/opendatacube/odc-tools/blob/17b6c5992b5870c45175fec8ef4889858271666c/libs/algo/odc/algo/_percentile.py#L65

Kirill888:
I think it's because you are re-using the same name in your map_blocks call

Kirill888:
Have a look at Dask's tokenize function

kieranricardo:
@Kirill888 they all seem to have different names (unless there's another name being used by dask). For my test arrays I tried the following:

```python

# assert no duplicates
assert len(set(arr.name for arr in output.data_vars.values())) == len(output.data_vars.values())

for arr in output.data_vars.values():
    print(arr.name)
```

prints out:

```python
band_1_pc_20
band_1_pc_60
band_2_pc_20
band_2_pc_60
```

Kirill888:
I'm never sure when Dask expect randomized section to the name, try adding that and see if problem persists, as in `band_1_pc_20-126736936178698`, it treats everything after `-` as unique token

kieranricardo:
still failing, and I think things are a little weirder than I thought. This first code snippet failed, which is expected bc its pretty much what's happening in my `xr_percentile` function. But the second snippet works?!?!

**Snippet 1 (fails)**

```python
names = ["band_1_pc_20", "band_1_pc_60"]
percentiles = [0.2, 0.6]
array_list = []

for name, percentile in zip(names, percentiles):
    yy = da.map_blocks(
        lambda x: np_percentile(x, percentile, nodata, high=high), 
        xx.data, 
        drop_axis=0, 
        meta=np.array([], dtype=xx.dtype),
        name=randomize(name),
    )
    
    array_list.append(yy)


assert (array_list[0].compute() != array_list[1].compute()).any()
```

**Snippet 2 (words)**

```python
a = da.map_blocks(
    lambda x: np_percentile(x, percentiles[0], nodata, high=high), 
    xx.data, 
    drop_axis=0, 
    meta=np.array([], dtype=xx.dtype),
    name=randomize(names[0]),
)

b = da.map_blocks(
    lambda x: np_percentile(x, percentiles[1], nodata, high=high), 
    xx.data, 
    drop_axis=0, 
    meta=np.array([], dtype=xx.dtype),
    name=randomize(names[1]),
)
    
assert (a.compute() != b.compute()).any()
```

kieranricardo:
this is probably just me blaming dask for my shoddy code - but maybe this is a dask bug?

libs/algo/odc/algo/_percentile.py
kieranricardo: oh actually I might as well just make this `auto`

libs/algo/odc/algo/_percentile.py
```diff
@@ -65,11 +67,11 @@ def xr_percentile(
         for percentile in percentiles:
             name = f"{band}_pc_{int(100 * percentile)}"
             yy = da.map_blocks(
-                lambda x: np_percentile(x, percentile, nodata, high=high), 
+                partial(np_percentile, percentile=percentile, nodata=nodata, high=high), 
                 xx_data, 
                 drop_axis=0, 
                 meta=np.array([], dtype=xx.dtype),
-                name=name,
+                name=randomize(name),
```
Kirill888: `randomize` is kinda lazy way, we should not be using it anymore, I know it's all over the place, instead it's best to use `dask.base.tokenize` and give correct set of inputs.

example usage:
https://github.com/opendatacube/odc-tools/blob/29ea012d4bc431ad2bc02b558d3ee19619b74069/libs/algo/odc/algo/_memsink.py#L64

Kirill888: in fact you can share token across sub-tasks, since they have unique names, so maybe something like

```python
tk = tokenize(xx_data, percentiles, nodata, high)
for percentile in percentiles:
      name = f"{band}_pc_{int(100 * percentile)}-{tk}"
```

libs/algo/odc/algo/_percentile.py
```diff
@@ -0,0 +1,78 @@
...
+            name = f"{band}_pc_{int(100 * percentile)}-{tk}"
+            yy = da.map_blocks(
+                partial(np_percentile, percentile=percentile, nodata=nodata), 
+                xx_data, 
+                drop_axis=0, 
+                meta=np.array([], dtype=xx.dtype),
+                name=name,
+            )
+            
+            data_vars[name] = (xx.dims[1:], yy)
```
Kirill888, line 75: pretty sure this key should not have `-{tk}` part.

libs/algo/tests/test_percentile.py
```diff
@@ -0,0 +1,88 @@
...
+    coords = {
+        "x": np.linspace(10, 20, band_1.shape[2]), 
+        "y": np.linspace(0, 5, band_1.shape[1]), 
+        "t": np.linspace(0, 5, band_1.shape[0])
+    }
+
+    data_vars = {"band_1": (("t", "y", "x"), band_1), "band_2": (("t", "y", "x"), band_2)}
+
+    dataset = xr.Dataset(data_vars=data_vars, coords=coords, attrs=attrs)
+    output = xr_percentile(dataset, [0.2, 0.6], nodata).compute()
```
Kirill888, line 85: add test for expected data variable output names.

Kirill888:
all good, just need to make sure output names are without dask tokens in them

libs/algo/odc/algo/_percentile.py
kieranricardo: good catch!

Kirill888 (approved):
:+1:

PR #205, opened 2021-05-25
Refactor save tasks

kieranricardo:
This PR refactors `save` into 3 seperate functions:

- `_get_dss` : find datasets if not supplied
- `save`: saves the datasets
- `_write_info`: writes the csvs

If the `dss` are supplied this has almost the exact same behaviour as previously except that the `cfg` dictionary will not contain a query. This is to support sibling processing which won't have a datacube query.

@Kirill888 I tested this on sandbox by creating a task db with the old code, and then one with the new code with and without dss supplied. And checked the datasets by running:

```python
all(
    all(d1 == d2 for d1, d2 in zip(task_reader_1.datasets(tile), task_reader_2.datasets(tile))) 
    for tile in task_reader_1.all_tiles
)
```

### DIFF
--- a/libs/stats/odc/stats/tasks.py
+++ b/libs/stats/odc/stats/tasks.py
@@ -1,7 +1,10 @@
+from random import random
 from typing import Optional, Tuple, Union, Callable, Any, Dict, List, Iterable, Iterator
+import random
 from types import SimpleNamespace
 from collections import namedtuple
 from datetime import datetime
+from itertools import chain, islice
 import pickle
 import json
 import os
@@ -99,48 +102,19 @@ class SaveTasks:
     def out_path(self, suffix: str) -> str:
         return out_path(suffix, self._output)
 
-    def save(
-        self,
+    def _get_dss(self,
         dc: Datacube,
         product: str,
-        temporal_range: Union[str, DateTimeRange, None] = None,
+        msg: Callable[[str], Any],
+        temporal_range: Optional[DateTimeRange] = None,
         tiles: Optional[TilesRange2d] = None,
-        predicate: Optional[Callable[[Dataset], bool]] = None,
-        msg: Optional[Callable[[str], Any]] = None,
-        debug: bool = False,
-    ) -> bool:
+    ):
         """
-        :param product: Product name to consume
-        :param temporal_range: Optionally  limit query in time
-        :param tiles: Optionally limit query to a range of tiles
-        :param predicate: If supplied filter Datasets as they come in with custom filter, Dataset->Bool
-        :param msg: Observe messages if needed via callback
-        :param debug: Dump some intermediate state to files for debugging
+        This returns a tuple containing:
+        - a generator of datasets
+        - the number of datasets in the generator
+        - a config dictionary containing the product, temporal range, tiles, and the datacube query used
         """
-
-        dt_range = SimpleNamespace(start=None, end=None)
-
-        def _update_start_end(x, out):
-            if out.start is None:
-                out.start = x
-                out.end = x
-            else:
-                out.start = min(out.start, x)
-                out.end = max(out.end, x)
-
-        def persist(ds: Dataset) -> CompressedDataset:
-            _ds = compress_ds(ds)
-            _update_start_end(_ds.time, dt_range)
-            return _ds
-
-        def msg_default(msg):
-            pass
-
-        if msg is None:
-            msg = msg_default
-
-        if isinstance(temporal_range, str):
-            temporal_range = DateTimeRange(temporal_range)
 
         cfg: Dict[str, Any] = dict(
             grid=self._grid, freq=self._frequency,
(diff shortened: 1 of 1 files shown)


libs/stats/odc/stats/tasks.py
```diff
@@ -116,7 +178,34 @@ def save(
         :param predicate: If supplied filter Datasets as they come in with custom filter, Dataset->Bool
         :param msg: Observe messages if needed via callback
         :param debug: Dump some intermediate state to files for debugging
+        :param dss: A generator of datasets to use
+        :param n_dss: The number of datasets in the generator
         """
+        
+        # either product and dc must be set, or dss and n_dss must be
+        valid_args = (product is not None) and (dc is not None)
+        valid_args |= (dss is not None) and (n_dss is not None)
```
Kirill888: I'd say `n_dss` needs to be optional, we won't know how many virtual datasets will come out, we can only know upper bound really, which is `min(count(p) for p in products)`

libs/stats/odc/stats/tasks.py
```diff
@@ -99,6 +99,66 @@ def __init__(
...
+        if "time" in query:
+            dss = chopped_dss(dc, freq="w", **query)
+        else:
+            if len(query) == 1:
+                dss = all_datasets(dc, **query)
+            else:
+                # note: this blocks for large result sets
+                dss = dc.find_datasets_lazy(**query)
+                
+        return dss, cfg, n_dss
```
Kirill888: I would put `n_dss` rigth after `dss`, so maybe `dss, n_dss, cfg`? or `cfg, dss, n_dss`

libs/stats/odc/stats/tasks.py
```diff
@@ -99,6 +99,66 @@ def __init__(
     def out_path(self, suffix: str) -> str:
         return out_path(suffix, self._output)
 
+    def get_dss(self,
```
Kirill888: Is this meant to be internal function?
then start with `_`, `_get_dss`.

Some description of output is needed, particularly expected content of `cfg`

Kirill888: Maybe it's time move that into dataclass of sorts.

libs/stats/odc/stats/tasks.py
kieranricardo: I think we still need an upped bound for `tqdm` to work right?

libs/stats/odc/stats/tasks.py
kieranricardo: I actually had this, I just forgot to commit 🤦

libs/stats/odc/stats/tasks.py
Kirill888: it supports unbounded as well, won't be as useful, but will work

libs/stats/odc/stats/tasks.py
kieranricardo: what do you mean by move into a data class?

Kirill888: nah, let's leave it as is for now.

but what I mean is something like this thing:

https://github.com/opendatacube/odc-tools/blob/488d75f625226d6aefefbaa8f4ba45f64a7692da/libs/stats/odc/stats/model.py#L135-L145

basically just a "struct" that spells out what we expect from "cfg".

libs/stats/odc/stats/tasks.py
```diff
@@ -99,6 +99,71 @@ def __init__(
...
+        This returns a tuple containing:
+        - a generator of datasets
+        - the number of datasets in the generator
+        - a config dictionary containing the product, temporal range, tiles, and the datacube query used
+        """
+        def msg_default(msg):
+            pass
+
+        if msg is None:
+            msg = msg_default
```
Kirill888: This should be in main function too/instead

Kirill888: let's get message fallback sorted and then it's good to go

Kirill888 (approved):

kieranricardo:
@Kirill888 I realised this doesn't quite to what we want - it still relies on querying the database to train the compression dictionary. But I'll can change this to just train on a subset of the input `dss`

Kirill888:
yeah, this was bugging me for a while, currently one needs to supply dictionary at construction of the db file, when really it should be possible to build it while "consuming" dataset stream, probably cache json blobs in RAM until have enough for training then training compression dictionary and then writing out the cache.

kieranricardo:
@Kirill888 I got it working without querying the Datacube  and the tasks I tested still compress to the same size. I also tested this with sibling datasets and it worked woo!

libs/stats/odc/stats/tasks.py
```diff
@@ -142,39 +206,30 @@ def msg_default(msg):
...
-        n_dss = dataset_count(dc.index, **query)
-        if n_dss == 0:
-            msg("Found no datasets to process")
-            return False
-
         msg(f"Processing {n_dss:,d} datasets")
 
         msg("Training compression dictionary")
-        zdict = dictionary_from_product_list(dc, [product], samples_per_product=100)
+        dss, dss_copy = tee(dss)
```
Kirill888: How smart is `tee`? my concern it will keep in RAM copy of unconsumed datasets from `dss_copy`.

probably safer to do islice to list, then copy into shuffle and train, then `chain` original list with the rest of the stream.

Kirill888: the docs seem to agree with my hunch

> This itertool may require significant auxiliary storage (depending on how much temporary data needs to be stored). In general, if one iterator uses most or all of the data before another iterator starts, it is faster to use list() instead of tee().

kieranricardo: good catch! I had wrongly assumed copying an iterator was almost "free"

Kirill888 (approved):
thanks

PR #210, opened 2021-06-04
Post messages in groups

alexgleith:
* Should make re-driving queues faster
* Tests are only sort of sensible, I think `moto` does strange things when working with fake queues

### DIFF
--- a/libs/aws/odc/aws/queue.py
+++ b/libs/aws/odc/aws/queue.py
@@ -8,12 +8,24 @@ def redrive_queue(
     to_queue_name: Optional[str] = None,
     limit: Optional[int] = 0,
     dryrun: bool = False,
+    max_wait: int = 10,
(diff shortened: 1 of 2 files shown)


Kirill888 (approved):

Kirill888:
@alexgleith can you tweak the test settings to not timeout or disable dead-letter test unless somehow requested to run it, waiting 1 minute in a single test for messages to end up in dead letter queue is too much.

Kirill888 (changes requested):
tests need tweaking to not timeout

alexgleith:
Yeah, I know the tests aren't working right.

I need some time to work out how Moto is actually working here. It's an odd use case and hard to test with the mocked environment!

Leave it with me.

Kirill888 (approved):

PR #211, opened 2021-06-07
Add better wild card support

kieranricardo:
This PR adds support for handling arbitrary directory wild cards for depth specified query url. For e.g. queries like the below are now allowed:

- `s3://dea-public-data/baseline/s2a_ard_granule/*-02/*A*/*.yaml`
- `s3://dea-public-data/baseline/s2a_ard_granule/*-02/*A*/eo3-ARD-METADATA.yaml`
- `s3://dea-public-data/baseline/s2a_ard_granule/*-02/*A*/

This works by dynamically cropping the prefix glob (`baseline/s2a_ard_granule/*-02/*A*` in the first example) to match the depth currently being searched in `s3_dir_dir`.

### DIFF
--- a/libs/aio/odc/aio/__init__.py
+++ b/libs/aio/odc/aio/__init__.py
@@ -4,6 +4,7 @@ import botocore
 import aiobotocore
 from aiobotocore.config import AioConfig
 import asyncio
+from fnmatch import fnmatch
 from types import SimpleNamespace
 from typing import Optional, Iterator, Any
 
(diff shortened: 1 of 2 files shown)


Kirill888 (approved):
`dirs_pred` implementation can probably be cleaned up a bit (maybe keep dir globs as a tuple/list rather string), but not ciritical

libs/aio/odc/aio/__init__.py
```diff
@@ -557,12 +558,19 @@ def do_dir_query(qq):
             stream = s3.find(qq.base, pred=lambda o: o.url.endswith(postfix), **kw)
     else:
         # fixed depth query
+        _, prefix = s3_url_parse(glob_pattern)
+        dirs_glob = '/'.join(prefix.split('/')[:-1])
```
Kirill888: since you split it back to list anyway, maybe no point putting it back into a string?

Kirill888: I actually meant `dirs_glob`, no point merging it back into a string if you only use it as a list inside `dirs_pred`
```python
dirs_glob = prefix.split('/')[:-1]
...
    _glob = '/'.join(dirs_glob[:n]) + '/'
```

kieranricardo: yeah good point, I've fixed this up. I'll rebase and merge 👍

Kirill888 (approved):
looks good, merge it

PR #212, opened 2021-06-09
Wofs summaries summary

Kirill888:
plugin to sum wofs annual summaries to compute full history summary.

### DIFF
--- a/libs/stats/odc/stats/_wofs.py
+++ b/libs/stats/odc/stats/_wofs.py
@@ -2,10 +2,12 @@
 Wofs Summary
 """
 from typing import Optional, Tuple
+import numpy as np
 import xarray as xr
 from odc.stats.model import Task
 from odc.algo.io import load_with_native_transform
 from odc.algo import safe_div, apply_numexpr, keep_good_only
+from odc.algo.io import dc_load
 from .model import StatsPluginInterface
 from . import _plugins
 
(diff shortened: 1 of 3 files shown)


libs/stats/odc/stats/_wofs.py
```diff
@@ -114,11 +117,98 @@ def reduce(self, xx: xr.Dataset) -> xr.Dataset:
...
+    @property
+    def measurements(self) -> Tuple[str, ...]:
+        return ("count_wet", "count_clear", "frequency")
+
+    def input_data(self, task: Task) -> xr.Dataset:
+        return dc_load(
+            task.datasets,
+            measurements=["count_wet", "count_clear"],
+            geobox=task.geobox,
+            chunks={},
```
omad, line 156: What does `chunks={}` do? Activates dask for parallelise loading, but forces it all into a chunk?

omad (approved):
Looks good to me. I had to look at it for way too long to work out what is was doing and why, so I'd prefer extra documentation. I'll make a PR and you can check my understanding.

Cheers

libs/stats/odc/stats/_wofs.py
Kirill888: yes single chunk along Y,X but chunk size of 1 for time.

Kirill888: so equivalent to `{'x': -1, 'y': -1', 'time': 1}`.

PR #216, opened 2021-06-10
Add documentation on WO Stats

omad:
Also cleans up some warnings that PyCharm pointed out.

### DIFF
--- a/libs/algo/odc/algo/io.py
+++ b/libs/algo/odc/algo/io.py
@@ -173,8 +173,8 @@ def _load_with_native_transform_1(
 
 
 def load_with_native_transform(
-    dss: List[Dataset],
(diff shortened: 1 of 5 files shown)


Kirill888 (approved):
awesome, thanks

libs/algo/odc/algo/io.py
```diff
@@ -173,8 +173,8 @@ def _load_with_native_transform_1(
 
 
 def load_with_native_transform(
-    dss: List[Dataset],
-    bands: Tuple[str, ...],
+    dss: Sequence[Dataset],
+    bands: Sequence[str, ...],
```
Kirill888:
```suggestion
    bands: Sequence[str],
```

PR #218, opened 2021-06-15
Support proj:transform and proj:shape in properties + null datetime

### LINKED ISSUE #217: Support `proj:shape` and `proj:transform` in properties
"change the stac transform to handle the proj:shape and `proj:transform` being in the properties rather than per-band."

Needed to fix some metadata reading for NBH.

Ping @alexgleith

MatthewJA:
Resolves #217 and also a bug where datetime: null was unsupported [despite being allowed](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#properties-object).

### DIFF
--- a/libs/index/odc/index/stac.py
+++ b/libs/index/odc/index/stac.py
@@ -78,7 +78,8 @@ def _find_self_href(item: Document) -> str:
 
 
 def _get_stac_bands(
-    item: Document, default_grid: str, relative: bool = False
+    item: Document, default_grid: str, relative: bool = False,
+    proj_shape: str = None, proj_transform: str = None,
 ) -> Tuple[Document, Document]:
     bands = {}
     grids = {}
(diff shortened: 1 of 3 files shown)


alexgleith (approved):
Pure awesome

libs/index/odc/index/stac.py
```diff
@@ -166,7 +172,7 @@ def _get_stac_properties_lineage(input_stac: Document) -> Tuple[Document, Any]:
         MAPPING_STAC_TO_EO3.get(key, key): _convert_value_to_eo3_type(key, val)
         for key, val in properties.items()
     }
-    if prop.get("odc:processing_datetime") is None:
+    if prop.get("odc:processing_datetime") is None and properties["datetime"] is not None:
```
Kirill888, line 175: this will raise `KeyError` if `datetime` field is not populated

MatthewJA: Thanks for fixing this @alexgleith. FWIW I believe STAC _requires_ a datetime field, which is then allowed to be null... good to not crash out here anyway though.

PR #219, opened 2021-06-15
Fix up bug with stac.py datetime, format code

alexgleith:
* Minor formatting
* Fix issue with stac.py to do .get() instead of assuming property exists.

### DIFF
--- a/libs/index/odc/index/stac.py
+++ b/libs/index/odc/index/stac.py
@@ -78,8 +78,11 @@ def _find_self_href(item: Document) -> str:
 
 
 def _get_stac_bands(
-    item: Document, default_grid: str, relative: bool = False,
-    proj_shape: str = None, proj_transform: str = None,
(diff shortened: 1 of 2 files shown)


libs/index/odc/index/stac.py
```diff
@@ -78,8 +78,11 @@ def _find_self_href(item: Document) -> str:
 
 
 def _get_stac_bands(
-    item: Document, default_grid: str, relative: bool = False,
-    proj_shape: str = None, proj_transform: str = None,
+    item: Document,
+    default_grid: str,
+    relative: bool = False,
+    proj_shape: str = None,
```
Kirill888: can't assign `None` into a `str`, should be `Optional[str]`

alexgleith: Good point.

Kirill888 (approved):

PR #220, opened 2021-06-20
Stats fc percentile

kieranricardo:
This PR adds a stats plugin class to orchestrate fc percentiles.

### DIFF
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -627,8 +627,6 @@ def _da_fuse_with_custom_op(xx: da.Array, op, name="fuse") -> da.Array:
     Out[0, y, x] = op(In[0:1, y, x], In[1:2, y, x], In[2:3, y, x]...)
 
     """
-    print("Bad array:", xx)
-    print("Type:", type(xx))
     can_do_flat = all([ch == 1 for ch in xx.chunks[0]])
     if not can_do_flat:
         slices = [xx[i : i + 1] for i in range(xx.shape[0])]
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -731,7 +731,7 @@ def _first_np(*aa: np.ndarray) -> np.ndarray:
     return aa[0]
 
 
-def choose_first(x: xr.DataArray, nodata=None) -> xr.DataArray:
+def choose_first(x: xr.DataArray) -> xr.DataArray:
     """
     ``Out[0, y, x] = In[0, y, x]``
 
@@ -743,12 +743,10 @@ def choose_first(x: xr.DataArray, nodata=None) -> xr.DataArray:
     )
 
 
-def _first_fuser(xx, **kw):
-    """
-    meant to be called by `xx.groupby(..).map(_nodata_fuser)`
-    """
+def _first_fuser(xx):
+
     if isinstance(xx, xr.Dataset):
-        return xx.map(choose_first, **kw)
+        return xx.map(choose_first)
     if xx.shape[0] <= 1:
         return xx
-    return choose_first(xx, **kw)
+    return choose_first(xx)
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -53,7 +53,7 @@ def keep_good_only(x, where, inplace=False, nodata=None):
     """
     if isinstance(x, xr.Dataset):
         return x.apply(
-            lambda x: keep_good_only(x, where, inplace=inplace), keep_attrs=True
+            lambda x: keep_good_only(x, where, inplace=inplace, nodata=nodata), keep_attrs=True
         )
 
     assert x.shape == where.shape
@@ -627,9 +627,11 @@ def _da_fuse_with_custom_op(xx: da.Array, op, name="fuse") -> da.Array:
     Out[0, y, x] = op(In[0:1, y, x], In[1:2, y, x], In[2:3, y, x]...)
 
     """
+    print("Bad array:", xx)
+    print("Type:", type(xx))
     can_do_flat = all([ch == 1 for ch in xx.chunks[0]])
     if not can_do_flat:
-        slices = [xx.data[i : i + 1] for i in range(xx.shape[0])]
+        slices = [xx[i : i + 1] for i in range(xx.shape[0])]
         return da.map_blocks(op, *slices, name=name)
 
     chunks, shapes = _get_chunks_asarray(xx)
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -725,28 +725,3 @@ def _nodata_fuser(xx, **kw):
     if xx.shape[0] <= 1:
         return xx
     return choose_first_valid(xx, **kw)
-
-
-def _first_np(*aa: np.ndarray) -> np.ndarray:
-    return aa[0]
-
-
-def choose_first(x: xr.DataArray) -> xr.DataArray:
-    """
-    ``Out[0, y, x] = In[0, y, x]``
-
-    Expects data in ``_, y, x`` order. Works on Dask inputs too.
-    """
-
-    return _fuse_with_custom_op(
-        x, _first_np, name="choose_first"
-    )
-
-
-def _first_fuser(xx):
-
-    if isinstance(xx, xr.Dataset):
-        return xx.map(choose_first)
-    if xx.shape[0] <= 1:
-        return xx
-    return choose_first(xx)
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -725,3 +725,30 @@ def _nodata_fuser(xx, **kw):
     if xx.shape[0] <= 1:
         return xx
     return choose_first_valid(xx, **kw)
+
+
+def _first_np(*aa: np.ndarray) -> np.ndarray:
+    return aa[0]
+
+
+def choose_first(x: xr.DataArray, nodata=None) -> xr.DataArray:
+    """
+    ``Out[0, y, x] = In[0, y, x]``
+
+    Expects data in ``_, y, x`` order. Works on Dask inputs too.
+    """
+
+    return _fuse_with_custom_op(
+        x, _first_np, name="choose_first"
+    )
+
+
+def _first_fuser(xx, **kw):
+    """
+    meant to be called by `xx.groupby(..).map(_nodata_fuser)`
+    """
+    if isinstance(xx, xr.Dataset):
+        return xx.map(choose_first, **kw)
+    if xx.shape[0] <= 1:
+        return xx
+    return choose_first(xx, **kw)
--- a/libs/stats/odc/stats/_fc_percentiles.py
+++ b/libs/stats/odc/stats/_fc_percentiles.py
@@ -64,7 +64,7 @@ class StatsFCP(StatsPluginInterface):
 
         wet = xx.wet
         dry = xx.dry
-        xx = _xr_fuse(xx.drop('wet'), partial(_first_valid_np, nodata=NODATA), '')
+        xx = _xr_fuse(xx.drop_vars(["wet", "dry"]), partial(_first_valid_np, nodata=NODATA), '')
         xx["wet"] = _xr_fuse(wet, _fuse_or_np, wet.name)
         xx["dry"] = _xr_fuse(dry, _fuse_and_np, wet.name)
         return xx
--- a/libs/stats/odc/stats/_fc_percentiles.py
+++ b/libs/stats/odc/stats/_fc_percentiles.py
@@ -88,10 +88,9 @@ class StatsFCP(StatsPluginInterface):
 
     @staticmethod
     def reduce(xx: xr.Dataset) -> xr.Dataset: 
-        # not 255 & True => 1
-        # not 255 & False => 1
-        # 255 & False => 2  
-        # 255 + True => 3
+        # (!all_bands_valid) & is_ever_wet => 0
+        # (!all_bands_valid) & (!is_ever_wet) => 1
+        # all_bands_valid => 2  
 
         mask = xx["dry"]
         wet = xx["wet"]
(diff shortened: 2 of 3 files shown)


kieranricardo:
@Kirill888 @omad wait sorry I think I spotted a bug - this one isn't quite ready for review

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,81 @@
...
+
+        1. Loads all the fc and WOfS bands
+        2. Extracts the clear dry and clear wet flags from WOfS
+        3. Drops the WOfS band
+        4. Stores the clear wet flags for QA later on 
+        5. Masks out all pixels that are not clear and dry to a nodata value of 255
+        6. Discards the clear dry flags
+        """
+        # use the dry flag to indicate good measurementd
+        dry = xx.water == 0
```
Kirill888: Pretty sure we need to ignore high-slope bit, as in we need to ignore wet observations on high-slope terrain as well as on flat terrain, and equally keep "dry" observations even on high-slope.

So you need to pull-down slope bit before computing dry/wet counts.

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,81 @@
...
+
+        chunks = {"y": -1, "x": -1}
+
+        xx = load_with_native_transform(
+            task.datasets,
+            bands=["water", "pv", "bs", "npv", "ue"],
+            geobox=task.geobox,
+            native_transform=self._native_tr,
+            fuser=None,
+            groupby=None,
```
Kirill888: grouping by solar day and fusing is required, otherwise you end up with double counting observations in North-South overlap.

kieranricardo: I have a question re. the `_nodata_fuser`: does it consider the bands seperately?

It looks like it does, and if so maybe something like this could happen:

(two copies of the same pixel on the same day)
```
pv    255    80
dry   True    False
```

Gets fused to:

```
pv    80
dry    True
```

causing us to incorrectly use a wet measurement. If that makes sense?

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,93 @@
...
+        6. Discards the clear dry flags
+        """
+
+        # set terrain flag to zero
+        water = da.bitwise_and(xx["water"], 0b11101111)
+        xx = xx.drop_vars(["water"])
+
+        # equivalent to water & 0b11101111
+        expression = "((water << 4) >> 4) + ((water >> 5) << 5)"
+        water = apply_numexpr(expression, xx, dtype="bool")
```
Kirill888: but you already dropped `water` from `xx` above?

libs/stats/odc/stats/_fc_percentiles.py
Kirill888: custom fuser is given a dataset object filled with dask arrays, it's up to you which bands to fuse with which fuser, so you want to use _nodata_fuser for data bands and `and` or `or`  for boolean bands.
keep in mind that fusing happens several times: once for every native projection and then again after reprojection.

libs/algo/odc/algo/_masking.py
```diff
@@ -725,3 +725,28 @@ def _nodata_fuser(xx, **kw):
...
+    )
+
+
+def _first_fuser(xx):
+
+    if isinstance(xx, xr.Dataset):
+        return xx.map(choose_first)
+    if xx.shape[0] <= 1:
+        return xx
+    return choose_first(xx)
```
Kirill888: there is already choose_first_valid, just picking "first" regardless of `nodata` doesn't make sense.

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: oh woops I pushed the wrong thing

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: oh wait I just realised that the `fuser` is called after the `native_transform` so my comment above isn't an issue. All the measurements that aren't dry and clear will be set to `nodata` before the fuser anyway so I can just use the `choose_first_valid` for the `fc` bands.

And I think using `or` for the `wet` band is the right choice here.

How do I apply different fusers to different bands?

Kirill888: fuser is applied in native projection as well as in final projection

Kirill888: but yes, after native transform was called

Kirill888: custom fuser is given an `xr.Dataset` object that was returned from native_transform, you can slice into it with data band names, apply nodata fuser to those, then add masking bands that are fused with `or` afterwards.

kieranricardo: oh I think I've got it - I just run load multiple times

kieranricardo:
@Kirill888 I think this is ready for a re-review when the get the chance :)

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,110 @@
...
+        xx["wet"] = load_with_native_transform(
+            task.datasets,
+            bands=["water"],
+            geobox=task.geobox,
+            native_transform=self._native_tr_wet,
+            fuser=_or_fuser,
+            groupby="solar_day",
+            resampling=self.resampling,
+            chunks=chunks,
+        )
```
Kirill888: This will probably result in loading `water` band twice.

To avoid that you need to return `wet` from `input_data`, then you'll need to implement custom fuser that will fuse data bands differently from the `wet` band.

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,110 @@
...
+
+        return xx
+
+    def reduce(self, xx: xr.Dataset) -> xr.Dataset:        
+        yy = xr_percentile(xx.drop_vars(["wet"]), [0.1, 0.5, 0.9], nodata=255)
+        yy["is_ever_wet"] = _or_fuser(xx["wet"]).squeeze(xx["wet"].dims[0], drop=True)
```
Kirill888: I don't think `is_ever_wet` is what we need on the output. But rather an enumerated type with value like
- valid (at least 3 valid dry obs)
- none (no obs at all)
- not sure ([1-3] obs and all dry)
- sure not ([1-3] obs some wet)

kieranricardo: I thought the `is_ever_wet` band could be combined with the `fc_percentiles` bands to derive this info:

- <= `100` data value -> valid
- `255` data value + `is_ever_wet=False` -> not sure
- `255` data value + `is_ever_wet=True` -> sure not

But I can add this explicitly as an enum type 👍

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: yeah it does unfortunately, I could do something like make a custom fuser that calls the `_or_fuser` if a boolean type is detected and calls `_nodata_fuser` otherwise

Kirill888: no need to "detect", you know exactly which bands to expect on input and which bands need to be fused with what logic.

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -89,9 +89,23 @@ def input_data(self, task: Task) -> xr.Dataset:
...
+        # not 255 & True => 1
+        # not 255 & False => 1
+        # 255 & False => 2  
+        # 255 + True => 3
+
         yy = xr_percentile(xx.drop_vars(["wet"]), [0.1, 0.5, 0.9], nodata=255)
-        yy["is_ever_wet"] = _or_fuser(xx["wet"]).squeeze(xx["wet"].dims[0], drop=True)
+        is_ever_wet = _or_fuser(xx["wet"]).squeeze(xx["wet"].dims[0], drop=True)
+
+        band, *bands = [band for band in yy.data_vars.keys() if band != "wet"]
```
Kirill888: `yy` doesn't have `wet` band anyway

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -89,9 +89,23 @@ def input_data(self, task: Task) -> xr.Dataset:
...
+        all_bands_valid = yy[band] != 255
```
Kirill888: let's give `255` label `nodata`

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -89,9 +89,23 @@ def input_data(self, task: Task) -> xr.Dataset:
...
+        for band in bands:
+            all_bands_valid &= yy[band] != 255
```
Kirill888: `&=` <-- this doesn't work for dask arrays

libs/stats/odc/stats/_fc_percentiles.py
Kirill888: Probably easier approach is something like this:

```python
    some_invalid = (yy.to_array() == yy.nodata).sum(axis=0, dtype='int8') > 0
```

- turn band into dimension
- turn array into boolean array T if `nodata`
- sum boolean array along band dimension
- if result more than 0 there is at least one band with nodata value

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,115 @@
...
+        data_bands = [band for band in xx.data_vars.keys() if band != "wet"]
+        for band in data_bands:
+            variables[band] = _fuse_with_custom_op(xx[band], partial(_first_valid_np, nodata=255))
+        variables["wet"] = _fuse_with_custom_op(xx["wet"], _fuse_or_np)
+        
+        for k, v in variables.items():
+            v._copy_attrs_from(xx.data_vars[k])
+
+        yy = type(xx)(variables, attrs=xx.attrs)
+        return yy
```
Kirill888: This can be achieved with this (probably)

```python
wet = xx.wet
xx = _xr_fuse(xx.drop('wet'), _first_valid_np, '')
xx["wet"] = _xr_fuse(wet, _fuse_or_np, wet.name)
```

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(
+        self,
+        resampling: str = "bilinear",
+    ):
+        self.resampling = resampling
+
+    @property
+    def measurements(self) -> Tuple[str, ...]:
+        _measurments = [f"{b}_pc_{p}" for b, p in product(["pv", "bs", "npv", "ue"], ["10", "50", "90"])]
+        _measurments.append("is_ever_wet")
```
Kirill888: didn't you rename it into `qa`

Kirill888: I don't know what band is needed, probably check with Mathew, it looks like `is_ever_wet` boolean classification was one of the proposed mechanisms

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,115 @@
...
+        6. Discards the clear dry flags
+        """
+
+        # set terrain flag to zero
+        water = da.bitwise_and(xx["water"], 0b11101111)
+        xx = xx.drop_vars(["water"])
+
+        # use the dry flag to indicate good measurements
+        dry = water == 0
+        xx = keep_good_only(xx, dry, nodata=255)
```
Kirill888: need to move this out of native transform, we want to discard observation that have wet classification in any of the variants.

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: I tried something like that initially and it didn't work - I think bc of the dimension changes.

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: my tests are using dask arrays and seem to pass - maybe its a new feature? I'll check the docs

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: it probably doesn't make much difference either way but the current thing doesn't do any type casting - so I'll leave as is for now

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: I found it in the docs https://docs.dask.org/en/latest/array-api.html#dask.array.logical_and so I guess they must've added it 🤷

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: oh wait I totally misread this sorry - this absolutely works and is way better!

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: I had a chat w/ Mathew and an enum band would be preferred. I've chucked it up on slack to discuss w/ Mathew and Emma exactly what values should be included

kieranricardo: they wanted the enum order reversed but other than that all good!

kieranricardo:
@Kirill888 I think its ready for another review - thanks for your patience!

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @staticmethod
+    def reduce(xx: xr.Dataset) -> xr.Dataset: 
+        # (!all_bands_valid) & is_ever_wet => 0
+        # (!all_bands_valid) & (!is_ever_wet) => 1
+        # all_bands_valid => 2  
+
+        mask = xx["dry"]
+        wet = xx["wet"]
+        xx.drop_vars(["dry", "wet"])
```
Kirill888: drop_vars does not modify `xx` it returns a copy without supplied bands

kieranricardo: @Kirill888 yep good catch! I added a test in to catch this and fixed it

Kirill888 (approved):

Kirill888:
@kieranricardo in the future please simplify history before merging changes like this

Kirill888:
I guess merge with squash is fine, but I feel like there were several distinct changes in this PR, there was some bug fix in there wasn't it?

kieranricardo:
@Kirill888 yeah there were 2 tiny bug fixes in there - I guess they should've been seperate PRs but they seemed so tiny

what's the difference in git history w/ squash and merge vs rebasing?

Kirill888:
squash option on github produce one commit, which is fine if there were iteration commits but there is only ONE common change. rebasing locally gives you more freedom to cleanup/re-order/merge related commits and end up with just a handful of commits

PR #226, opened 2021-06-24
Add accessory files to the resulting ODC dictionary

alexgleith:
This fixes a bug where accessory files weren't passed through the STAC to ODC transform.

Implementing this means that indexed datasets will have accessories like thumbnails available for clicking in Explorer.

### DIFF
--- a/libs/index/odc/index/stac.py
+++ b/libs/index/odc/index/stac.py
@@ -86,14 +86,24 @@ def _get_stac_bands(
 ) -> Tuple[Document, Document]:
     bands = {}
     grids = {}
+    accessories = {}
+
     assets = item.get("assets", {})
 
+    def _get_path(asset):
+        path = asset["href"]
+        if relative:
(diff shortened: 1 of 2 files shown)


libs/index/odc/index/stac.py
```diff
@@ -86,14 +86,24 @@ def _get_stac_bands(
...
     bands = {}
     grids = {}
+    accessories = {}
+
     assets = item.get("assets", {})
 
+    def _get_path(asset):
+        path = asset["href"]
+        if relative:
+            path = Path(path).name
```
jeremyh, line 96: We haven't seen any datasets with subfolders, have we?

jeremyh (approved):

libs/index/odc/index/stac.py
alexgleith: Not in the tests...

PR #228, opened 2021-06-25
Support two products w/ fusing for cli save tasks

kieranricardo:
This PR adds supports for fusing

### DIFF
--- a/libs/stats/odc/stats/_cli_save_tasks.py
+++ b/libs/stats/odc/stats/_cli_save_tasks.py
@@ -178,7 +178,7 @@ def _parse_products(dc, products, temporal_range):
     paired_dss = groupby(dss, key=lambda ds: (ds.center_time, ds.metadata.region_code))
     
     error_logger = ErrorLogger(products)
-    paired_dss = _filter_bad_with_logging(paired_dss, error_logger)
+    paired_dss = error_logger.filter(paired_dss)
     n_dss = max(dataset_count(dc.index, time=query["time"], product=product) for product in products)
 
     products = [dc.index.products.get_by_name(product) for product in products]
@@ -188,15 +188,6 @@ def _parse_products(dc, products, temporal_range):
     product = fused_product.name
 
     return dss, n_dss, product, error_logger
-
-
-def _filter_bad_with_logging(groups, error_logger):
-    for _, ds_group in groups:
-        ds_group = tuple(ds_group)
-        if not error_logger.check(ds_group):
-            error_logger.append(ds_group)
-        else:
-            yield ds_group
 
 
 class ErrorLogger:
@@ -213,3 +204,12 @@ class ErrorLogger:
     
     def check(self, ds_group):
         return len(ds_group) == len(self.products)
+
+    def filter(self, groups):
+        for _, ds_group in groups:
+            ds_group = tuple(ds_group)
+            if not self.check(ds_group):
+                self.append(ds_group)
+            else:
+                yield ds_group
+        
--- a/libs/stats/odc/stats/_cli_save_tasks.py
+++ b/libs/stats/odc/stats/_cli_save_tasks.py
@@ -114,8 +114,13 @@ def save_tasks(
 
     dc = Datacube(env=env)
     products = products.split("+")
-    dss, n_dss, product = _parse_products(dc, products, temporal_range)
-
+    if len(products) == 1: 
+        product = products[0]
+        dss = None
+        n_dss = None
+    else:
+        dss, n_dss, product = _parse_products(dc, products, temporal_range)
+        
     if output == "":
         if temporal_range is not None:
             output = f"{product}_{temporal_range.short}.db"
@@ -140,9 +145,6 @@ def save_tasks(
     if gqa is not None:
         predicate = gqa_predicate
 
-    if len(products) != 1: 
-        dc = None
-        product = None
     try:
         ok = tasks.save(
             dc,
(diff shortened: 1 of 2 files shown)


libs/stats/odc/stats/_cli_save_tasks.py
```diff
@@ -151,3 +162,38 @@ def gqa_predicate(ds):
...
         # exit with error code, failure message was already printed
         sys.exit(3)
+
+
+def _parse_products(dc, products, temporal_range):
+    if len(products) == 1:
+        product = products[0]
+        dss = None
+        n_dss = None
+    elif len(products) == 2:
```
Kirill888: I think this check should be in main, it kinda already there, and this should just work on a sequence of products, even if that sequence has one element, it just won't be called with anything less than 2

Kirill888: sorry fusing, should accept a sequence of exactly 2 elements for now at least until we generalize to more than that.

libs/stats/odc/stats/_cli_save_tasks.py
```diff
@@ -151,3 +162,38 @@ def gqa_predicate(ds):
...
+        
+        query = {"product": products}
+        query.update(
+                temporal_range.dc_query(pad=0.6)
+            ) 
+        dss = ordered_dss(
+            dc, key=lambda ds: (ds.center_time, ds.metadata.region_code), **query
+        )
+        paired_dss = groupby(dss, key=lambda ds: (ds.center_time, ds.metadata.region_code))
+        paired_dss = [list(t[1]) for t in paired_dss]
```
Kirill888: this will run out of RAM on large enough queries, it really needs to be a streaming operation

libs/stats/odc/stats/_cli_save_tasks.py
kieranricardo: yeah it will only accept two but its probably better that we let the fusing functions throw the errors.

libs/stats/odc/stats/_cli_save_tasks.py
```diff
@@ -169,26 +169,17 @@ def gqa_predicate(ds):
...
     paired_dss = groupby(dss, key=lambda ds: (ds.center_time, ds.metadata.region_code))
-    paired_dss = [list(t[1]) for t in paired_dss]
-    bad_dss = [x for x in paired_dss if len(x) != 2]
-    paired_dss = [x for x in paired_dss if len(x) == 2]
-    
-    for x in groupby(bad_dss, key=lambda ds: ds[0].type.name):
-        name, _iter = x
-        print(f"Warning: product {name} has {len(list(_iter))} unpaired datasets")
-    
+    paired_dss = (tuple(t[1]) for t in paired_dss)
```
Kirill888: using unpacking syntax here is preferable to indexing, something like
```python
paired_dss = (tuple(group) for _, group in paired_dss)
```

in fact you can combine lines above and below as

```python
paired_dss = (tuple(group) for _, group in paired_dss if len(group) == expected_group_size)
```

libs/stats/odc/stats/_cli_save_tasks.py
```diff
@@ -169,26 +169,17 @@ def gqa_predicate(ds):
...
+    paired_dss = (x for x in paired_dss if len(x) == len(products))
```
Kirill888: so with this logic we have no record of mismatched datasets

kieranricardo: We have a record of how many we expect (n_dss) and the number that are actually paired. So we have the number of datasets that are missing but no other info.

I could do something like:
- split into paired and bad dss tiers with tee
- Iterate over paired first
- Iterate over bad

But this could still yield OOM if the number of bad datasets is large. What do you think?

Kirill888: I would do filtering iterator that keeps count and maybe logs things

```python
def _filter_bad_with_logging(groups, error_state_object):
    for g in groups:
       if group_no_good(g):
          error_state_object.bad_count += 1
          log_count_whatever(g, error_state_object) # update whatever counts in error_state_object and maybe log issues
       else:
          yield g
#...
errors = ... # SimpleNamescpase(...) or some dataclass or whatever
clean_stream = _filter_bad_with_logging(dirty_stream, errors)
for _ in clean_stream:
     do_stuff(...)
report_error_summaries(errors)
```

libs/stats/odc/stats/_cli_save_tasks.py
```diff
@@ -143,11 +154,62 @@ def gqa_predicate(ds):
...
+def _parse_products(dc, products, temporal_range):
+        
+    query = {"product": products}
+    query.update(temporal_range.dc_query(pad=0.6)) 
+    dss = ordered_dss(dc, key=lambda ds: (ds.center_time, ds.metadata.region_code), **query)
+    paired_dss = groupby(dss, key=lambda ds: (ds.center_time, ds.metadata.region_code))
+    
+    error_logger = ErrorLogger(products)
+    paired_dss = error_logger.filter(paired_dss)
+    n_dss = max(dataset_count(dc.index, time=query["time"], product=product) for product in products)
```
Kirill888: shouldn't that be `min(..)`?

kieranricardo: I originally had max to give us more info about the number of unpaired datasets but this is unnecessary now - so yeah `min` will give us a tighter upper bound on the number of paired datasets.

Kirill888 (approved):

PR #238, opened 2021-06-30
Update s3-to-dc add update-if-exists and share code with sqs-to-dc

alexgleith:
* Update the s3-to-dc so that it can update-if-exists
* Share more code between sqs-to-dc and s3-to-dc
* Add some better docs ;-)

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
@@ -8,9 +8,9 @@ from typing import Tuple
 
 import click
 from datacube import Datacube
-from datacube.utils import changes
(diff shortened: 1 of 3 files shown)


apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -91,6 +77,12 @@ def dump_to_odc(
     default=False,
     help="If set, update instead of add datasets",
 )
+@click.option(
```
carioca-au, line 80: This is a nice one!!!

carioca-au (approved):
@alexgleith Thank you for better readmes and the **update if exists** function

PR #241, opened 2021-07-02
Add the filedb S3 uri to the SQS message

supermarkion:
In the previous solution, the filedb S3 uri saved in K8s job config as a hardcode and global value. So the TaskRunner can get filedb (then init TaskReader instance) without consume any SQS message.

In order to remove that hard-code value from K8s job config, I defer the TaskReader initialization in the TaskRunner. Once the application begin to consume the message, then grab the filedb S3 uri then initalize the TaskReader part. What is more, add an if-else statement to check the filedb S3 file had been download to local or not before.


libs/stats/odc/stats/tasks.py
```diff
@@ -451,7 +506,28 @@ def stream_from_sqs(
             sqs_queue = get_queue(sqs_queue)
 
         for msg in get_messages(sqs_queue, visibility_timeout=visibility_timeout, **kw):
-            # TODO: switch to JSON for SQS message body
             token = SQSWorkToken(msg, visibility_timeout)
-            tidx = parse_task(msg.body)
+            tidx, filedb = parse_sqs(msg.body)
+
+            # avoid the download and update again, how ever, we have to setup an exception to handle the unit test filedb
+            if filedb.split('/')[-1] != 'test_tiles.db':
```
kieranricardo: this might be cleaner if we check whether the  path is s3 or not

Will this work?

```suggestion
            if filedb[:3] == 's3:'
```

kieranricardo (commented):
@supermarkion I think it would be cleaner if the initialising from SQS work was done in the beginning. You could do something like this in `__init__`:

```python
sqs_queue = get_queue(sqs_queue)
msg = next(get_messages(sqs_queue, visibility_timeout=visibility_timeout, **kw))
token = SQSWorkToken(msg, visibility_timeout)
tidx, filedb = parse_sqs(msg.body)
bucket, key = s3_url_parse(filedb)
local_cache_file = key.split("/")[-1]
self.init_from_sqs(filedb)
..... # other stuff

token.cancel()
```

This reads the first message but doesn't delete it from the queue. 

What do you think?

supermarkion:
> @supermarkion I think it would be cleaner if the initialising from SQS work was done in the beginning. You could do something like this in `__init__`:
> 
> ```python
> sqs_queue = get_queue(sqs_queue)
> msg = next(get_messages(sqs_queue, visibility_timeout=visibility_timeout, **kw))
> token = SQSWorkToken(msg, visibility_timeout)
> tidx, filedb = parse_sqs(msg.body)
> bucket, key = s3_url_parse(filedb)
> local_cache_file = key.split("/")[-1]
> self.init_from_sqs(filedb)
> ..... # other stuff
> 
> token.cancel()
> ```
> 
> This reads the first message but doesn't delete it from the queue.
> 
> What do you think?

That could be one way. The assumption of this design is 'all messages are come from same config/running'. I believe we have no plan to support streaming processing in odc-stats?

supermarkion:
> > @supermarkion I think it would be cleaner if the initialising from SQS work was done in the beginning. You could do something like this in `__init__`:
> > ```python
> > sqs_queue = get_queue(sqs_queue)
> > msg = next(get_messages(sqs_queue, visibility_timeout=visibility_timeout, **kw))
> > token = SQSWorkToken(msg, visibility_timeout)
> > tidx, filedb = parse_sqs(msg.body)
> > bucket, key = s3_url_parse(filedb)
> > local_cache_file = key.split("/")[-1]
> > self.init_from_sqs(filedb)
> > ..... # other stuff
> > 
> > token.cancel()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This reads the first message but doesn't delete it from the queue.
> > What do you think?
> 
> That could be one way. The assumption of this design is 'all messages are come from same config/running'. I believe we have no plan to support streaming processing in odc-stats?

I just found the AWS SQS does not support the peek feature. 

- It can: grab first message and set short timeout (but waste one retry).

- It cannot read the first message but not delete it from queue.

libs/stats/odc/stats/tasks.py
supermarkion: I changed it as: 
`if urlparse(filedb).scheme == 's3':`

kieranricardo:
hey @supermarkion! just to clarify, does each SQS message contain the `s3` url of a different dataset cache file?

kieranricardo:
oh actually sorry scratch my last message, I just realised they are all the same DB file

libs/stats/odc/stats/tasks.py
```diff
@@ -459,9 +515,29 @@ def stream_from_sqs(
...
+            tidx, filedb = parse_sqs(msg.body)
+
+            # avoid the download and update again, how ever, we have to setup an exception to handle the unit test filedb
+            if urlparse(filedb).scheme == 's3':
+                bucket, key = s3_url_parse(filedb)
+            else: # if it is the test_tiles.db, we load it from local
+                key = filedb 
+            local_cache_file = key.split("/")[-1]
+            if not os.path.isfile(local_cache_file):  # use the download filedb from S3 as the init context flag
+                self.init_from_sqs(filedb)
```
kieranricardo, line 549: does this read the cache file for each message?

libs/stats/odc/stats/tasks.py
```diff
@@ -459,9 +515,29 @@ def stream_from_sqs(
...
+                # first time to access the filedb, then it can do the resolution check
+                if self.resolution is not None:
+                    _log.info(f"Changing resolution to {resolution[0], resolution[1]}")
```
kieranricardo: Not sure if I'm missing something, but it looks like `resolution` and `cfg` aren't defined here.

supermarkion:
> hey @supermarkion! just to clarify, does each SQS message contain the `s3` url of a different dataset cache file?

the assumption is: each odc-stats run will has their own SQS. So each SQS message (from each queue) should have the same s3 db uri.

libs/stats/odc/stats/tasks.py
supermarkion: The control here is: useing cache db_file as 'flag'. 

- if not in local filesystem, we download it from S3, and read it.
- if in local filesystem, we assume the current worker already read it (then skip cache file loading).

Of course, someone may argue 
> if download the s3 file, but application crash. Then the 'cache db_file' flag does not work.

However, consider we are using the K8s pod/docker containter to run it. The application crash will also clean up the application file system. So use local db_file as flag could be safe.

libs/stats/odc/stats/tasks.py
supermarkion: You are right, fixed it here: https://github.com/opendatacube/odc-tools/blob/feature/move-db-file-sqs/libs/stats/odc/stats/tasks.py#L390

supermarkion: the `cfg` is here: https://github.com/opendatacube/odc-tools/blob/feature/move-db-file-sqs/libs/stats/odc/stats/tasks.py#L439

libs/stats/odc/stats/tasks.py
```diff
@@ -459,9 +519,29 @@ def stream_from_sqs(
...
+            # avoid the download and update again, how ever, we have to setup an exception to handle the unit test filedb
+            if urlparse(filedb).scheme == 's3':
+                bucket, key = s3_url_parse(filedb)
+            else: # if it is the test_tiles.db, we load it from local
+                key = filedb 
+            local_cache_file = key.split("/")[-1]
+            if not os.path.isfile(local_cache_file):  # use the download filedb from S3 as the init context flag
+                self.init_from_sqs(filedb)
+                # first time to access the filedb, then it can do the resolution check
+                if self.resolution is not None:
```
kieranricardo: One last comment, I think this could moved into `__init__from_sqs`

kieranricardo:
@supermarkion I think this is looking good! just one last comment and then I think this is good to go

libs/stats/odc/stats/tasks.py
supermarkion: Gave me moment to change it.

supermarkion:
> @supermarkion I think this is looking good! just one last comment and then I think this is good to go

Already moved the resolution check to init method.

kieranricardo (approved):
LGTM!

Kirill888:
@supermarkion for future PRs please use rebase when bringing your branch inline with `develop` and similarly when merging PRs. It's ok to push rebased feature branches, it only becomes problematic when collaborating with someone on the same branch.

```
# Bring your branch up to date with develop
git pull --rebase origin develop
# Push rebased branch to github
git push --force-with-lease
```

supermarkion:
@Kirill888 sure, no problem. thanks.

PR #245, opened 2021-07-08
initial work done to support landsat bitmask geomedian

NikitaGandhi:
PR raised for review

**TODO**
* write tests
* any product specific requirements


libs/stats/odc/stats/_gm_bitmask.py
```diff
@@ -0,0 +1,146 @@
+"""
+Fractional Cover Percentiles
```
alexgleith: Not FC :-)

alexgleith:
Is there a reason why this isn't in with the other Geomedian transforms?

Also, we can probably call it LSUSGS rather than bitmask... although, maybe bitmask is better...

Even better is having one Gm with the option to use a bitmask or enumerated mask.

NikitaGandhi:
> Is there a reason why this isn't in with the other Geomedian transforms?
> 
> Also, we can probably call it LSUSGS rather than bitmask... although, maybe bitmask is better...
> 
> Even better is having one Gm with the option to use a bitmask or enumerated mask.

Implementation is quite different for both `StatsGMLS`  (which is extension of `StatsGM` class - https://github.com/opendatacube/odc-tools/blob/develop/libs/stats/odc/stats/_gm.py#L128) and `StatsGMLSBitMask` 
but I can try

alexgleith:
Up to you how you approach it! Keeping things simple, and incremental steps makes sense to me.

libs/stats/odc/stats/_gm_bitmask.py
```diff
@@ -0,0 +1,146 @@
...
+            |   |   |   x---------------------> cloud_shadow
+            |   |   o-------------------------> snow
+            |   x-----------------------------> clear
+            o---------------------------------> water
+        """
+        mask_band = xx[self.mask_band]
+        xx = xx.drop_vars([self.mask_band])
+
+        # set cloud_mask (cloud + cloud_shadow) bitmask
+        cloud_mask = da.bitwise_and(mask_band, 0b00011000) == 0
```
kieranricardo: I think the `QA_PIXEL` is 16 bit (I missed this earlier - sorry!) so this needs to have eight more zeros:

```suggestion
        cloud_mask = da.bitwise_and(mask_band, 0b0000_0000_0001_1000) == 0
```

libs/stats/odc/stats/_gm_bitmask.py
```diff
@@ -0,0 +1,146 @@
...
+        xx["cloud_mask"] = cloud_mask
+
+        # set no_data bitmask
+        keeps = da.bitwise_and(mask_band, 0b00000001) == 0
```
kieranricardo: And the same down here:

```suggestion
        keeps = da.bitwise_and(mask_band, 0b0000_0000_0000_0001) == 0
```

libs/stats/odc/stats/_gm_bitmask.py
```diff
@@ -0,0 +1,146 @@
...
+            |   |   |   |   o-----------------> cloud
+            |   |   |   x---------------------> cloud_shadow
+            |   |   o-------------------------> snow
+            |   x-----------------------------> clear
+            o---------------------------------> water
+        """
+        mask_band = xx[self.mask_band]
+        xx = xx.drop_vars([self.mask_band])
+
+        # set cloud_mask (cloud + cloud_shadow) bitmask
```
kieranricardo: maybe add something to this comment to specify that True= no cloud, Flase=clouse
```suggestion
        # set cloud_mask (cloud + cloud_shadow) bitmask - True=no cloud, False=cloud
```

kieranricardo (commented):
looking good @NikitaGandhi!
 
Mostly this just needs tests

kieranricardo:
@alexgleith @NikitaGandhi imo this implementation has a couple of advantages over the other GM classes and so I think it might be better to initially implement this without deriving from (and changing) the base class bc its easier and we don't risk breaking untested code. 

In a future PR we should change the GM base class to this pattern - which should be pretty easy with the right tests :) 

The advantages over the old GM stuff are:
- this implementation avoids the anti pattern  of loading in the mask band twice. 
- code is cleanly separated into the  `reduce` and `_native_op_data_band` functions which makes testing easier 

but we can always change the base class now for consistency - up to you @NikitaGandhi !

NikitaGandhi:
Let's keep it separate for this PR and will slowly improve overall implementation

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,154 @@
...
+            0-----------------------------------------------------------------> cirrus_confidence
+        """
+        mask_band = xx[self.mask_band]
+        xx = xx.drop_vars([self.mask_band])
+
+        # set cloud_mask (cloud + cloud_shadow) bitmask - True=non-cloud, False=cloud
+        cloud_mask = da.bitwise_and(mask_band, 0b0000_0000_0001_1000) == 0
+        xx["cloud_mask"] = cloud_mask
+
+        # set no_data bitmask - True=no-data
```
kieranricardo: Its the other way round here - True=data, False=Nodata

NikitaGandhi: > Its the other way round here - True=data, False=Nodata

Thanks! change is coming

kieranricardo:
looks good! are you planning on implementing tests here or in another PR @NikitaGandhi?

alexgleith:
Issues that I've noticed:
* There are large areas that have zero observations in the `_count` band, not sure why
* NODATA is coming through as `65535` but is set in the metadata as `0`

Image below is the count, where black is zero and white goes up to 20
image

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        mask_band = xx[self.mask_band]
+        xx = xx.drop_vars([self.mask_band])
+
+        # set cloud_mask (cloud + cloud_shadow) bitmask - True=non-cloud, False=cloud
+        cloud_mask = da.bitwise_and(mask_band, 0b0000_0000_0001_1000) == 0
+        xx["cloud_mask"] = cloud_mask
+
+        # set no_data bitmask - True=data, False=no-data
+        keeps = da.bitwise_and(mask_band, 0b0000_0000_0000_0001) == 0
```
alexgleith, line 94: I think this is our problem. We're doing this on a stack of data, right?

I thin we need to do this the other way. We should only drop where there are All nodata.

This is saying where it's never nodata, then keep it. But some scenes have nodata in different areas.

We want to say if it's ALWAYS nodata, then DROP IT.

libs/stats/tests/test_gm_ls_bitmask.py
```diff
@@ -0,0 +1,114 @@
...
+    index = pd.MultiIndex.from_tuples(tuples, names=["time", "solar_day"])
+    coords = {
+        "x": np.linspace(10, 20, band_red.shape[2]),
+        "y": np.linspace(0, 5, band_pq.shape[1]),
+        "spec": index,
+    }
+
+    data_vars = {"band_red": (("spec", "y", "x"), band_red), "QA_PIXEL": (("spec", "y", "x"), band_pq)}
+    attrs = dict(crs="epsg:32633", grid_mapping="spatial_ref")
+    xx = xr.Dataset(data_vars=data_vars, coords=coords, attrs=attrs)
```
kieranricardo, line 37: We're getting errors in our tests bc geomedian requires that a `nodata` value is set. 

```suggestion
    xx = xr.Dataset(data_vars=data_vars, coords=coords, attrs=attrs)
    xx['band_red'].attrs['nodata'] = 0 
```

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,160 @@
...
+    def _fuser(self, xx):
+        """
+        Fuse cloud_mask with AND, and apply mask_cleanup if requested
+        """
+        cloud_mask = xx["cloud_mask"]
+        xx = _xr_fuse(xx.drop_vars(["cloud_mask"]), partial(_first_valid_np, nodata=0), '')
+        xx["cloud_mask"] = _xr_fuse(cloud_mask, _fuse_and_np, cloud_mask.name)
+
+        if self.filters is not None:
+            xx["cloud_mask"] = mask_cleanup(xx["cloud_mask"], self.filters)
```
kieranricardo: We're using a filter designed to expand and contract a mask where `True=cloud`. Ours is the other way round! 

We should change our mask so that `True=cloud`. This means we'll need to use `or` fusing instead of `and`, and use `erase_bad` in reduce instead of `keep_good_only`

NikitaGandhi:
@alexgleith @kieranricardo - PR is ready to review again.

Thanks @alexgleith for fixing docker.

kieranricardo (approved):
@NikitaGandhi looks good to me! 🎉

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,160 @@
...
+            |   |   |   x-----------------------------------------------------> snow_ice_confidence
+            |   |   o---------------------------------------------------------> snow_ice_confidence
+            |   x-------------------------------------------------------------> cirrus_confidence
+            0-----------------------------------------------------------------> cirrus_confidence
+        """
+        mask_band = xx[self.mask_band]
+        xx = xx.drop_vars([self.mask_band])
+
+        # set cloud_mask (cloud + cloud_shadow) bitmask - True=cloud, False=non-cloud
+        cloud_mask = da.bitwise_and(mask_band, 0b0000_0000_0001_1000) != 0
```
alexgleith:
```suggestion
        cloud_mask = da.bitwise_and(mask_band, 0b0000_0000_0001_1010) != 0
```

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,163 @@
...
+        cloud_mask = xx["cloud_mask"]
+        xx = xx.drop_vars(["cloud_mask"])
+
+        # erase pixels with cloud
+        xx = erase_bad(xx, cloud_mask)
+
+        gm = geomedian_with_mads(xx, **cfg)
+        gm = gm.rename(self.renames)
+
+        # TODO: add scaling on produce gm again?
```
cbur24: @NikitaGandhi Would you like me to have a go at writing this bit of code? At the very least to add the conceptual outline of how we want the data to be output so others can comment on it.

NikitaGandhi: sure chad, as long we keep it optional :)

cbur24:
@NikitaGandhi @alexgleith I had a go writing a for-loop in the `reduce` function of `_gm_ls_bitmask.py` to convert the Landsat GM back to surface reflectance.  Because I don't often work with classes and config yamls, I'm not sure if I set the flag `return_SR` (return surface reflectance) in an appropriate way, so I'll let you change that if necessary.

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # erase pixels with cloud
+        xx = erase_bad(xx, cloud_mask)
+
+        gm = geomedian_with_mads(xx, **cfg)
+        gm = gm.rename(self.renames)
+
+        # Scale USGS Landsat bands into surface reflectance
+        if return_SR:
+            for band in gm.data_vars:
```
kieranricardo, line 143: I think this could be:

```suggestion
            for band in self.bands:
```

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,172 @@
...
+                if band in self.bands:
+                    gm[band] = (scale*10000 * gm[band]+offset*10000).astype(gm[band].dtype)
```
kieranricardo: @cbur24 `stats` already unscales the `geomedian` and `edev` bands. And `edev` can just be scaled so something like this should work:

```suggestion
                    gm[band] = 10000 * gm[band]
            gm['edev'] = 10000 * gm['edev']
```

Unfortunately though the `bcdev` and `sdev` bands aren't translation invariant and are computed with the scaled data - but that's a seperate issue!

cbur24: @kieranricardo Africa's USGS Landsat Collection 2 data is scaled very differently form Australia's.  `Stats` returns the data in its original scaling, which is not equivalent to surface reflectance for USGS L2C2. This code snippet is to reapply the USGS's scale and offset factors so we can save the data in surface reflectance values (i.e. ARD values) rather than their raw digital numbers (which the user would then have to apply a scale and offset each time they loaded the gm).  
The multiplication by 10,000 is to ensure the Landsat GM matches the same range of values as our Sentinel-2 gm product, and to allow for saving the data in integer format because USDS SR is 0-1.  So in other words, the code should be fine as it is.
 
The `edev` stuff is more complicated because we aren't sure that stats is reapplying the scale and offset factors as it should, so we need to look into it further.

kieranricardo: @cbur24 makes sense!

If you're still looking for it here's the line of code that transforms `edev` back into it's original scaling:

https://github.com/opendatacube/odc-tools/blob/aaaa4e673016e3a7d5b3c97834412a2bb9635a10/libs/algo/odc/algo/_geomedian.py#L438

so something like this should work:

```
gm['edev'] = 10000 * scale *gm['edev']
```

To avoid bias I think the transformed geomedian should be rounded before being caste to an int:

```suggestion
                    gm[band] = np.round((scale*10000 * gm[band]+offset*10000)).astype(gm[band].dtype)
```

And I have no idea if this ends up being an issue or not but scaling the geomedian output by `0.275` (I think thats what `scale * 10_000` ends up being) and casting to an integer ends up decreasing the precision by 4x. I don't know what the precision of the input data actually is so sorry if this is completely irrelevant!

fangfy: > so something like this should work:
> 
> ```
> gm['edev'] = 10000 * scale *gm['edev']
> ```
>
> To avoid bias I think the transformed geomedian should be rounded before being caste to an int:
>

Both above sound good.
 
> And I have no idea if this ends up being an issue or not but scaling the geomedian output by `0.275` (I think thats what `scale * 10_000` ends up being) and casting to an integer ends up decreasing the precision by 4x. I don't know what the precision of the input data actually is so sorry if this is completely irrelevant!

It's reflectance between 0 - 1 scaled to 0-10000, so rounding error smaller than 0.01% which is more than enough.

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,172 @@
...
+        Fuse cloud_mask with OR, and apply mask_cleanup if requested
+        """
+        cloud_mask = xx["cloud_mask"]
+        xx = _xr_fuse(xx.drop_vars(["cloud_mask"]), partial(_first_valid_np, nodata=0), '')
+        xx["cloud_mask"] = _xr_fuse(cloud_mask, _fuse_or_np, cloud_mask.name)
+
+        # apply filters - [r1, r2]
+        # r1 = shrinks away small areas of the mask
+        # r2 = adds padding to the mask
+        if self.filters is not None:
```
kieranricardo: @NikitaGandhi @alexgleith the fusing may not work as expected here. It turns out that `fuser` is called twice during `load_with_native_transform` so the filtering is also being applied twice which is probably not the intended behaviour. The same (potential?) issue also happens in `load_enum_filtered` and affects all the other stats geomedian classes.

We can avoid this by moving the filtering into `reduce`

libs/stats/odc/stats/_gm_ls_bitmask.py
cbur24: Added the requested changes but haven't actually tested if the `xarray.DataArray.round()` method works because my sandbox is busy with other processing right now. Will test when I get a chance

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,174 @@
...
+            resampling=self.resampling,
+            chunks=chunks,
+        )
+
+        return xx
+
+    def reduce(self, xx: xr.Dataset) -> xr.Dataset:
+        scale = 0.0000275
+        offset = -0.2
+        return_SR = False
```
kieranricardo:
```suggestion
```

We should make `return_SR` a class attribute (`self.return_SR`) that can specified with a config file.

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,174 @@
...
+        cloud_mask = xx["cloud_mask"]
+        cfg = dict(
+            maxiters=1000,
+            num_threads=1,
+            scale=scale,
+            offset=offset,
+            return_SR=return_SR,
```
kieranricardo: Geomedian doesn't have a `return_SR` option so this isn't needed here

```suggestion
```

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # erase pixels with cloud
+        xx = xx.drop_vars(["cloud_mask"])
+        xx = erase_bad(xx, cloud_mask)
+
+        gm = geomedian_with_mads(xx, **cfg)
+        gm = gm.rename(self.renames)
+
+        # Scale USGS Landsat bands into surface reflectance
+        if return_SR:
```
kieranricardo:
```suggestion
        if self.return_SR:
```

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,174 @@
...
+            for band in gm.data_vars:
+                if band in self.bands:
+                    gm[band] = scale*10000 * gm[band]+offset*10000
+                    gm[band] = gm[band].round().astype(np.uint16)
```
kieranricardo:
```suggestion
            for band in self.bands:
                gm[band] = scale*10000 * gm[band]+offset*10000
                gm[band] = gm[band].round().astype(np.uint16)
```

kieranricardo (changes requested):

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -0,0 +1,175 @@
...
+            **other,
+    ):
+        self.mask_band = mask_band
+        self.resampling = resampling
+        self.bands = bands
+        self.filter = filter
+        self.work_chunks = work_chunks
+        self.renames = aux_names
+        self.aux_bands = list(aux_names.values())
+        self.return_SR = False  # Flag to scale USGS Landsat bands into surface reflectance
```
alexgleith: I don't think this is the correct terminology.

Can we say "return_scale" maybe? And make it default to `10,000`.

kieranricardo (approved):
@NikitaGandhi this looks good to me! 🎉 🎉

kieranricardo:
@NikitaGandhi is this ready to be merged?

NikitaGandhi:
> @NikitaGandhi is this ready to be merged?

Let's merge this for now. I will raise more PR as we progress

PR #247, opened 2021-07-08
add thread pool to dump_to_odc

carioca-au:


apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -118,6 +165,12 @@ def dump_to_odc(
 )
 @click.argument("uri", type=str, nargs=1)
 @click.argument("product", type=str, nargs=1)
+@click.option(
```
alexgleith: Perhaps we make this configurable for n-threads?

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -55,6 +56,52 @@ def dump_to_odc(
...
+                metadata,
+                uri,
+                dc,
+                doc2ds,
+                update,
+                update_if_exists,
+                allow_unsafe,
+            )
+            for uri, metadata in uris_docs
+        ]
```
alexgleith, line 101: Does this load everything into a list first? Because I think it will be slow.

Everything else is a generator, which is a streaming flow of URIs then YAML docs off S3.

alexgleith: Can we do a functools function to handle the arguments instead?

alexgleith (changes requested):

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
carioca-au: Easy

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -163,14 +162,9 @@ def execute_index_update_dataset(metadata, uri, datacube, doc_to_ds, updated, if
     default=False,
     help="Needed when accessing requester pays public buckets",
 )
+@click.option('--n-threads', default=100, help='Needed when using multithreading to perform better')
```
alexgleith: Does the default mean it's always on? Default should be `None` or `0`, I think.

Issue #209, opened 2021-06-03
Possible improvements to new xr_percentile function


robbibt:
Hey @kieranricardo and @Kirill888, I've just had a chance to test out the new `xr_percentile` function. It seems to function far faster than the built-in `.quantile` functionality in `xarray`, which is really exciting! I did have a few pieces of feedback though which might help it be easier to use and to fit into existing workflows.

**Issue 1:** The function currently returns an `xr.Dataset` with a data variable for each percentile the user requests. E.g. below, I've requested a `0.01` and a `0.999` percentile. These are used to label the new variables by appending the percentiles onto the original variable name. 

![image](https://user-images.githubusercontent.com/17680388/120572698-c0c4eb00-c45f-11eb-8890-82c81a9b8c59.png)

This approach feels a bit clunky to to me as the user can't anticipate the naming convention used for the new band names, especially as the input `0.01` value is converted to an integer `1`. It also produces issues where the user requests higher than 0.01 precision, e.g. the `0.999` above is clipped to `0.99` in the band name (e.g. if a user requests both `0.995` and `0.999`, they get combined into one in the output).

The native `xarray` solution is to instead produce an `xr.Dataset` with a new "quantile" dimension, which is then labelled with the requested quantiles. This feels a bit more elegant to me:

![image](https://user-images.githubusercontent.com/17680388/120572837-ff5aa580-c45f-11eb-9ba8-2a48aca3b3b0.png)

**Issue 2:** It would be nice if this function also supported non-dask input data too, as for a lot of smaller-scale science team applications we don't always need to use dask throughout an entire workflow. If I try running it on a `xr.Dataset` in memory, I get this error:

![image](https://user-images.githubusercontent.com/17680388/120573675-5319be80-c461-11eb-973f-fbdd4ac135df.png)

Kirill888:
Agree on both counts.

RichardScottOZ:
This sounds great given quantile doesn't work on dask I think?

RichardScottOZ:
and a few hundred times faster than this would be great:-
3 variables, 44 bands, 100 million pixels each approx

![image](https://user-images.githubusercontent.com/72196131/123183978-dca53500-d4d1-11eb-95e9-0377b3698848.png)

PR #248, opened 2021-07-09
Xarray percentile - support non-dask input and percentiles in a new 

kieranricardo:
This PR does a couple of things and addresses #209:

- renames the old `xr_percentile` function to `xr_quantile_bands`
- adds a new function named `xr_quantile` that outputs the quantiles as a dimension
- adds support for non dask input  in `xr_quantiles` and `xr_quantiles_bands`

@robbibt is this what you had in mind? :)


robbibt:
@kieranricardo This looks brilliant! Thanks so much for updating this - I'm testing it now and should have some feedback early this week!

robbibt:
Hey Kieran, thanks so much for this, and apologies for the delay in reviewing it! The revised function works amazingly quickly and the quantile dimension is exactly what I had in mind - will make it much more compatible with code that used to user the slower `.quantile` methods. A few bits of feedback - very happy for some of these to be pushed back to future work depending on complexity!
1. It's likely users will want to apply this function to data that doesn't necessarily have a time dimension (e.g. I currently have data that is labelled by "year"). From what I can tell the function does actually work for different dimension names - is it currently based on the first dimension in the list? If so, I think at minimum we should update the docstring to communicate this.
2. Related to the 1. above: one of the most powerful features of `xarray` is being able to specify the `dim` over which to apply functions, e.g. `dim=['x', 'y']` or `dim='time'` to reduce data spatially or temporally. It would be amazing if this function could support this kind of functionality rather than being hard-coded to `time`, but if we think it's best to leave it as a specialised time dimension reducing function, it would be good to emphasise this in the docstring too and point users to the slower funcs if they want to do more complicated things.
3. The current `nodata` behavior could perhaps be improved if this value was read directly from the `nodata` attribute on the input `xr.DataArray` data, e.g. `ds.nbart_red.nodata`, with a fallback to `NaN` or a custom supplied value if this doesn't exist. I can see that there's a requirement for the nodata value to be the smallest/largest/`NaN`, so perhaps this might require a warning if this isn't the case. 
4. Rename the `quantile` arg to `q` for compatibility with the [numpy](https://numpy.org/doc/stable/reference/generated/numpy.quantile.html) and [xarray](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.quantile.html) implementation?

robbibt (changes requested):
Hey Kieran, thanks so much for this, and apologies for the delay in reviewing it! The revised function works amazingly quickly and the quantile dimension is exactly what I had in mind - will make it much more compatible with code that used to user the slower `.quantile` methods. A few bits of feedback - very happy for some of these to be pushed back to future work depending on complexity!
1. It's likely users will want to apply this function to data that doesn't necessarily have a time dimension (e.g. I currently have data that is labelled by "year"). From what I can tell the function does actually work for different dimension names - is it currently based on the first dimension in the list? If so, I think at minimum we should update the docstring to communicate this.
2. Related to the 1. above: one of the most powerful features of `xarray` is being able to specify the `dim` over which to apply functions, e.g. `dim=['x', 'y']` or `dim='time'` to reduce data spatially or temporally. It would be amazing if this function could support this kind of functionality rather than being hard-coded to `time`, but if we think it's best to leave it as a specialised time dimension reducing function, it would be good to emphasise this in the docstring too and point users to the slower funcs if they want to do more complicated things.
3. The current `nodata` behavior could perhaps be improved if this value was read directly from the `nodata` attribute on the input `xr.DataArray` data, e.g. `ds.nbart_red.nodata`, with a fallback to `NaN` or a custom supplied value if this doesn't exist (I think this is what other `odc.tools` funcs do). I can see that there's a requirement for the nodata value to be the smallest/largest/`NaN`, so perhaps this might require a warning if this isn't the case. 
4. Rename the `quantile` arg to `q` for compatibility with the [numpy](https://numpy.org/doc/stable/reference/generated/numpy.quantile.html) and [xarray](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.quantile.html) implementation?

robbibt (approved):
I'd love to start using this func for some upcoming coastal work, so I think the best plan might be to merge this as-is, and I can raise the extra suggestions below as an issue for future work. Thanks again @kieranricardo!

Kirill888:
@kieranricardo can you please rebase against develop rather than merge, and also cleanup commit history a bit, maybe merge some commits, or even all of them.

kieranricardo:
@Kirill888 will a squash and merge be sufficient?

Kirill888:
@kieranricardo I don't know is there one commit worth of changes in here? You still need to pull in changes from develop too

PR #250, opened 2021-07-13
use rust geomedian

kieranricardo:
This PR adds the option to use the `datacube_compute` `geomedian` in the `gm-ls-bitmask` plugin. To use `datacube_compute` instead of `hdstats`:

1. Install `datacube_compute`: https://github.com/opendatacube/datacube-compute
2. Set `use_hdstats=False` in the config for the `gm-ls-bitmask` pluging


kieranricardo:
@alexgleith @NikitaGandhi I don't think we should merge this until after some more discussions but its here if you want to run some tests

alexgleith:
Cool!

NikitaGandhi:
@kieranricardo this require `datacube-compute` package installed in docker

omad:
This has to go on hold for the moment, we can come back to it later.

PR #251, opened 2021-07-13
install hdstats in to test docker image

NikitaGandhi:
- Install hdstats library - needed for ls-gm testing
- New image is published from develop branch only

### DIFF
--- a/docker/constraints.txt
+++ b/docker/constraints.txt
@@ -55,7 +55,7 @@ google-resumable-media==1.2.0
 googleapis-common-protos==1.53.0
 greenlet==1.1.0
 HeapDict==1.0.1
-hdstats==0.1.8.post1
+hdstats==0.2.1
 hypothesis==6.12.0
 idna==2.10
 imageio==2.9.0
--- a/docker/constraints.txt
+++ b/docker/constraints.txt
@@ -55,7 +55,7 @@ google-resumable-media==1.2.0
 googleapis-common-protos==1.53.0
 greenlet==1.1.0
 HeapDict==1.0.1
-hdstats==0.2.1
+hdstats==hdstats-0.1.8.post1
 hypothesis==6.12.0
 idna==2.10
 imageio==2.9.0
--- a/docker/constraints.txt
+++ b/docker/constraints.txt
@@ -55,7 +55,7 @@ google-resumable-media==1.2.0
 googleapis-common-protos==1.53.0
 greenlet==1.1.0
 HeapDict==1.0.1
-hdstats==hdstats-0.1.8.post1
+hdstats==0.1.8.post1
 hypothesis==6.12.0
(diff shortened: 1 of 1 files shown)


kieranricardo:
@NikitaGandhi this only changes the constraints file so AFAIK if `hdstats` wasn't getting installed before this won't cause it to be installed.

Does this fix CI in your PR? It might be easier to make these changes in your other PR so we can see whether it fixes CI

alexgleith:
Hey @NikitaGandhi like @kieranricardo says, you need to add the requirement to `requirements.txt` and the version goes in `constrants.txt`.

NikitaGandhi:
> @NikitaGandhi this only changes the constraints file so AFAIK if `hdstats` wasn't getting installed before this won't cause it to be installed.
> 
> Does this fix CI in your PR? It might be easier to make these changes in your other PR so we can see whether it fixes CI

test runs on docker-compose and it pulls image from dockerhub so my PR never going to pass until that docker image has hdstats lib install

NikitaGandhi:
> Hey @NikitaGandhi like @kieranricardo says, you need to add the requirement to `requirements.txt` and the version goes in `constrants.txt`.

I check requirement.txt and it has hdstats

alexgleith:
Need to look into how the image is built... also, we might not be allowed to run the hdstats code here. @GypsyBojangles ?

alexgleith:
Hmm. Maybe it just needs a rebuild, need to trigger this: https://github.com/opendatacube/odc-tools/blob/develop/.github/workflows/build-test-docker.yml

NikitaGandhi:
> https://github.com/opendatacube/odc-tools/blob/develop/.github/workflows/build-test-docker.yml

It publishes new image from develop branch only

```
        if [ ${GITHUB_REF:-} = "refs/heads/develop" ]; then
           push_image=yes
        fi
```

alexgleith:
Let's just merge and do a new build.

docker/constraints.txt
```diff
@@ -55,6 +55,7 @@ google-resumable-media==1.2.0
 googleapis-common-protos==1.53.0
 greenlet==1.1.0
 HeapDict==1.0.1
+hdstats==0.2.1
```
alexgleith, line 58: I think this should be 
```suggestion
hdstats==hdstats-0.1.8.post1
```

https://packages.dea.ga.gov.au/hdstats/

kieranricardo:
I think I found where we need to add this.

update  the `main.yml` file "Install dependencies" step to add `hdstats`:

```
 - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        python -m pip install \
         wheel \
         setuptools-scm \
         twine \
         hdstats
```

kieranricardo:
oh wait maybe that's not the right place

PR #258, opened 2021-07-18
#244 Use mechanism from datacube-docker to test

### LINKED ISSUE #244: There's no integration tests
Some of the tools in this repository rely on the ODC Database to do their work.

We do mock things like SQS or datasets and run simple functional tests, but some tools like `sqs-to-dc` or `s3-to-dc` really need to be tested end-to-end.

Can we have a discussion about how to set up some integration tests?

The simplest way is to have a shell script that calls the tools, with a Docker Compose file that sets up an environment.

More complex is having an environment that can be bootstapped and then mocking things like SQS and S3 to do the testing from.

Thoughts?

whatnick:
Signed-off-by: Tisham Dhar <whatnickd@gmail.com>

Closes #244 
- [x] Hack together elements from [datacube-docker](https://github.com/opendatacube/datacube-docker/blob/main/index/assets/bootstrap-odc.sh) to perform integration testing
- [x] Support testing of Click based tools using [standard pattern](https://click.palletsprojects.com/en/8.0.x/testing/)
- [x] Support coverage
- [ ] Fix up missing/broken links to product definitions see : https://github.com/GeoscienceAustralia/dea-config/issues/797

Coverage is at 52%
```
TOTAL                                                       5560   2683    52%
```

Not a bad start at all.


apps/dc_tools/assets/bootstrap-odc.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
alexgleith, line 1: This script is now in two places...

Maybe it should be graduated to a Python file?

alexgleith (approved):
Looks like it doesn't increase run time and does add some valuable tests!

I don't really think that the bootstrap shell script should be included, since it's not actually used anywhere and it's now duplicated in two places, so the canonical version is...?

apps/dc_tools/assets/bootstrap-odc.sh
whatnick: We could have different ones. The tests are now in 2 places too. We consolidate / create then dedup.

PR #267, opened 2021-07-21
update filter to support morphological closing - useful in Coastlines

NikitaGandhi:
added a third parameter to filters (r1, r2, r3). This will allow us to try morphological closing(to fill small gaps in cloud) as suggested by @robbibt 

> For interest, here's the cloud cleanup recipe we used for Coastlines: morphological closing (to fill small gaps in cloud and cloud shadow - particularly a problem with DEA's cloud shadow mask), opening (to remove narrow false positive clouds), then dilation (to remove all cloud fringes).

```
# Close mask to remove small holes in cloud, open mask to     
# remove narrow false positive cloud, then dilate    
mask = odc.algo.binary_closing(mask, r3)    
mask_cleaned = odc.algo.mask_cleanup(mask, (r1, r2))
```

### DIFF
--- a/libs/stats/odc/stats/_pq_bitmask.py
+++ b/libs/stats/odc/stats/_pq_bitmask.py
@@ -1,6 +1,14 @@
 """
 USGS Landsat pixel quality
+
+pq_band = input band for cloud masking
+aerosol_band = input band for aerosol masking
+filters to clear cloud mask - [[r1, r2, r3], ...]
+    r1 = shrinks away small areas of the mask
+    r2 = adds padding to the mask
+    r3 = remove small holes in cloud - morphological closing
 """
+
 from functools import partial
 from typing import List, Optional, Tuple
 
@@ -8,7 +16,7 @@ import dask.array as da
 import xarray as xr
 
 from odc.algo import mask_cleanup, keep_good_only
-from odc.algo._masking import _xr_fuse, _first_valid_np, _fuse_or_np
+from odc.algo._masking import _xr_fuse, _first_valid_np, _fuse_or_np, binary_closing
 from odc.algo.io import load_with_native_transform
 from odc.stats.model import Task
 
@@ -26,7 +34,7 @@ class StatsPQLSBitmask(StatsPluginInterface):
             self,
             pq_band: str = "QA_PIXEL",
             aerosol_band: Optional[str] = None,
-            filters: Optional[List[Tuple[int, int]]] = [(1, 2)],
+            filters: Optional[List[Tuple[int, int, int]]] = [[1, 2, 2]],
             resampling: str = "nearest",
     ):
         self.filters = filters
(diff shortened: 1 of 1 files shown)


NikitaGandhi:
Need some advise here on filter options - @robbibt @alexgleith @cbur24 and Fang
I was thinking to try this filters options -  `[[1, 2, 2], [2, 5, 2]]`

alexgleith:
I thought "opening" and "dilation" are the same thing?

robbibt:
Hey @NikitaGandhi, this looks good to me. There were two main reasons we implemented closing for Coastlines:
1. Landsat 7 SLC-off gaps caused cloud and cloud shadow to be broken up, which could sometimes cause cloud masking data to be excessively removed during the opening operation. Closing helped (but didn't completely solve) this issue.
2. The second (most important) reason was that the DEA cloud shadow mask has some serious artefacts, e.g. image below. This meant that the closing operation would often remove narrow cloud shadow features. By doing closing first, we closed these gaps so that the opening operation would inadvertently not remove all cloud shadow.

image

I don't know what the cloud mask is like for Africa, so 2. may not be as relevant. 1. may be pretty important through, so I would recommend testing any algorithm on Landsat 7 SLC-off (post 2003 data).

Finally, it would be good to do some negative tests too... perhaps `[[1, 2, 0], [2, 5, 0]]` as well?

robbibt:
@alexgleith: there's four main morphological operators:
1. Dilation (growing True pixels outwards)
2. Erosion (the opposite of dilation; eroding away True pixels)
3. Closing (first dilating, then eroding True pixels. Used to close small gaps between True pixels)
4. Opening (first eroding, then dilating True pixels. Used to remove small/narrow areas of True pixels)

alexgleith:
Ok, so the three step process proposed in this PR does the following:
* closing
  * dilation
  * erosion 
* dilation
* erosion

So it's doing four steps...

Do we really need to do all four?

robbibt:
`odc.algo.mask_cleanup` does opening followed by dilation (not erosion followed by dilation). So it's really:
* closing
  * dilation
  * erosion 
* opening
  * erosion
  * dilation
* dilation

So you're right, you could probably simplify that workflow to perhaps:

* dilation
* erosion 
* dilation

Although if the purpose of this is to first test whether the closing step is necessary, it might be worth testing them separately at the start and then optimise later.

cbur24:
Agreed upon set of morphological tests:
```
[opening_radius, dilation_radius, closing_radius]
# [2, 2, 2] # Do morphological operations improve the mask?
# [2, 0, 2] # Dilation removes lots of data, so test without it
# [2, 2, 0] # Is binary closing even necessary?
# [5, 2, 2] # Bigger opening radii can remove larger (wider) coastal false positives, but at what cost?
```

cbur24 (approved):

PR #268, opened 2021-07-21
Inventory improvements

carioca-au:
- Added thread pool to speed up the process
- Added filters prefix, suffix and contains
-


alexgleith:
Does this threadpool actually make it faster?

Also, soaking with a default of 1,000 threads is extreme, isn't it?

alexgleith (approved):

PR #269, opened 2021-07-22
provide support for dataset filter

NikitaGandhi:
provide a support for dataset filter

- added a new optional options - `--dataset-filter`
- this will allow us to query product with search filter like
```
query: {'product': 'ls8_sr', 'collection_category': 'T1'}
```

### DIFF
--- a/libs/stats/odc/stats/_cli_save_tasks.py
+++ b/libs/stats/odc/stats/_cli_save_tasks.py
@@ -1,3 +1,5 @@
+import json
+
 import click
 import sys
 from odc.io.text import click_range2d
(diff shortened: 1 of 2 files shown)


libs/stats/odc/stats/_cli_save_tasks.py
```diff
@@ -170,13 +180,16 @@ def gqa_predicate(ds):
...
 
 
-def _parse_products(dc, products, temporal_range):
-        
-    query = {"product": products}
+def _parse_products(dc, products, temporal_range, dataset_filter):
+
+    filter = {}
+    if dataset_filter:
+        filter = json.loads(dataset_filter)
```
kieranricardo: I think this code could be moved into the save_tasks function - then we don't need to duplicate this code.

uchchwhash (approved):
Looks good to me!

kieranricardo (approved):
LGTM! :)

PR #270, opened 2021-07-23
revised pd bitmask stats implementation

NikitaGandhi:
- revised pd bitmask stats implementation
- test coverage for _pq_bitmask.py

### DIFF
--- a/libs/stats/tests/test_pq_bitmask.py
+++ b/libs/stats/tests/test_pq_bitmask.py
@@ -152,11 +152,40 @@ def test_reduce_for_clear_aerosol(dataset_with_aerosol_band):
     reduce_result = xx.compute()
 
     assert set(reduce_result.data_vars.keys()) == set(
-        ["total", "clear", "clear_1_1_0", "clear_aerosol", "clear_aerosol_2_2_2"]
+        ["total", "clear", "clear_1_1_0", "clear_aerosol"]
     )
+
+    expected_result = np.array(
+        [[2, 2], [1, 2]]
+    )
+    clear = reduce_result["clear"].data
+    assert (clear == expected_result).all()
+
+    expected_result = np.array(
+        [[1, 2], [1, 2]]
+    )
+    clear_aerosol = reduce_result["clear_aerosol"].data
+    assert (clear_aerosol == expected_result).all()
+
+def test_reduce_for_clear_aerosol_with_filter(dataset_with_aerosol_band):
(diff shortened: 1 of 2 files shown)


libs/stats/tests/test_pq_bitmask.py
```diff
@@ -0,0 +1,162 @@
...
+
+def test_fuser(dataset):
+    pq = StatsPQLSBitmask()
+
+    xx = pq._native_tr(dataset)
+    xx = xx.groupby("solar_day").map(pq._fuser)
+    fuser_result = xx.compute()
+
+    expected_result = np.array(
+        [[True, True], [True, True]],
```
kieranricardo: I think this should be:

```suggestion
        [[True, False], [False, True]],
```

Because the second and third pixels have some non-clear measurements on that day. Maybe an `or` ruser is being used instead of an `and`?

libs/stats/tests/test_pq_bitmask.py
```diff
@@ -0,0 +1,162 @@
...
+
+def test_fuse_for_clear_aerosol(dataset_with_aerosol_band):
+    pq = StatsPQLSBitmask(pq_band = "QA_PIXEL", aerosol_band = "SR_QA_AEROSOL")
+
+    xx = pq._native_tr(dataset_with_aerosol_band)
+    xx = xx.groupby("solar_day").map(pq._fuser)
+    fuser_result = xx.compute()
+
+    expected_result = np.array(
+        [[True, True], [True, True]],
```
kieranricardo: Maybe this should be:

```suggestion
        [[False, True], [False, True]],
```

kieranricardo (commented):
The tests look good except for maybe the `test_fuser` - I think an `or` fuser is being used here instead of an `and` fuser. The convention in stats is that a pixel is considered cloudy if it is ever cloudy on that day so I think an `and` fuser is more appropriate here. What do you reckon?

kieranricardo (approved):
LGTM!

alexgleith (approved):
Looks fine to me!

cbur24:
Copied from slack:
Applying morphological operators to the `aerosol` band may not make sense since atmospheric aerosols aren't forming somewhat regular shapes like clouds. And while we know the quality issues of the cloud mask and can somewhat mitigate them with these binary operations, we don't know the quality of aerosol classifications...so maybe we need to reconsider the order of masking operations (e.g. cloudy regions-->apply morphology--> then aerosol-masking?)

cbur24 (approved):

PR #271, opened 2021-07-23
Set nodata flag to zero

kieranricardo:
This PR fixes an issue where a large portion of pixels are always set to `nodata`. This issue occurs because prior to this PR any pixel that overlapped with a wofs pixel (from the same day) that was not a dry and valid was set to nodata. Due their alignment, the triangular edge regions of tiles are always set to nodata causing any valid data from other scenes that overlaps with these regions to be set to nodata.

image

### DIFF
--- a/libs/stats/odc/stats/_fc_percentiles.py
+++ b/libs/stats/odc/stats/_fc_percentiles.py
@@ -44,15 +44,23 @@ class StatsFCP(StatsPluginInterface):
         Loads in the data in the native projection. It performs the following:
 
         1. Loads all the fc and WOfS bands
-        2. Set high slope terrain flag to 0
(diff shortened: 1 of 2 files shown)


emmaai (approved):
💯

emmaai:
I'm not fan of hard code of numbers/binary/hex in general. However, it was like that before this pr, I'm gonna live with it.

MatthewJA (commented):
LGTM too! Agreed with Emma for future tidying...

omad:
My brain has started it's regular Friday afternoon shutdown procedure, but

This looks more like a work around than a proper fix, and I'm not sure if there's any negative repercussions.

It seems the root issue would have to do with the order that we're masking and merging in...

But maybe this is okay, I can think about it more next week, or we can talk about it.

PR #273, opened 2021-07-27
Fix fc percentile wofs masking

kieranricardo:
The previous wofs masking assumed that if a pixel was `nodata` in `wofs` then it was `nodata` in `fc`. This turns out not to be true on the edges of scenes for some reason (maybe `fc` scenes are slightly larger?) causing thicks lines to appear in `wofs` masked areas near the edges of scenes.

This PR fixes this and handles the `nodata` pixels properly.

For each day, the fused daily WOfS measurement is considered:
- `bad` if at least one measurement on that day has one of the: cloud; cloud shadow; terrain shadow; low solar angle;  or non contiguous flags set to `True`
- `dry` if at least one measurement on that day is `(~bad) & dry` and there are no `wet` or `bad` measurements
- `wet` if at least one measurement on that day is `(~bad) & wet` and there are no `dry` or `bad` measurements

An `fc` measurement is only included if:
- there is at least one valid `fc` measurement from that day
- the corresponding fused daily `WOfS` measurement is `dry`

How does this sound?

### DIFF
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -727,12 +727,11 @@ def _nodata_fuser(xx, **kw):
     return choose_first_valid(xx, **kw)
 
 
-def fuse_mean_np(*aa, nodata):
+def _fuse_mean_np(*aa, nodata):
     assert len(aa) > 0
     
     out = aa[0].astype(np.float32)
     count = (aa[0] != nodata).astype(np.float32)
-
     for a in aa[1:]:
         out += a.astype(np.float32)
         count += (a != nodata)
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -730,14 +730,14 @@ def _nodata_fuser(xx, **kw):
 def fuse_mean_np(*aa, nodata):
     assert len(aa) > 0
     
+    out = aa[0].astype(np.float32)
     count = (aa[0] != nodata).astype(np.float32)
-    out = aa[0].astype(np.float32) * count
 
     for a in aa[1:]:
-        mask = (a != nodata)
-        out += a.astype(np.float32) * mask
-        count += mask
+        out += a.astype(np.float32)
+        count += (a != nodata)
 
+    out -= (len(aa) - count) * nodata
     out = np.round(out / count).astype(aa[0].dtype)
     out[count == 0] = nodata
     return out
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -725,3 +725,19 @@ def _nodata_fuser(xx, **kw):
     if xx.shape[0] <= 1:
         return xx
     return choose_first_valid(xx, **kw)
+
+
+def fuse_mean_np(*aa, nodata):
+    assert len(aa) > 0
+    
+    count = (aa[0] != nodata).astype(np.float32)
+    out = aa[0].astype(np.float32) * count
+
+    for a in aa[1:]:
+        mask = (a != nodata)
+        out += a.astype(np.float32) * mask
+        count += mask
+
+    out = np.round(out / count).astype(aa[0].dtype)
+    out[count == 0] = nodata
+    return out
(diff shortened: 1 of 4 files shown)


libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -41,40 +41,38 @@ def measurements(self) -> Tuple[str, ...]:
...
+        bad = xx.bad
+
+        xx = _xr_fuse(xx.drop_vars(["wet", "dry", "bad"]), partial(_first_valid_np, nodata=NODATA), '')
+        
+        wet = _xr_fuse(wet, _fuse_or_np, wet.name)
+        dry = _xr_fuse(dry, _fuse_or_np, dry.name)
+        bad = _xr_fuse(bad, _fuse_or_np, dry.name)
+        
+        xx["wet"] = wet & (~dry) & (~bad)
+        xx["dry"] = dry & (~wet) & (~bad)
```
emmaai: I don't understand why these two lines are necessary. `wet`, `dry` and `bad` is mutually exclusive from the line `54-56`. Anything happened in line `67-71` makes it not the case?

libs/stats/odc/stats/_fc_percentiles.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 from odc.stats.model import Task
 from odc.algo.io import load_with_native_transform
-from odc.algo import keep_good_only
+from odc.algo import keep_good_only, apply_numexpr
```
emmaai: `apply_numexpr` never used

emmaai (commented):
Can we NOT do masking by wofs `nodata`? sorry didn't look into what `keep_good_only` does. In `c2` we used landsat 7 data between 2010 and 2013 but not having stripe issues. Only guess that could be caused by wofs `nodata`, since it's the only difference on wofs mask between `c2` and `c3`. `c2` keeps the pixels if wofs says `0b00000001` but fc says it's good.

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: they are initially mutually exclusive for each individual input scene, but we fuse overlapping scenes from the same day with an `or` operation which which can cause them to not be mutually exclusive. E.g. say a pixel is in the overlap between two scenes and in one scene its dry and in the other its wet, then the fused pixel will be both wet and dry unless we do something like this

libs/stats/odc/stats/_fc_percentiles.py
kieranricardo: oh yep good catch!

kieranricardo:
@emmaai atm I'm only including pixels that have at least one `WOfS` `dry` measurement (from a solar day) and all the other `WOfS` measurements are either `dry` or `nodata`. So this is still considering the `WOfS` `nodata` flag - if all the `WOfS` measurements are `nodata` then a pixel isn't included!

The reason we can't filter on the `nodata` flag like the other `WOfS` flags is that there are large regions at the edges of tiles that are `nodata` and so these regions will never be included if filter in the same way - this is the initial problem I had where a lot of scenes were mostly blank.

Does that answer your question? happy to chat in person :)

kieranricardo:
Following an investigation with @emmaai and help from Lan Wei we determined that there is a difference in the `nodata` values in `WOfS` and `fc`. This is because `WOfS` is `nodata` if the surface reflectance or` fmask` is `nodata`, while `fc` is `nodata` only if the surface reflectance is `nodata`. Plotted below is a set of pixels that are `nodata` in `fmask` but valid in the surface reflectance.

Partially due to this complication we have decided to apply the `WOfS` masking step in the native transform and reproject that result, instead of reprojecting and fusing the masks.

image

emmaai (approved):
💯 
didn't realize test failed, but I'm happy with the change once @kieranricardo fix the test part.

libs/algo/odc/algo/_masking.py
```diff
@@ -730,14 +730,14 @@ def _nodata_fuser(xx, **kw):
 def fuse_mean_np(*aa, nodata):
     assert len(aa) > 0
```
emmaai, line 732: how about one line...
` return np.nanmean(aa[~np.isclose(aa, nodata)])`

emmaai: oops...my bad, sorry @kieranricardo I was late.

PR #275, opened 2021-07-27
Fix up integration tests and add in multiple products handling per yaml

alexgleith:
* Fix up the integration tests again, after I broke them.
* Fix up broken assumption that there's one doc per YAML file

### DIFF
--- a/.github/workflows/test-dc-tools.yml
+++ b/.github/workflows/test-dc-tools.yml
@@ -24,13 +24,16 @@ jobs:
       run: |
         cd tests
         docker-compose up -d
-        # Manually install over the top
+        # Manually install over the top to fix weird file symbolic links in CI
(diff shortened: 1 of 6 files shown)


alexgleith:
Hey @omad just FYI, the products list used in tests here, this one: https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/master/workspaces/prod-products.csv has missing product definitions in it.

Not sure if it matters! Ping @whatnick, FYI, the test is running but is `xfail` until this is rectified.

whatnick:
I have lodged an issue in dea-config. Since I have no clout at DEA, not
sure when it will be rectified. We should decouple from DEA and run our
tests using public datasets, this will raise the spectre of requester-pays
and S3 keys in CI.

See : https://github.com/GeoscienceAustralia/dea-config/issues/797

PR #277, opened 2021-07-28
Switch stac api to dc to using pystac-client

alexgleith:
* Set up indexing from a STAC API using a different tool

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/add_update_products.py
+++ b/apps/dc_tools/odc/apps/dc_tools/add_update_products.py
@@ -121,7 +121,7 @@ def cli(csv_path: str, update_if_exists: bool):
     # TODO: Add in some QA/QC checks
     added, updated, failed = add_update_products(dc, csv_path, update_if_exists)
 
-    logging.info(f"Added: {added}, Updated: {updated} and Failed: {failed}")
+    print(f"Added: {added}, Updated: {updated} and Failed: {failed}")
 
     # If nothing failed then this exists with success code 0
     sys.exit(failed)
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -4,38 +4,54 @@
 import logging
 import os
 import sys
-from typing import Generator, Optional, Tuple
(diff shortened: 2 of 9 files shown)


apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -4,38 +4,54 @@
...
-                "image/tiff; application=geotiff; profile=cloud-optimized",
-                "image/tiff; application=geotiff",
-            ]:
-                asset_link = os.path.dirname(asset["href"])
+        # Override self with canonical
+        if link.rel == "canonical":
+            self_link = link.target
+
+    for name, asset in item.assets.items():
+        if asset.media_type in [
```
carioca-au: This won't get USGS new standard which is **image/vnd.stac.geotiff; cloud-optimized=true"**

isn't it better to change to 
`if 'geotiff' in asset.media_type`

alexgleith: It's better to use the `role` of data, really...

But USGS make their QA bands "metadata" role, so that doesn't help.

I'll do what you suggest, I think it is more flexible :-)

apps/dc_tools/odc/apps/dc_tools/add_update_products.py
```diff
@@ -121,7 +121,7 @@ def cli(csv_path: str, update_if_exists: bool):
     # TODO: Add in some QA/QC checks
     added, updated, failed = add_update_products(dc, csv_path, update_if_exists)
 
-    logging.info(f"Added: {added}, Updated: {updated} and Failed: {failed}")
+    print(f"Added: {added}, Updated: {updated} and Failed: {failed}")
```
whatnick, line 124: Personal preference is is logging to print, and debug using log levels in environment. I will try to add that as an enhancement to the CLI tools. https://docs.python.org/3/howto/logging.html

alexgleith: I changed this so that it was consistent with some of the other tools, and with testing using `result.output` with `CliRunner`.

alexgleith: Logging with info here might mean that this is hidden.

PR #278, opened 2021-07-28
Fix up PySTAC changes for new release, add Microsoft PC compatibility for stac-to-dc

alexgleith:
* Fix up Stats and ESRI-LC STAC file creation
* Run a test against indexing `nasadem` from Microsoft's PC
* Fix up dependencies a bit.


whatnick (commented):
Pip / PyPI does not like GitHub dependencies ... Would love to get rid of those.

alexgleith:
> Pip / PyPI does not like GitHub dependencies ... Would love to get rid of those.

This is just temporary until we have a new release out of the STAC folks. Should be soon.

PR #279, opened 2021-07-29
Simplify the ESRI to DC tool

alexgleith:
* In hindsight, everything should just use STAC API to DC


alexgleith:
Fix for this is proposed here: https://github.com/opendatacube/datacube-core/pull/1167

Or here: https://github.com/opendatacube/datacube-core/pull/1150

alexgleith:
Closing because we really should just delete the esri-to-dc.

PR #283, opened 2021-07-30
Change to threading for maximum power

alexgleith:
* Threading is best
* This processes `1,000` docs in less than a minute including checking if they're indexed
* It handled `10,000` in  4 mins!

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -10,12 +10,23 @@ import click
 import pystac
 from datacube import Datacube
 from datacube.index.hl import Doc2Dataset
-from odc.apps.dc_tools.utils import (allow_unsafe, index_update_dataset, limit,
-                                     update_if_exists)
+from odc.apps.dc_tools.utils import (
+    allow_unsafe,
+    index_update_dataset,
+    limit,
+    update_if_exists,
+)
 from odc.index.stac import stac_transform, stac_transform_absolute
-from pystac_client import Client, ItemSearch
+from pystac.item import Item
+from pystac_client import Client
 
-logging.basicConfig(level=logging.INFO, format='%(asctime)s: %(levelname)s: %(message)s', datefmt='%m/%d/%Y %I:%M:%S')
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s: %(levelname)s: %(message)s",
+    datefmt="%m/%d/%Y %I:%M:%S",
+)
+
+import concurrent
 
 
 def _parse_options(options: Optional[str]) -> Dict[str, Any]:
@@ -27,7 +38,9 @@ def _parse_options(options: Optional[str]) -> Dict[str, Any]:
                 key, value = option.split("=")
                 parsed_options[key] = value
             except Exception as e:
-                logging.warning(f"Couldn't parse option {option}, format is key=value, exception was {e}")
+                logging.warning(
+                    f"Couldn't parse option {option}, format is key=value, exception was {e}"
+                )
 
     return parsed_options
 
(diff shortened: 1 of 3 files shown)


apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -60,26 +73,36 @@ def _guess_location(item: pystac.Item) -> Tuple[str, bool]:
...
+def item_to_meta_uri(item: Item) -> Generator[Tuple[dict, str, bool], None, None]:
+    uri, relative = _guess_location(item)
+    try:
+        metadata = item.to_dict()
+        if relative:
+            metadata = stac_transform(metadata)
+        else:
+            metadata = stac_transform_absolute(metadata)
+    except KeyError as e:
+        logging.error(f"Failed to handle item with KeyError: '{e}'\n The URI was {uri}")
```
omad, line 85: This exception handling looks wrong, it looks like it'll just blindly continue after logging an error.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -102,27 +125,34 @@ def stac_api_to_odc(
...
     # Do the indexing of all the things
     success = 0
     failure = 0
 
-    for dataset, uri in datasets_uris:
-        try:
-            index_update_dataset(
-                dataset,
-                uri,
+    sys.stdout.write("\rIndexing from STAC API...\n")
```
omad, line 132: Why `sys.stdout.write()` instead of logging?

whatnick (approved):

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -102,27 +125,34 @@ def stac_api_to_odc(
...
                 update_if_exists=update_if_exists,
-                allow_unsafe=allow_unsafe_changes,
-            )
-            success += 1
-        except Exception as e:
-            logging.warning(f"Failed to handle dataset: {uri} with exception {e}")
-            failure += 1
+                allow_unsafe=allow_unsafe,
+            ): item.id
+            for item in search.get_all_items()
```
omad, line 143: How many items might this be expected to process? If it's 'a-lot a-lot', this will chew a lot of RAM.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
whatnick: I believe logging.info disappears on some runtimes. I would love better logging levels support / structlog in these tools.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
alexgleith: You're right. Fixed.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
alexgleith: `get_all_items` is lazy loaded, so it's fine :-)

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
alexgleith: Logging with info means datacube indexing is noisy (indexing dataset xxx for every dataset) and doing it this way (with `\r`) means that we get a fast ticking single line output, which is nice.

omad: Okay, that does sound nice for interactive use, so long as it doesn't mask failures.

It would be nice to have comments starting that's what's happening. I don't know \r of the top of my head.

I also agree with Tish that a better logging/setup wouldn't go astray, with different configurations for interactive or automated deployments.

PR #287, opened 2021-08-02
Fix stac dependencies and cleanup test docker

Kirill888:
- declare minimum dependency on pystac 
- simplify and document test docker
- make sure tests can run outside of test docker (skip some if no db, set expected AWS env vars)
- use local http server for some tests to speed up, make more robust


.github/workflows/test-dc-tools.yml
```diff
@@ -24,8 +24,6 @@ jobs:
       run: |
         cd tests
         docker-compose up -d
-        # Manually install over the top to fix weird file symbolic links in CI
-        docker-compose exec -T tools-tester pip install -r /code/docker/rr-odc-tools.in
```
alexgleith, line 28: If we remove this, the apps don't get installed properly, so we can't run the boostrap script.

>Traceback (most recent call last):
  File "/env/bin/dc-sync-products", line 33, in <module>
    sys.exit(load_entry_point('odc-apps-dc-tools', 'console_scripts', 'dc-sync-products')())
  File "/env/bin/dc-sync-products", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/usr/lib/python3.8/importlib/metadata.py", line 503, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: odc-apps-dc-tools

This is ignored in the CI build, but for a different reason... the script should run successfully and fail!

I am ok with removing the bootstrap-odc.sh tools from CI testing, but @whatnick likes it in there.

Kirill888: The way test docker works is that it installs everything in edit mode at build time, so that step has been done already, then you just bring in new version of the code into the docker via mount and run tests against that.

You can just exec into docker and run tests in there, edit files outside of docker and run tests again without leaving your current exec session. This is how it was designed to run.

Only time this would fail is if test docker is stale. It becomes stale when:

- new odc library or app is added
- new dependency is added

And a proper solution to that is to make test docker rebuild a proper dependency of the running test stage.

So the question should be what is special about `dc-sync-products` that it refuses to run when installed in edit mode. Note that other tools from the same lib, like `dc-index-from-tar` work just fine without that extra line there. Also why nothing fails in the pipeline when this line is omitted?

@alexgleith @whatnick

alexgleith: I looked at it a bunch, and it's because on CI all the `-e` symbolic links get weird. I'm kind of surprised the tests work ok, actually.

But if you look at it, the installed binary for `dc-sync-products` can't find the python files it was registered with when running `-e` to install it. I bet that `dc-index-from-tar` doesn't work either.

Kirill888: Yes, it's because I was testing locally and haven't re-installed in edit mode after that app was added.

It's broken because of the way `entry_point`s are handled when installed in edit mode, basically package metadata is missing and so shell script doesn't work, as it has no entry points.

This should fix the "not running" part:

https://github.com/opendatacube/odc-tools/pull/287/commits/851e165a60330f68be82cb192fbfd8a638505cb9

But a proper way is to move re-install (with `--no-deps`) into docker entrypoint.

alexgleith: I'm pretty sure the underlying issue is docker-in-docker and magic file system abstractions. Installing with `-e` does work in most situations. GitHub Actions is very particular.

+1 to docker entrypoint, which is essentially the same as my solution.

alexgleith (approved):
Looks good! I like the docs and notes on how to do things.

.github/workflows/test-dc-tools.yml
Kirill888: seems to be running:

```
08/02/2021 06:49:45: INFO: Found 63 products in the CSV https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/master/workspaces/prod-products.csv
08/02/2021 06:49:45: INFO: Found 0 products in the Datacube
08/02/2021 06:49:45: INFO: Creating index: dix_aster_aloh_group_composition_lat_lon_time
08/02/2021 06:49:45: INFO: Creating index: dix_aster_aloh_group_composition_time_lat_lon
08/02/2021 06:49:45: INFO: Creating index: dix_aster_aloh_group_composition_platform
08/02/2021 06:49:45: INFO: Creating index: dix_aster_aloh_group_composition_instrument
08/02/2021 06:49:45: INFO: Added product aster_aloh_group_composition
08/02/2021 06:49:45: INFO: Creating index: 
```

Kirill888: @alexgleith 
> I'm pretty sure the underlying issue is docker-in-docker and magic file system abstractions. Installing with `-e` does work in most situations. GitHub Actions is very particular.
> 
> +1 to docker entrypoint, which is essentially the same as my solution.

When you `pip install -e .` for a module `<mylib>`, pip creates `<mylib>.egg-info/` directory with package metadata in the code folder, next to `setup.py` and also adds a text file `<mylib>.egg-link` in `site-packages` with an absolute path to the library baked into that file.

If you delete `.egg-info` directory (because clean checkout), you can still `import <mylib>`, because `.egg-link` file still exists (inside docker in this case, or just in your env on a dev machine), but extra information about the package, like `entry_points` which defines "console_scripts" will no longer work. As a result running `my-script` will error out, but running `python -m <mylib>.apps.app1` will still work, because imports work.

PR #288, opened 2021-08-02
Add dataset filters option to cli run

kieranricardo:
This PR adds the ability to specify limited filters for selecting scenes from tasks.

Atm users can specify a combination of or filters that match values of `properties` in the  metadata doc, and can check time ranges. This is the MVP to filter different satellites for different time ranges.


alexgleith:
Alchemist does filtering using Datacube Core functions, see: https://github.com/opendatacube/datacube-alchemist/blob/main/datacube_alchemist/cli.py#L102

And see: https://github.com/opendatacube/datacube-core/blob/develop/datacube/ui/click.py

I think we should only write a new filtering process if it's really a lot simpler... do you think this is?

kieranricardo:
@alexgleith I made this PR to handle queries like this:

```python
(date in first_half_of_year) or ((date in second_half_of_year) and (platform == landsat_8))
```

Is there a way to do this with `datacube` functionality? I was under the impression that there isn't, but if there is that would be a much better solution!

kieranricardo:
@alexgleith @omad 
Merging this for now to make testing `fc-percentiles` easier. But happy to discuss a different API and/or implementation in another issue/PR :)

PR #290, opened 2021-08-02
Pq ls atmos opacity masking for landsat 7 and 5

NikitaGandhi:
calculating aerosol masking for landsat 7 and 5 using SR_ATMOS_OPACITY band as an input measurement. 

Thanks @cbur24 for providing calculation here:

```
#Mask -9999 as NaNs, then multiply by scaling factor
opaque = ls7['SR_ATMOS_OPACITY'].where(ls7['SR_ATMOS_OPACITY'] != -9999) * 0.001

#Threshold opacity layer to set bad pixels to True
# Atmospheric opacity; < 0.1 = clear; 0.1 - 0.3 = average; > 0.3 = hazy
opaque = opaque > 0.3
```


alexgleith (approved):
Looks really neat to me, @NikitaGandhi !

cbur24 (approved):

PR #293, opened 2021-08-05
update masking to support third argument

NikitaGandhi:
update masking lib to support third argument so the operation can be performed in this order:
- morphological closing with r3 (>0)
- then opening with r1 (>0)
- followed by dilation with r2 (>0)

Doing this will fixes:
- overlap issue as the whole masking operation is performed using single dask task

### DIFF
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -385,7 +385,7 @@ def binary_closing(xx: xr.DataArray, radius: int = 1, **kw) -> xr.DataArray:
 
 def mask_cleanup_np(mask: np.ndarray, filter: Dict = dict(closing=0, opening=2, dilation=5)) -> np.ndarray:
     """
-    Given binary mask and apply morphological closing(>0) then opening(>0) followed by dilation(>0).
+    Apply morphological closing(>0) then opening(>0) followed by dilation(>0) on given binary mask.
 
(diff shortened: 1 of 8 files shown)


cbur24 (approved):

robbibt:
I wonder if we should consider renaming the arguments to make them clearer than `r1, r2, r3`, e.g. `opening, dilation, closing` (or ideally, reordering them too so that they follow the order they are run: `closing, opening, dilation`).

NikitaGandhi:
- Updated filter to dict instead of tuple. Use `0` if you want to omit the specific operation. It is backward compatible and support below options.
Usage: 
**dict(closing=2, opening=1, dilation=1)** - produce xx["mask_band_2_1_1"]
**dict(opening=1, dilation=1)** - produce xx["mask_band_1_1"]

- Updated stats gm implementation filter definition to dict - no impact
- Updated pq implementation filter definition to dict - no impact

cbur24 (approved):
Nice, this should help limit the confusion with the morphological operators

robbibt (approved):
Looks great to me!

kieranricardo (approved):
looks good!

NikitaGandhi:
Thanks @Kirill888. As this PR was closed already, I have incorporated this change into this PR https://github.com/opendatacube/odc-tools/pull/298

PR #294, opened 2021-08-06
Fc percentile nodata

kieranricardo:
This PR changes to the `percentile` code to copy the `nodata` attributes from the incoming dataset


supermarkion (approved):
All changes are FC percentile. No issue to crash other components.

LGTM on the nodata design.

PR #298, opened 2021-08-08
calculate bit masking using flags

### LINKED ISSUE #295: offset not applied correctedly in geomedian with mads output
https://github.com/opendatacube/odc-tools/blob/7cff7aaac9d74aaea34ee08f9a55f0302b0bcc53/libs/algo/odc/algo/_geomedian.py#L415

should be `offset = - offset/scale` (missing a negative sign).

NikitaGandhi:
- calculate bit masking using flags
- fix #295 
- cleanup and refactor code

### DIFF
--- a/libs/algo/odc/algo/_geomedian.py
+++ b/libs/algo/odc/algo/_geomedian.py
@@ -412,7 +412,7 @@ def geomedian_with_mads(
     if not is_float:
         gm_data = da.map_blocks(
             lambda x: from_float_np(
-                x, yxbt.dtype, nodata, scale=1 / scale, offset=offset / scale
+                x, yxbt.dtype, nodata, scale=1 / scale, offset=-offset / scale
             ),
             gm_data,
             dtype=yxbt.dtype,
--- a/libs/stats/odc/stats/_gm_ls_bitmask.py
+++ b/libs/stats/odc/stats/_gm_ls_bitmask.py
@@ -26,7 +26,6 @@ class StatsGMLSBitmask(StatsPluginInterface):
             # provide flags with high cloud bits definition
             flags: Dict[str, Optional[Any]] = dict(
                 cloud="high_confidence",
-                # cloud_shadow="high_confidence",
                 cirrus="high_confidence",
             ),
             nodata_flags: Dict[str, Optional[Any]] = dict(nodata=False),
@@ -159,7 +158,7 @@ class StatsGMLSBitmask(StatsPluginInterface):
         gm = geomedian_with_mads(xx, **cfg)
         gm = gm.rename(self.renames)
 
-        # Scale USGS Landsat bands into surface reflectance
+        # Rescaling gm bands between 0-10000 (approx) range
         # for band in gm.data_vars:
         #     if band in self.bands:
         #         gm[band] = scale * sr_scale * gm[band] + offset * sr_scale
--- a/libs/stats/odc/stats/_gm_ls_bitmask.py
+++ b/libs/stats/odc/stats/_gm_ls_bitmask.py
@@ -34,8 +34,8 @@ class StatsGMLSBitmask(StatsPluginInterface):
             aux_names=dict(smad="smad", emad="emad", bcmad="bcmad", count="count"),
             resampling: str = "nearest",
             work_chunks: Tuple[int, int] = (400, 400),
-            scale=1/0.0000275,
-            offset=-0.2/0.0000275,
+            scale: float = 1/0.0000275,
+            offset: float = -0.2/0.0000275,
             **other,
     ):
         self.mask_band = mask_band
--- a/libs/stats/odc/stats/_gm_ls_bitmask.py
+++ b/libs/stats/odc/stats/_gm_ls_bitmask.py
@@ -34,8 +34,8 @@ class StatsGMLSBitmask(StatsPluginInterface):
             aux_names=dict(smad="smad", emad="emad", bcmad="bcmad", count="count"),
             resampling: str = "nearest",
             work_chunks: Tuple[int, int] = (400, 400),
-            scale: float = 1/0.0000275,
-            offset: float = -0.2/0.0000275,
+            scale: float = 0.0000275,
+            offset: float = -0.2,
             **other,
     ):
         self.mask_band = mask_band
--- a/libs/stats/odc/stats/_gm_ls_bitmask.py
+++ b/libs/stats/odc/stats/_gm_ls_bitmask.py
@@ -2,18 +2,16 @@
 Landsat QA Pixel Geomedian
 """
 from functools import partial
-from typing import Dict, Optional, Tuple
+from typing import Dict, Optional, Tuple, Any
 
-import dask.array as da
 import xarray as xr
-import numpy as np
+from datacube.utils import masking
 from odc.algo import geomedian_with_mads, keep_good_only, erase_bad
 from odc.algo._masking import _xr_fuse, _first_valid_np, mask_cleanup, _fuse_or_np
 from odc.algo.io import load_with_native_transform
+from odc.stats import _plugins
+from odc.stats.model import StatsPluginInterface
 from odc.stats.model import Task
-
-from . import _plugins
-from .model import StatsPluginInterface
 
 
 class StatsGMLSBitmask(StatsPluginInterface):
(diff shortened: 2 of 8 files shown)


libs/algo/odc/algo/_geomedian.py
```diff
@@ -412,7 +412,7 @@ def geomedian_with_mads(
     if not is_float:
         gm_data = da.map_blocks(
             lambda x: from_float_np(
-                x, yxbt.dtype, nodata, scale=1 / scale, offset=offset / scale
+                x, yxbt.dtype, nodata, scale=1 / scale, offset=-offset / scale
```
alexgleith, line 415: I guess the offset was `0` in the S-2 GeoMAD, right? So there's no bug in the data we have already produced.

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -25,19 +23,32 @@ def __init__(
...
+                cirrus="high_confidence",
+            ),
+            nodata_flags: Dict[str, Optional[Any]] = dict(nodata=False),
             filter: Optional[Dict[str, int]] = None, # dict(closing=int,opening=int,dilation=int)
-            aux_names=dict(smad="sdev", emad="edev", bcmad="bcdev", count="count"),
+            aux_names=dict(smad="smad", emad="emad", bcmad="bcmad", count="count"),
             resampling: str = "nearest",
             work_chunks: Tuple[int, int] = (400, 400),
+            scale: float = 1/0.0000275,
+            offset: float = -0.2/0.0000275,
```
fangfy: default to `scale = 0.0000275` and `offset = -0.2 ` should be good now that we don't try to scale output to a different 10000 scaling.

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -25,19 +23,32 @@ def __init__(
             self,
             bands: Optional[Tuple[str, ...]] = None,
             mask_band: str = "QA_PIXEL",
+            # provide flags with high cloud bits definition
+            flags: Dict[str, Optional[Any]] = dict(
+                cloud="high_confidence",
+                # cloud_shadow="high_confidence",
```
cbur24: Shouldn't we have `cloud_shadow` masked by default @fangfy ?

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -148,14 +160,14 @@ def reduce(self, xx: xr.Dataset) -> xr.Dataset:
         gm = gm.rename(self.renames)
 
         # Scale USGS Landsat bands into surface reflectance
```
alexgleith: It's really only scaling to the range 0-10000 (approx). It's already surface reflectance. This is a pedantic note though!

libs/stats/tests/test_gm_ls_bitmask.py
```diff
@@ -31,10 +31,26 @@ def dataset():
         "y": np.linspace(0, 5, band_pq.shape[1]),
         "spec": index,
     }
-
-    data_vars = {"band_red": (("spec", "y", "x"), band_red), "QA_PIXEL": (("spec", "y", "x"), band_pq)}
-    attrs = dict(crs="epsg:32633", grid_mapping="spatial_ref")
-    xx = xr.Dataset(data_vars=data_vars, coords=coords, attrs=attrs)
+    pq_flags_definition = {'snow': {'bits': 5, 'values': {'0': 'not_high_confidence', '1': 'high_confidence'}},
```
alexgleith: Are you sure you can't get this from the ODC?

alexgleith (commented):
Looks good. A couple of little comments.

libs/algo/odc/algo/_geomedian.py
NikitaGandhi: Yes this never got picked up as s-2 run without any offset

Kirill888:
I just want to say that I find change to "dict" a bit confusing. For starters, `closing` and `opening` both use `dilation` and `erosion` internally (applied in different order), also unlike tuple, values in dictionary do not have native order, so it's harder to understand what the operation will do when looking at the call site of the function.

If you need to generalize this code to support longer chains of morphological operations, then I would suggest to use something like this 

```python
def mask_cleanup(
    mask: xr.DataArray,
    filter: Iterable[Tuple[str, int]] = [("opening", 2), ("dilation", 5)],
    name: Optional[str] = None
) -> xr.DataArray:
```

So you supply operations you want performed, in the order you want them performed. This allows to actually use any of this operations more than once. And reads much better at call site. Even better use enumeration instead of strings for `opening, closing, dilation, erosion`.

Kirill888:
`opening(R)` is the same as `erosion(R) |> dilation(R)`. So 

```
opening(3) |> dilation(5)
```
 
becomes 

```
erosion(3) |> dilation(3) |> dilation(5)
```
which is the same as 

```
erosion(3) |> dilation(8)
```

It might be that performing two smaller dilations one after another is faster than running single bigger dilation once, so operations done as they are now is probably fine, but this is why I find this interface via dictionary of parameters confusing.

libs/stats/tests/test_gm_ls_bitmask.py
robbibt: It should be accessible via something like `xx.pixel_quality.flags_definition`, yep

libs/stats/odc/stats/_gm_ls_bitmask.py
fangfy: > Shouldn't we have `cloud_shadow` masked by default @fangfy ?

That was one of the test options I used because I notice a lot of false cloud shadow in the city. However I agree maybe in the end, we will still just mask all cloud_shadow.

libs/stats/tests/test_gm_ls_bitmask.py
fangfy: > I just want to say that I find change to "dict" a bit confusing. For starters, `closing` and `opening` both use `dilation` and `erosion` internally (applied in different order), also unlike tuple, values in dictionary do not have native order, so it's harder to understand what the operation will do when looking at the call site of the function.
> 
> If you need to generalize this code to support longer chains of morphological operations, then I would suggest to use something like this
> 
> ```python
> def mask_cleanup(
>     mask: xr.DataArray,
>     filter: Iterable[Tuple[str, int]] = [("opening", 2), ("dilation", 5)],
>     name: Optional[str] = None
> ) -> xr.DataArray:
> ```
> 
> So you supply operations you want performed, in the order you want them performed. This allows to actually use any of this operations more than once. And reads much better at call site. Even better use enumeration instead of strings for `opening, closing, dilation, erosion`.

I like this suggestion if we can actually infer the order and type of operation from the parameter supplied. I agree that in the end it's just a series of erosion and dilation options. Closing of 5, opening of 5 and dilation of 5 would be the same as dilation of 5, erosion of 10 and dilation of 10.

NikitaGandhi: This is to build a new dataset object (e.g. xx) as an input dataset for geomad/pq testing. 

So I was thinking to get the definition from product unless there is a better option?

Kirill888: @fangfy order is known, since `Iterable` is an ordered sequence, it could be a list or a tuple. So the question is what's the best way to represent "operataion+parameter", options would be

- `Tuple[str, int]`
- `Tuple[Enum, int]`
- Custom class

robbibt: > I just want to say that I find change to "dict" a bit confusing. For starters, `closing` and `opening` both use `dilation` and `erosion` internally (applied in different order), also unlike tuple, values in dictionary do not have native order, so it's harder to understand what the operation will do when looking at the call site of the function.
> 
> If you need to generalize this code to support longer chains of morphological operations, then I would suggest to use something like this
> 
> ```python
> def mask_cleanup(
>     mask: xr.DataArray,
>     filter: Iterable[Tuple[str, int]] = [("opening", 2), ("dilation", 5)],
>     name: Optional[str] = None
> ) -> xr.DataArray:
> ```
> 
> So you supply operations you want performed, in the order you want them performed. This allows to actually use any of this operations more than once. And reads much better at call site. 

> I like this suggestion if we can actually infer the order and type of operation from the parameter supplied. I agree that in the end it's just a series of erosion and dilation options. Closing of 5, opening of 5 and dilation of 5 would be the same as dilation of 5, erosion of 10 and dilation of 10.

@fangfy @Kirill888 I agree, the advantage of generalising it and allowing the user to define order is that it would allow more flexibility to go from something like `opening > dilation` as an initial test, to optimising to something like `erosion > dilation` for performance reasons once they're happy the first option does what they want

fangfy: > @fangfy order is known, since `Iterable` is an ordered sequence, it could be a list or a tuple. So the question is what's the best way to represent "operataion+parameter", options would be
> 
> * `Tuple[str, int]`
> * `Tuple[Enum, int]`
> * Custom class

true, I guess if you provide "binary_closing" instead of "closing", you can directly apply and chain the operations? 
(BTW did I leave comment in the wrong thread? sorry not sure how that happened... )

NikitaGandhi: All are very good suggestions. As this PR is not targeting this work, I have raise new issue ticket here - https://github.com/opendatacube/odc-tools/issues/300 so we can discuss it separately :)

fangfy: sounds good. I suggest we merge this PR and worry about improving that later.

fangfy (approved):

alexgleith (approved):

PR #304, opened 2021-08-10
Change version management strategy #299

### LINKED ISSUE #299: Version management and PyPI publishing
I would like to change the way libraries in `odc-tools` are versioned. Currently `setuptools_scm` is used to automatically compute version information from tags in git history. The primary downside of this approach is that all sub-modules share common version since they all reside in the same git repo.

I would like to switch to more manual version management approach, this would allow us to release different sub-modules independently of each other. I still expect to retain "release to packages.dea.ga.gov.au on every commit to develop branch" behaviour. This can be achieved using github action build number to provide an increment `0.2.dev{build-number-for-this-lib}`.

I think this change is a prerequisite for publishing to PyPI.

Kirill888:
1. Use `setup.cfg` to configure setuptools instead of `setup.py`
2. Add `__version__` property to every package
3. Use manual version management instead of `setuptools_scm`
4. Use github actions "run number" instead of git to increment patch
   component of the version number on every commit to develop branch
5. Include long_description from README.md for every package
6. Use `packages = find_namespace:` directive for automatic package finding

Closes #299


alexgleith (approved):
Looks super clean and I'm sure it works. From my take, it's taking the hardcoded value (like `0.2.`) and adding 'run number' to it, resulting in 0.2.123, is that about right?

I think it would be a good idea to document this process, for future folks. Approving anyway.

Kirill888:
> Looks super clean and I'm sure it works. From my take, it's taking the hardcoded value (like `0.2.`) and adding 'run number' to it, resulting in 0.2.123, is that about right?

yes, except it generates `0.2.dev123`, and then when we release to pypi it will become `0.2.0`, `dev` prefix is important because:

```python
from packaging import version
assert version.parse('0.2.0') > version.parse('0.2.dev123')
```

PR #308, opened 2021-08-11
Dilate nodata and non-contiguity bita

kieranricardo:
This branch dilates the nodata and non-contiguity bits when calculating a wofs summary. This more aggressively masks out regions in wofs to hopefully yield a cleaner summary.

This branch is mostly here for testing, this may lead to unacceptable level of data loss.

@omad I won't be merging this but I'm creating a review just to check if this is right


omad:
Can you please add a reference link and/or description of how this is calculated.

omad:
I'm pretty sure is right, and, that you don't want to close this @kieranricardo , so, I'm going to close this.

Issue #309, opened 2021-08-12
Task: establish release process for publishing to PyPI


Kirill888:
Current State
==========

On every PR merge to `develop` branch all sub-modules are rebuilt with automatically generated `dev` version number and then uploaded to `https://packages.dea.ga.gov.au/` (via a push to S3 bucket under Digital Earth Australia AWS account).

This allows installation via `pip install` so long as `--extra-index-url="https://packages.dea.ga.gov.au/"` option is supplied. This works fine and we intend to keep pushing to `packages.dea.` for every merge to `develop` branch. But there are some issues with this as well:

- Version changes for every package even if no code was changed in that particular module
- Extra steps for installation since it's not on pypi
- Potential security concern if someone uploads malicious versions of `odc-` packages to PyPI

Publishing to PyPI
==============

There are no major blockers to this. But there are some issues to iron out for automating this process. Main issue is that we have multiple packages under one git repository, as a result there is no obvious way to do release tagging.

There are several approached we can take:

1. Keep common versions across all libs. This simplifies release process at the expense of bumping versions of packages that have not really changed.
2. Don't use git tags and instead trigger release when `_version.py` for a given module changes, this would allow independent version management, but makes it harder to track from what point in git history was used to make a release.
3. Use per library tags. For example to release `odc.algo` with version `0.1.2`, edit `odc/algo/_version.py` and create tag named `algo-0.1.2`. We can then automate that with GitHub Actions workflow that can be manually triggered by an operator.

Immediate Actions
==============

I would like to manually push all the `odc.*` libs to PyPI with version `0.2.0a`, current latest version in `packages.dea.*` is `0.2.dev2249`. I will then change automatic version numbering for `dev` releases to be 

```
{major}.{minor}.{micro+1}-dev{BuildNumber}
```
so that `dev` releases take precedence over what's on PyPI, next dev version will then be:

```
0.2.1-dev{BuildNumber}
```

And version recorded in `odc/{lib}/_version.py` will match the last release to PyPI, so will contain `0.2.0a` in there.

uchchwhash:
Option 1 looks pretty good to me!

Kirill888:
fyi @gadomski

gadomski:
From a semantic versioning perspective, option 1 could be problematic once (if?) **odc-tools** hits v1. If one package needs a breaking change, it wouldn't really be correct to increment the major version for _all_ **odc-tools** packages.

[...]

PR #310, opened 2021-08-12
Change the way version is patched for dev releases (#309)

Kirill888:
- `_version.py` contains latest released version
- `{major}.{minor}.{micro}[a|b]` is edited to be
  `{major}.{minor}.{micro+1}-dev{build-number}` for dev release

This way dev releases are guaranteed to have more recent version than the last
released version, and `_version.py` files contain last released version and not
"next version".


omad (approved):
Sounds good.
Probably should describe the versioning in the patch script and readme too though.

PR #311, opened 2021-08-13
Updating installation docs

Kirill888:
- removed `--extra-index-url=...`
- use kebab-case for package names consistently
- add section for conda users


alexgleith (approved):
Awesome!

PR #312, opened 2021-08-16
Make STAC region_code and product_name finding more generic

alexgleith:


libs/index/odc/index/stac.py
```diff
@@ -28,21 +30,48 @@
 }
 
 
+def _get_region_code(properties: Dict[str, Any]) -> str:
```
Kirill888, line 40: My main concern with this approach is how to generalize/extend it. There are two modes here

1. Simple mode: some property if present maps to a string that contains region code
2. Landsat mode: several keys need to be present then some code needs to be supplied to form `region_code`

I would do something along these lines:
```python
candidates = ["odc:region_code", "s2:mgrs_tile", "io:supercell_id"]  # Extend by adding more candidates for new data sources
for rc_key in candidates:
   rc = properties.get(rc_key, None)
   if rc is not None:
       return rc
# do landsat later
```

```

File: libs/index/odc/index/stac.py [hunk 2]
```diff
@@ -28,21 +30,48 @@
...
 def _stac_product_lookup(
     item: Document,
 ) -> Tuple[str, Optional[str], str, Optional[str], str]:
     properties = item["properties"]
-    platform = properties.get("eo:platform", properties.get("platform", None))
 
     dataset_id: str = item["id"]
     dataset_label = item.get("title")
-    product_name = get_in(["odc:product"], properties, platform)
+    product_name = get_in(["odc:product"], properties, None)
```
Kirill888: Why `get_in` instead of plain `.get`?

libs/index/odc/index/stac.py
alexgleith: Yeah, I was thinking that too. I'll fix it up.

Kirill888 (approved):
:+1:

PR #313, opened 2021-08-17
Add asset rewrite option

alexgleith:
* Enables rewriting from `https://` to `s3://` or `https://example.com/` to `/vsiaz/` for all assets

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -241,9 +241,9 @@ def cli(
         config["max_items"] = limit
 
     if rewrite_assets is not None:
-        rewrite = list(rewrite_assets.split(";"))
+        rewrite = list(rewrite_assets.split(","))
         if len(rewrite) != 2:
-            raise ValueError("Rewrite assets argument needs to be two strings split by ';'")
+            raise ValueError("Rewrite assets argument needs to be two strings split by ','")
 
     # Do the thing
     dc = Datacube()
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -225,6 +225,7 @@ def cli(
     Iterate through STAC items from a STAC API and add them to datacube.
     """
     config = _parse_options(options)
+    rewrite = None
 
     # Format the search terms
     if bbox:
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -45,7 +45,9 @@ def _parse_options(options: Optional[str]) -> Dict[str, Any]:
     return parsed_options
 
 
-def _guess_location(item: pystac.Item) -> Tuple[str, bool]:
+def _guess_location(
+    item: pystac.Item, rewrite: Optional[Tuple[str, str]] = None
+) -> Tuple[str, bool]:
     self_link = None
     asset_link = None
     relative = True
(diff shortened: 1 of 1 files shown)


apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -45,7 +45,9 @@ def _parse_options(options: Optional[str]) -> Dict[str, Any]:
     return parsed_options
 
 
-def _guess_location(item: pystac.Item) -> Tuple[str, bool]:
+def _guess_location(
+    item: pystac.Item, rewrite: Optional[Tuple[str, str]] = None
```
Kirill888, line 49: Just replacing prefix of a URL with another is not always sufficient, I would recommend rewrite to have type
`: Optional[Callable[[str], str]]`. For `http[s]->s3` one needs to undo url encoding applied to the path component and not just replace `https://some.host.tld/maybe_prefix/` with `s3://bucket/`. For example character `+` is encoded as `%2B` in http urls, but is just a `+` in S3 urls.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -57,12 +59,18 @@ def _guess_location(item: pystac.Item) -> Tuple[str, bool]:
...
 
-    for name, asset in item.assets.items():
+    for _, asset in item.assets.items():
         if "geotiff" in asset.media_type:
             asset_link = os.path.dirname(asset.href)
             break
 
+    if rewrite is not None:
+        for _, asset in item.assets.items():
+            if "geotiff" in asset.media_type:
```
Kirill888, line 71: I know this is used through out the code, but I think this is wrong way to go, why are we assuming that only geotiff image type will be used with STAC. I think we should look for `data` role rather than filtering on media type.
And in this specific case why not rewrite metadata urls as well as imagery urls?

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -216,10 +240,15 @@ def cli(
     if limit is not None:
         config["max_items"] = limit
 
+    if rewrite_assets is not None:
+        rewrite = list(rewrite_assets.split(";"))
```
Kirill888: in your help `,` character is used, but this one is `;`, I'd say `,` is more command line friendly

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -216,10 +240,15 @@ def cli(
...
+        if len(rewrite) != 2:
+            raise ValueError("Rewrite assets argument needs to be two strings split by ';'")
```
Kirill888: `;` -> `,`

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
alexgleith: We can't just stick to `data` role, as other assets have roles like `metadata` for SCL.

In fact, the Landsat STAC isn't specifying `data` for the data bands! https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-8-c2-l2/items

alexgleith: For metadata URLs, some are direct links to .json files on a http/s URI, and may be useful there still.

Maybe it's not going to be possible to handle generically like this.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
alexgleith: Ok, fixing.

apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
alexgleith: How would you pass that in from the command line?

Kirill888: you won't, the only supported method would be `http->s3` (and maybe reverse?), and configuration for that would be "prefix" to replace and with "what", but it will also need to call `from urllib.parse import unquote` on the path. Although I'm not sure what characters are allowed in s3 urls, I know for sure that issue `+` vs `%2B` exists.

Kirill888: in http url `+` is ` `, but in `s3` url `+` is just `+`, and ` ` is allowed "as is".

Kirill888 (approved):
will do for now, assuming people don't use awkward characters in urls, but still rewrite as function makes more sense

PR #314, opened 2021-08-18
Make test docker self-contained

Kirill888:
- install db into docker
- run tests as non-root user with UID:GID matching that of code checkout (to avoid anoying root:root files)
- move coverage config file


alexgleith:
I'm pretty strongly against running postgres in the same container. It is against the whole Docker paradigm.

Using Docker Compose means we get a simple way to run a separate, simple Postgres service and doing it inside the container doesn't provide any benefits that I can see.

I agree that the UID issue is a pain on Linux and I don't know an elegant way to solve it, so I'm ok with your proposals in that regard.

Kirill888:
@alexgleith this is not a deployment, this a developer tool for developers. Unlike docker-compose solution this one offers following advantages
- Only one docker image to pull in from outside or build locally
- No need to install extra tools, not matter how prevalent docker-compose is, it is an extra tool
- Faster access to database as unix sockets are used and not TCP
- No need to setup db passwords as unix sockets are used and not TCP
- Database state is wiped clean on every test run
- Doesn't leave docker processes hanging around when "done"
- Can just exec into it and keep running tests just the same as exec into compose

Kirill888:
If you see advantages to using docker-compose based workflows you can continue to do so as you can bring outside db for testing, you are not limited to using local instance of it, so the only disadvantage is a tiny increase in docker image size.

docker/Makefile
```diff
@@ -84,10 +86,11 @@ dkr-no-deps:
 run-test:
 	@docker run --rm \
     -v $(CODE):/code \
-    $(DKR_IMG) pytest --timeout=30 .
+    $(DKR_IMG) with-test-db env AWS_DEFAULT_REGION=us-west-2 \
```
gadomski, line 89: @Kirill888 so make me smart, running tests isn't an entirely local process, it touches AWS as well?

Kirill888: No we do not connect to aws, we use moto library, but it still needs default region defined, although it should probably be defined in a fixture and not here.

I do think that some tests reach out for external resources via http, so it's not fully self contained, we should fix that

gadomski: Roger, sounds good. I know pystac-client has had good success with https://pytest-vcr.readthedocs.io/en/latest/ for robustifying network-y tests.

Kirill888: that one looks cool, we just use `pytest.httpserver`, but not everywhere, some places grab files from github, making tests easy to write is a major undertaking

Kirill888:
@alexgleith with this change one can start running integration tests without any extra knowledge as all required database setup steps are done by a script, current setup with compose on the other hand does not do that, it just provides an empty database with some default credentials and requires extra steps to run, you can only find those extra steps in some workflow yaml. Can you please have a second look on this?

alexgleith:
I can see the advantage of not having a lingering container on a dev box. It feels complicated, but if it works, then I'm ok with it.

alexgleith (approved):

Kirill888:
thanks, most complexity is actually in matching UID to the outside, db part is fairly small and not all that more than what compose requires.

PR #317, opened 2021-08-20
Add eodataset3 to odc-tools

supermarkion:
We need the eodatasets3 in develop branch firstly. 

The feature branch always grab the test-image from develop branch to run test. So I cannot add new library to test-image in feature branch.

All test passed, only the code codecov fail.

### DIFF
--- a/docker/constraints.txt
+++ b/docker/constraints.txt
@@ -143,6 +143,7 @@ pyparsing==2.4.7
 pyproj==3.1.0
 pyrsistent==0.18.0
 pystac==1.1.0
+eodatasets3==0.22.0
 pystac-client==0.2.0
 pytest==6.2.4
 pytest-cov==2.12.1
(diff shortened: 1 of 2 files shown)


kieranricardo (approved):
LGTM :)

Kirill888:
agree it's a bit of a pain, ideally one would always build docker image as described by the current branch to run the tests, but that can be slow and might also fail for random external reasons that have nothing to do with the current changes. I guess we could do a more complex workflow where we 

1. Check if feature branch has changes that are likely to require new docker image (files changed under `docker/**` or any `setup.cfg` files when diffing with `develop` branch)
   - YES: build fresh docker from this branch
   - NO: just pull in default test docker (saves time and reduces probability of dealing with transient issues)

And if working on `develop` branch already, similar rules can apply but instead of diffing branches you diff files changed since last build, github provides enough information to figure this out programmatically.

PR #321, opened 2021-08-23
Added extra SQS functions:

carioca-au:
- get_queue_attributes
- list_queues

### DIFF
--- a/libs/aws/odc/aws/queue.py
+++ b/libs/aws/odc/aws/queue.py
@@ -2,6 +2,15 @@ import boto3
 import itertools
 from typing import Mapping, Any, Iterable, Optional
 
+from odc.aws import mk_boto_session
+
+
+def _create_sqs_client(
+    region: Optional[str] = None
+):
+
+    return mk_boto_session().create_client("sqs", region)
+
 
 def redrive_queue(
     queue_name: str,
@@ -74,7 +83,7 @@ def redrive_queue(
     return count
 
 
-def get_queue(queue_name: str):
+def get_queue(queue_name: str, region: Optional[str] = None):
     """
     Return a queue resource by name, e.g., alex-really-secret-queue
     """
@@ -83,25 +92,25 @@ def get_queue(queue_name: str):
     return queue
 
 
-def list_queues(region: Optional[str]):
+def list_queues(region: Optional[str] = None):
     """
     Return a list of queues which the user is allowed to see
     """
-    client = boto3.client('sqs', region_name=region)
-    queues = client.list_queues()
-    return queues
+    sqs = _create_sqs_client(region=region)
+    return sqs.list_queues()
 
 
-def get_queue_attributes(queue_name: str, region: Optional[str], attribute: Optional[str]) -> dict:
+def get_queue_attributes(queue_name: str, region: Optional[str] = None, attribute: Optional[str] = None) -> dict:
     """
     Return informed queue's attribute or a list of queue's attributes when attribute isn't informed
     """
 
-    if attribute is None:
-        attribute = 'All'
+    sqs = _create_sqs_client(region=region)
 
-    client = boto3.client('sqs', region_name=region)
-    attributes_dict = client.get_queue_attributes(
+    if attribute is None:
+        attribute = ['All']
+
+    attributes_dict = sqs.get_queue_attributes(
         QueueUrl=queue_name,
         AttributeNames=[attribute]
     )
--- a/libs/aws/odc/aws/queue.py
+++ b/libs/aws/odc/aws/queue.py
@@ -83,6 +83,31 @@ def get_queue(queue_name: str):
     return queue
 
 
+def list_queues(region: str):
+    """
+    Return a list of queues which the user is allowed to see
+    """
+    client = boto3.client('sqs', region_name=region)
+    queues = client.list_queues()
+    return queues
+
+
+def get_queue_attributes(queue_name: str, region: str, attribute: Optional[str]) -> dict:
+    """
+    Return informed queue's attribute or a list of queue's attributes when attribute isn't informed
+    """
+
+    if attribute is None:
+        attribute = 'All'
+
+    attributes_dict = client = boto3.client('sqs', region_name=region)
+    client.get_queue_attributes(
+        QueueUrl=queue_name,
+        AttributeNames=[attribute]
+    )
+    return attributes_dict
+
+
 def publish_message(queue, message: str, message_attributes: Mapping[str, Any] = {}):
     """
     Publish a message to a queue resource. Message should be a JSON object dumped as a
--- a/libs/aws/odc/aws/queue.py
+++ b/libs/aws/odc/aws/queue.py
@@ -97,24 +97,38 @@ def list_queues(region: Optional[str] = None):
     Return a list of queues which the user is allowed to see
     """
     sqs = _create_sqs_client(region=region)
-    return sqs.list_queues()
+    list_queues_return = sqs.list_queues()
+
+    return list_queues_return.get('QueueUrls', [])
 
 
 def get_queue_attributes(queue_name: str, region: Optional[str] = None, attribute: Optional[str] = None) -> dict:
     """
     Return informed queue's attribute or a list of queue's attributes when attribute isn't informed
+    Valid attribute options:
+        'ApproximateNumberOfMessages',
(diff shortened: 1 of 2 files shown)


libs/aws/odc/aws/queue.py
```diff
@@ -83,6 +83,31 @@ def get_queue(queue_name: str):
...
+
+def get_queue_attributes(queue_name: str, region: str, attribute: Optional[str]) -> dict:
+    """
+    Return informed queue's attribute or a list of queue's attributes when attribute isn't informed
+    """
+
+    if attribute is None:
+        attribute = 'All'
+
+    attributes_dict = client = boto3.client('sqs', region_name=region)
```
alexgleith: This is not right.

libs/aws/odc/aws/queue.py
```diff
@@ -83,6 +83,31 @@ def get_queue(queue_name: str):
     return queue
 
 
+def list_queues(region: str):
+    """
+    Return a list of queues which the user is allowed to see
+    """
+    client = boto3.client('sqs', region_name=region)
+    queues = client.list_queues()
```
alexgleith: I think this should re-use `get_queue`

```
def get_queue(queue_name: str):
    """
    Return a queue resource by name, e.g., alex-really-secret-queue
    """
    sqs = boto3.resource("sqs")
    queue = sqs.get_queue_by_name(QueueName=queue_name)
    return queue
    ```

**Review by alexgleith — CHANGES_REQUESTED** [review]
Please look at the other functions that are in this file and align to them.

Also please write tests.

File: libs/aws/odc/aws/queue.py (hunk 2 above)
Comment (carioca-au) [reply libs/aws/odc/aws/queue.py]: It does not work. 
**get_queue** uses _**resource("sqs")**_ which returns a **sqs.Queue** object not a **client**, therefore **sqs.Queue** does not have **list_queues** function, this must come from a client.

Comment (carioca-au) [reply libs/aws/odc/aws/queue.py]: image
image

File: libs/aws/odc/aws/queue.py (hunk 1 above)
Comment (carioca-au) [reply libs/aws/odc/aws/queue.py]: this is totally wrong sorry

Comment (carioca-au) [reply libs/aws/odc/aws/queue.py]: Fixed

File: libs/aws/odc/aws/queue.py [hunk 3]
```diff
@@ -2,6 +2,15 @@
...
 from typing import Mapping, Any, Iterable, Optional
 
+from odc.aws import mk_boto_session
+
+
+def _create_sqs_client(
+    region: Optional[str] = None
+):
+
+    return mk_boto_session().create_client("sqs", region)
```
alexgleith: This is completely different

libs/aws/odc/aws/queue.py
```diff
@@ -74,7 +83,7 @@ def post_messages(to_queue, messages):
     return count
 
 
-def get_queue(queue_name: str):
+def get_queue(queue_name: str, region: Optional[str] = None):
```
alexgleith: Is this being used?

libs/aws/odc/aws/queue.py
```diff
@@ -83,6 +92,31 @@ def get_queue(queue_name: str):
...
 
+def list_queues(region: Optional[str] = None):
+    """
+    Return a list of queues which the user is allowed to see
+    """
+    sqs = _create_sqs_client(region=region)
+    return sqs.list_queues()
+
+
+def get_queue_attributes(queue_name: str, region: Optional[str] = None, attribute: Optional[str] = None) -> dict:
```
alexgleith: Perhaps attribute should be attributes (a list)

libs/aws/odc/aws/queue.py
```diff
@@ -83,6 +92,31 @@ def get_queue(queue_name: str):
...
+    """
+    Return informed queue's attribute or a list of queue's attributes when attribute isn't informed
+    """
+
+    sqs = _create_sqs_client(region=region)
```
alexgleith: Why not `queue = get_queue(...`

and then https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.get_queue_attributes

libs/aws/odc/aws/queue.py
alexgleith: Ah,  you're right about this one, sorry.

alexgleith: You could use a resource still, and filter, like this says: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.ServiceResource.queues

libs/aws/odc/aws/queue.py
carioca-au: image

carioca-au: however it's possible to do 
`queue = get_queue(...` 
`queue.attributes` / `queue.attributes.get('<attribute_wanted>')`

But in this case, you are retrieving all attributes every time, and the other way you are retrieving just the attribute that you want to. It's a bit less expensive request

libs/aws/odc/aws/queue.py
carioca-au: Changed

libs/aws/odc/aws/queue.py
carioca-au: this was creating a client from a new session, there are examples of that around the code. However, due to the changes recommended it's not necessary anymore and was removed.

libs/aws/odc/aws/queue.py
carioca-au: This function will always return a dict:
image

carioca-au:
> Please look at the other functions that are in this file and align to them.
> 
> Also please write tests.

Tests were done and functions were changed

libs/aws/odc/aws/queue.py
```diff
@@ -83,6 +83,39 @@ def get_queue(queue_name: str):
...
 
+def list_queues(region: Optional[str] = None):
+    """
+    Return a list of queues which the user is allowed to see
+    """
+    sqs = boto3.resource("sqs")
+    return list(sqs.queues.all())
+
+
+def get_queue_attributes(queue_name: str, attribute: Optional[str] = None) -> dict:
```
alexgleith, line 94: I think this should accept a queue, not a string.

libs/aws/odc/aws/queue.py
```diff
@@ -83,6 +83,39 @@ def get_queue(queue_name: str):
     return queue
 
 
+def list_queues(region: Optional[str] = None):
+    """
+    Return a list of queues which the user is allowed to see
+    """
+    sqs = boto3.resource("sqs")
+    return list(sqs.queues.all())
```
alexgleith, line 91: Do you think it might be better to just use the built in generator, instead of converting it to a list?

Also, what about filtering? This has a `filter` method https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.ServiceResource.queues

alexgleith (approved):
Looks ok. Could be improved still, but I think it will be fine.

The test only tests for success, doesn't check attributes that don't exist.

And it also required getting all the queues locally to filter them.

libs/aws/odc/aws/queue.py
carioca-au: I'm following what other functions already do such as :
[get_queue](https://github.com/opendatacube/odc-tools/blob/ad937a7aca4af5519ac03a034ff2a5cd97499637/libs/aws/odc/aws/queue.py#L81)

if I receive a queue as a parameter, the user must have to use **get_queue** before calling my function and will have to inform the queue name anyways.

libs/aws/odc/aws/queue.py
carioca-au: I thought using a list would be an easier way to deal with the data, once that I don't believe this function will ever return that many queues.
About the **filter**, the advantage would be to provide a **prefix** filtering the queues by that. However, the point of this function is to return all queues available to the requester.
In my case, I will get just dead queues and I'll have to filter that once I have all queues.

libs/aws/odc/aws/queue.py
alexgleith: Look at `get_messages`

libs/aws/odc/aws/queue.py
alexgleith: Ah I see, a prefix filter is no good when we're putting a suffix on.

PR #322, opened 2021-08-23
Feature TCW Percentile

kieranricardo:


alexgleith:
The Tassled Cap algorithm produces other measures, so Brightness, Greeness and Wetness.

I reckon we should implement all three for this!

https://en.wikipedia.org/wiki/Tasseled_cap_transformation

libs/stats/odc/stats/_tcw_percentiles.py
```diff
@@ -0,0 +1,92 @@
...
+        Loads data in its native projection.
+        """
+
+        bad = (xx["fmask"] & 0b0000_1101) != 0
+        xx = xx.drop_vars(["fmask"])
+        
+        for band in xx.data_vars.keys():
+            bad = bad | (xx[band] == -999)
+
+        tcw = 0.0135 * xx['blue'] + 0.2021 * xx['green'] + 0.3102 * xx['red'] + 0.1584 * xx['nir'] - 0.6806 * xx['swir1'] - 0.6109 * xx['swir2']
```
emmaai: Can we make those coefficient as input parameters? It'd be easier if we need the other indices. @alexgleith made a fair point on it.

libs/stats/odc/stats/tasks.py
```diff
@@ -210,7 +211,15 @@ def msg_default(msg):
         if isinstance(temporal_range, str):
             temporal_range = DateTimeRange(temporal_range)
 
-        if dss is None:
+        if '-' in product:
```
emmaai, line 241: If I understand this correctly. It's gathering datasets for multiple output products. I'll assume it's needed in mangroves or it's an attempt to run tcw percentile on fly? Is the component of `product` different from product name we see in the output such that there is no `-` in each one?

kieranricardo: yep that's right. I've added it here so that `tcw` can use all the landsat products simultaneously. AFAIK the input product names aren't anywhere in the output

kieranricardo:
@alexgleith @emmaai thanks for your feedback! should the brightness, greenness, and wetness tassled cap measurements be in the same or different products?

alexgleith:
> @alexgleith @emmaai thanks for your feedback! should the brightness, greenness, and wetness tassled cap measurements be in the same or different products?

In my view, they should all be in the one product.

The algorithm is Tassled Cap and the outputs (measurements) are Wetness, Greenness and Brightness.

emmaai:
> @alexgleith @emmaai thanks for your feedback! should the brightness, greenness, and wetness tassled cap measurements be in the same or different products?

I don't think we need greenness or bightness anywhere, especially  the percentile, yet. Only mangroves need the wetness. I'm guessing we don't publish tc any percentile as a formal product? It's only temporary? If that's the case, no need for those two. Or my guess is wrong?

kieranricardo:
In that case I'll make the coefficients a parameter like you suggested @emmaai

libs/stats/odc/stats/_tcw_percentiles.py
```diff
@@ -0,0 +1,95 @@
...
+    NAME = "ga_tcw_percentiles"
+    SHORT_NAME = NAME
+    VERSION = "0.0.1"
+    PRODUCT_FAMILY = "twc_percentiles"
+
+    def __init__(
+        self,
+        resampling: str = "bilinear",
+        coefficients: Dict[str, float] = {
+            "red": 0.0135, "blue": 0.2021, "green": 0.3102, "nir": 0.1584, "swir1": -0.6806, "swir2": -0.6109
```
emmaai: Sorry my handwriting was so bad and even wrote some  figures wrong... Can you use the line below? Thanks.
```                 
'wetness': {'blue': 0.0315, 'green': 0.2021, 'red': 0.3102, 'nir': 0.1594,
                             'swir1': -0.6806, 'swir2': -0.6109}
```

libs/stats/odc/stats/_tcw_percentiles.py
```diff
@@ -0,0 +1,95 @@
...
+        _measurments = ["tcw_pc_10", "tcw_pc_50", "tcw_pc_90"]
+        _measurments
+        return _measurments
+
+    def _native_tr(self, xx):
+        """
+        Loads data in its native projection.
+        """
+
+        bad = (xx["fmask"] & 0b0000_1101) != 0
```
emmaai: We need contiguity in the mask as well  bad:= `nbart_contiguity == 0`. Sorry I missed that in our conversation.
```
-   aliases:
    - nbart_contiguity
    dtype: uint8
    flags_definition:
        contiguous:
            bits:
            - 0
            values:
                '0': false
                '1': true
    name: oa_nbart_contiguity
    nodata: 255
    units: '1'
```

libs/stats/odc/stats/tasks.py
emmaai: OK...I got a bit confused then. Is `product` the output product or input product, from so far what I read, it's output product (I could be wrong). I'll assume it's product name intuitively, e.g., in this case `tcw-percentiles`. so when you do `product.split('-')`, we got `tcw` and `percentile`? Can you give me an output string of `product` if possible?

emmaai (changes requested):
Can you please change the parameters and the mask in `_tcw_percentiles.py`? Thank you.

libs/stats/odc/stats/_tcw_percentiles.py
kieranricardo: no worries, I think I copied some down wrong too 🤦

kieranricardo:
@emmaai that's for input products. In this case I use the 3 landsat product: `ga_ls8c_ard`,  `ga_ls7e_ard` and  `ga_ls5t_ard`. And I submit the product string `"ga_ls8c_ard-ga_ls7e_ard-ga_ls5t_ard"` to the `cli_save_tasks` which then extracts scenes from all three products to be used in the tcw percentile

libs/stats/odc/stats/_tcw_percentiles.py
kieranricardo: I've added this in :) can you check that I've got the 1s and 0s right?

libs/stats/odc/stats/tasks.py
emmaai: all right, I'm gonna believe there is a grammar in naming product so `-` won't be in any product name.

emmaai (approved):
💯

PR #324, opened 2021-08-24
Refactored list_queues and removed get_queue_attributes

carioca-au:
As per @alexgleith suggestions, now:
- list_queues returns a generator

### DIFF
--- a/libs/aws/odc/aws/queue.py
+++ b/libs/aws/odc/aws/queue.py
@@ -77,7 +77,7 @@ def redrive_queue(
     if len(message_group) > 0:
         message_group = post_messages(alive_queue, message_group)
 
-    # Return the number of messages that were redriven.
+    # Return the number of messages that were re-driven.
     return count
 
 
@@ -90,37 +90,22 @@ def get_queue(queue_name: str):
     return queue
 
 
-def list_queues(region: Optional[str] = None):
+def get_queues(prefix: str = None, contains: str = None):
     """
-    Return a list of queues which the user is allowed to see
+    Return a list of sqs queues which the user is allowed to see and filtered by the parameters provided
     """
     sqs = boto3.resource("sqs")
-    return list(sqs.queues.all())
 
+    queues = sqs.queues.all()
+    if prefix is not None:
+        queues = queues.filter(QueueNamePrefix=prefix)
 
-def get_queue_attributes(queue_name: str, attribute: Optional[str] = None) -> dict:
-    """
-    Return informed queue's attribute or a list of queue's attributes when attribute isn't informed
-    Valid attribute options:
-        'ApproximateNumberOfMessages',
-        'ApproximateNumberOfMessagesDelayed',
-        'ApproximateNumberOfMessagesNotVisible',
-        'CreatedTimestamp',
-        'DelaySeconds',
-        'LastModifiedTimestamp',
-        'MaximumMessageSize',
-        'MessageRetentionPeriod',
-        'QueueArn',
-        'ReceiveMessageWaitTimeSeconds',
-        'VisibilityTimeout'
-    """
-
-    queue = get_queue(queue_name=queue_name)
-
-    if attribute is None:
-        return queue.attributes
-
-    return {attribute: queue.attributes.get(attribute)}
+    if contains is not None:
+        for queue in queues:
+            if contains in queue.attributes.get('QueueArn').split(':')[-1]:
+                yield queue
+    else:
+        yield from queues
 
 
 def publish_message(queue, message: str, message_attributes: Mapping[str, Any] = {}):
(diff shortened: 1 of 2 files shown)


libs/aws/tests/test_aws.py
```diff
@@ -111,3 +111,47 @@ def test_get_queue_attributes(aws_env):
...
+        'VisibilityTimeout'
+    ]
+
+    all_attributes = get_queue_attributes(queue_name="queue1")
+
+    assert len(all_attributes) == len(valid_attributes)
+
+    for att in valid_attributes:
+        att_returned = get_queue_attributes(queue_name="queue1", attribute=att)
+        assert att_returned.get(att)
```
alexgleith: Asserting that an object exists isn't really enough.

Maybe assert is not None? Or that the attribute is what you set it to?

carioca-au: Sure

carioca-au: **is not None** it is because the values are created by the library when the queue is created. So it might change in the future.

libs/aws/tests/test_aws.py
```diff
@@ -111,3 +111,47 @@ def test_get_queue_attributes(aws_env):
...
+def test_get_missing_queue_attribute(aws_env):
+    """
+    When the attribute isn't found in the queue boto 3 returns the attribute with value None
+    """
+    resource = boto3.resource("sqs")
+
+    resource.create_queue(QueueName="queue1")
+
+    # FifoThroughputLimit is an attribute specific for FIFO queues
+    att = 'FifoThroughputLimit'
```
alexgleith: Err, why not just use "FakeAttribute"
```suggestion
    att = 'FakeAttribute'
```

carioca-au: Because I wanted to show that you might use a valid attribute but because this isn't a FIFO queue, it won't have that and will generate an exception

libs/aws/tests/test_aws.py
```diff
@@ -111,3 +111,47 @@ def test_get_queue_attributes(aws_env):
...
+@mock_sqs
+def test_get_missing_queue_attribute(aws_env):
+    """
+    When the attribute isn't found in the queue boto 3 returns the attribute with value None
+    """
+    resource = boto3.resource("sqs")
+
+    resource.create_queue(QueueName="queue1")
+
+    # FifoThroughputLimit is an attribute specific for FIFO queues
```
alexgleith: This comment is out of date now

carioca-au: Fair enough, I commit from your suggestion and that didn't remove that.

carioca-au: Done

libs/aws/odc/aws/queue.py
```diff
@@ -87,37 +87,12 @@ def get_queue(queue_name: str):
     return queue
 
 
-def list_queues(region: Optional[str] = None):
+def list_queues():
```
alexgleith:
```suggestion
def get_queues():
```

Let's be consistent with the other one

alexgleith (changes requested):

libs/aws/odc/aws/queue.py
```diff
@@ -87,37 +87,12 @@ def get_queue(queue_name: str):
     return queue
 
 
-def list_queues(region: Optional[str] = None):
+def get_queues():
```
alexgleith: Can you please include a filter here that only returns queues with a name that contains a string.

Kirill888:
@carioca-au you might need to pull in changes from `develop` branch to get tests to work again, please user `--rebase` option followed by `push --force-with-lease`

carioca-au:
> @carioca-au you might need to pull in changes from `develop` branch to get tests to work again, please user `--rebase` option followed by `push --force-with-lease`

Thank you for the heads up. I did the rebase.

Kirill888:
@carioca-au this looks good to me, it's just it took 9 commits to get there, but it looks like this is one logical commit. Now that it's working, can you please collapse all your commits to just one (or however many logical commits you think this should have) and force push it here, then we can merge.

libs/aws/odc/aws/queue.py
```diff
@@ -90,37 +90,16 @@ def get_queue(queue_name: str):
     return queue
 
 
-def list_queues(region: Optional[str] = None):
+def get_queues(prefix: str = None):
```
alexgleith: I think this should include a 'name_contains' argument too, so we can filter queues by an arbitrary string and not just a prefix.

Kirill888 (approved):

alexgleith:
This is ready for review now, @Kirill888 please :-)

Kirill888 (approved):

alexgleith (approved):

Issue #318, opened 2021-08-20
Changes that require new libraries to test cause build failures


Kirill888:
Problem
======

You add new code that uses new library, you write tests, you update test docker to have new library, your tests pass locally. But when you make PR old test docker image is used to test your code and your tests fail on CI.

Proposed Solution
==============

Ideally one would always build docker image as described by the current branch to run the tests, but that can be slow and might also fail for random external reasons that have nothing to do with the current changes. I propose we implement a more complex workflow where we 

1. Check if feature branch has changes that are likely to require new docker image (files changed under `docker/**` or any `setup.cfg` files when diffing with `develop` branch)
   - YES: build fresh docker from this branch
   - NO: just pull in default test docker (saves time and reduces probability of dealing with transient issues)

And if running on `develop` branch (after PR is merged), similar rules can apply but instead of diffing branches you diff files changed since the last build, github provides enough information to figure this out programmatically.

So basically:
- Check if docker rebuild is needed based on changes being tested
   - `[YES]` Build fresh docker
   - `[NO ]` Pull "cached" docker from dockerhub

PR #326, opened 2021-08-24
Reworking test actions

Kirill888:
This is primarily to address issue #318, basically the change is to build test docker before every test so that if there are changes to test environment needed to support new work on a branch one can still run tests.

We are using github caching and `docker buildx` features to make docker rebuilding somewhat more tolerable. It still takes about a minute even when there are no changes to remake docker image. And then another extra minute to install all the different odc-tools libraries in develop mode. But it's more flexible.

I had to change test docker significantly to achieve reasonable performance for rebuilds. Main change was shrinking docker *context* to just `docker/` folder. So we no longer install `odc-tool` in develop mode during build, but instead do this on every test run (with caching to speed up local testing). This does add almost a minute before tests start on github actions, but this is really the only proper way for fresh checkouts, what we had before did not update package information which is stored next to code in develop mode install, not a problem for libraries but "apps" would not run without re-install.

Other major change is collating all different workflows into one with multiple jobs instead.


alexgleith (approved):
Looks ok to me.

README.md
```diff
@@ -67,10 +67,8 @@ Use `conda env update -f <file>` to install all needed dependencies for
 channels:
   - conda-forge
 dependencies:
-  - pip=20
-
   # Datacube
-  - datacube>=1.8.4
+  - datacube>=1.8.5
```
gadomski, line 71: suggestion (non-blocking): Protect against unexpected major/minor version bumps

Do we want to protect against an unexpected minor/major bump, e.g. do `datacube~=1.8.5` or `datacube~=1.8`?

libs/dtools/README.md
```diff
@@ -1,36 +1,11 @@
 odc.dtools
 ==========
 
-Tools for configuring rasterio dask workers for efficient COG access.
+Internal lib, might be moved/removed.
```
gadomski, line 4: suggestion (non-blocking): Provide if/when guidance

It might be useful to provide guidance on when this should be removed, e.g. "might be removed if ???". Just to provide some additional context in the even that this _doesn't_ get removed and someone finds this README in two years.

gadomski (approved):
A couple non-blocking comments, LGTM.

One top level (non-blocking) note: there's a lot of version bumps as a part of this PR, which seem like separate changes from a test infrastructure rebuild. That shouldn't block this PR but it does introduce additional variables to this change set unrelated to the original purpose.

README.md
Kirill888: > One top level (non-blocking) note: there's a lot of version bumps as a part of this PR, which seem like separate changes from a test infrastructure rebuild. That shouldn't block this PR but it does introduce additional variables to this change set unrelated to the original purpose.

This seems to be a common problem when tweaking test infra. We basically use "frozen" set of dependencies to prevent changes from "outside" causing failures. But we don't have any automated processes to bump those "frozen" sets. As a result when there is a major change this often causes a cascade of changes to make things work again, bumping different libraries, tweaking tests if those bumps cause problems etc.. 

> Do we want to protect against an unexpected minor/major bump, e.g. do datacube~=1.8.5 or datacube~=1.8?

With the new `pip` that now (finally!) attempts to find a consistent set of all packages having loose constraints slows it down a lot, so these version constraints are mostly to help `pip` to resolve things better. Remember this is purely internal test docker, we don't have those constraints on libraries. But for this specific case I will need 1.8.5 datacube for `odc-stac` as I need hashable `GeoBox` class.

libs/dtools/README.md
Kirill888: that's a leftover lib from early days, most of the functionality it provided has been moved to `datacube` proper now, doesn't look like it's being used by any other odc-tools lib.

```
pip show odc.dtools

Name: odc-dtools
Version: 0.2.0a0
Summary: Miscellaneous dask.distributed helper methods
Home-page: https://github.com/opendatacube/odc-tools/
Author: Open Data Cube
Author-email: 
License: Apache License 2.0
Location: /home/kirill/wk/odc-tools/libs/dtools
Requires: distributed
Required-by: 
```

we need to prune those things quite a bit, and possibly merge some of the separate packages into one.

README.md
gadomski: > Remember this is purely internal test docker, we don't have those constraints on libraries. But for this specific case I will need 1.8.5 datacube for odc-stac as I need hashable GeoBox class.

For sure...this inline comment was for the top-level README in the conda environment instructions, so presumably folks would use this line to set up their own environment for non-test use cases.

I think I flagged this because I saw another project get bit by just doing a "greater than or equal to" dependency recently: https://github.com/dask/s3fs/pull/510. Again, not a big deal, but wanted to flag `~=` as a probably-better pattern than `>=`.

Kirill888: Ah it's for conda...

@gadomski agree with your points about version tracking in general, but this is "closely related" lib with very slow release cycle. I was mostly wanting to avoid mentioning 1.8.4 release as it was broken on windows and had some other issues, and 1.8.3 is a year old at this point. Soon we'll have code in develop branch that genuinely needs features (fixes really) introduced in 1.8.5.

gadomski: Ah gotcha, no worries! Still learning the context here. Thanks.

PR #335, opened 2021-08-26
STAC to EO3 style Dataset

Kirill888:
Initial implementation of "STAC Item -> datacube.Dataset" code.

### Assumptions

- All Items across collection share the same structure
  - We pick first observed item from any given collection to construct datacube Product for ALL items in that collection
  - Same bands across all items
  - Same grouping of bands as far as resolution goes (eo3 grids)
- User provides information about bands: pixel types, units and nodata (but there is wildcard support)
- Proj extension must be present on every Item

### Current Limitations

- Random UUID only for now
  -  I'm planning to have an option for deterministic UUID generation from a set of keys that uniquely identify a dataset. Those keys will have to be supplied by the user or from "library of known collections".

### DIFF
--- a/libs/index/odc/stac/__init__.py
+++ b/libs/index/odc/stac/__init__.py
@@ -1,0 +1,16 @@
+"""
+STAC Item -> ODC Dataset[eo3]
+"""
+
+from ._eo3 import (
+    BandMetadata,
+    ConversionConfig,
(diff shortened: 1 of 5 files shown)


alexgleith (approved):
Looks good to me! Much cleaner than my dodgy stac to dc code.

PR #339, opened 2021-08-26
[odc-stac] Allow 3x3 proj:transforms in `asset_geobox`

gadomski:
As demonstrated in https://github.com/gadomski/chalkboard/blob/062dd84c7d7323066c0a2af238bc123e29aaf88f/odc-stac.ipynb, 3x3 matrices are not allowed in `odc.stac._eo3.asset_geobox`. This patch adds 3x3 support to `asset_geobox` while checking to ensure the matrix is an affine transform. Includes a unit test.

Sidecar changes:
- Update the stac_extensions list to be urls, which is required to use `ProjectionExtension` in `asset_geobox`.

### DIFF
--- a/libs/stac/odc/stac/_eo3.py
+++ b/libs/stac/odc/stac/_eo3.py
@@ -77,7 +77,14 @@ def asset_geobox(asset: pystac.asset.Asset) -> GeoBox:
     assert _proj.crs_string is not None
 
     h, w = _proj.shape
-    affine = Affine(*_proj.transform)
+    if len(_proj.transform) == 9:
+        if _proj.transform[6:] != [0, 0, 1]:
+            raise ValueError(
+                f"Asset transform is not affine: {_proj.transform}")
+        transform = _proj.transform[0:6]
+    else:
+        transform = _proj.transform
+    affine = Affine(*transform)
     return GeoBox(w, h, affine, _proj.crs_string)
 
 
--- a/libs/stac/tests/data/S2A_28QCH_20200714_0_L2A.json
+++ b/libs/stac/tests/data/S2A_28QCH_20200714_0_L2A.json
@@ -2,9 +2,9 @@
   "type": "Feature",
   "stac_version": "1.0.0-beta.1",
   "stac_extensions": [
-    "eo",
-    "view",
-    "proj"
+    "https://stac-extensions.github.io/eo/v1.0.0/schema.json",
+    "https://stac-extensions.github.io/view/v1.0.0/schema.json",
+    "https://stac-extensions.github.io/projection/v1.0.0/schema.json"
   ],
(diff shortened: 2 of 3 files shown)


Kirill888 (approved):
:+1:

libs/stac/odc/stac/_eo3.py
```diff
@@ -77,7 +77,14 @@ def asset_geobox(asset: pystac.asset.Asset) -> GeoBox:
     assert _proj.crs_string is not None
 
     h, w = _proj.shape
-    affine = Affine(*_proj.transform)
+    if len(_proj.transform) == 9:
+        if _proj.transform[6:] != [0, 0, 1]:
```
Kirill888, line 81: probably worth considering adding this check and "normalization" of transform data to the extension itself.

gadomski: The extension allows [both 6 and 9 element transforms](https://github.com/stac-extensions/projection#projtransform) so both are valid.

gadomski: That is to say, I think it'd be more on the `affine` package's side to accept both six and nine element matrices -- I found it strange it didn't accept 9.

Kirill888: I know, but the last 3 values are only allowed to be `0,0,1`, does extension check for that? One could add a property like, `.transform6` or something, that would only return first 6 values with verfication that if there are 9 values last 3 are `0,0,1`

Kirill888: Another option would be something like `.to_affine() -> affine.Affine`, but that would require dependency on `affine` or some fancy dynamic code that checks for presence of that lib.

Kirill888: > That is to say, I think it'd be more on the affine package's side to accept both six and nine element matrices -- I found it strange it didn't accept 9.

Affine matrix only has 6 degrees of freedom, so that kinda makes sense. 
It's Proj extension that allows 9 element representation of 6 element data, so it only makes sense for it to check that the extra 3 element "padding" is of expected value OR we can just say that first 6 elements are used, and if you choose to put more it won't have any effect no matter what values are there.

gadomski: Yeah, there could be space for additional logic here: https://github.com/stac-utils/pystac/blob/5ddfa839c708967227351f8c8372cb97f287fdc6/pystac/extensions/projection.py#L239-L256.

> Another option would be something like .to_affine() -> affine.Affine, but that would require dependency on affine or some fancy dynamic code that checks for presence of that lib.

This would probably be an anti-pattern for PySTAC, which tried to keep its dependency footprint very light.

PR #319, opened 2021-08-20
Add eodatasets3 odc-stats

supermarkion:
Hi, folks, 

As our 11-Aug-2021 review meeting discussed. In this eodatasets3 integration PR, 

Remove pystac section:
1. https://github.com/opendatacube/odc-tools/blob/develop/libs/stats/odc/stats/model.py#L360

Add eodatasets3 section:
1. create dataset.assembler in: https://github.com/opendatacube/odc-tools/blob/feature/integrate-eodataset3/libs/stats/odc/stats/model.py#L354
2. dataset.assembler to odc-metadata: https://github.com/opendatacube/odc-tools/blob/feature/integrate-eodataset3/libs/stats/odc/stats/io.py#L295
3. dataset.assembler to odc-stac: https://github.com/opendatacube/odc-tools/blob/feature/integrate-eodataset3/libs/stats/odc/stats/io.py#L291
4. dataset.assembler to accessories (e.g. proc-info.yaml, thumbnail.jpeg):
4.1  proc_info.ymal: https://github.com/opendatacube/odc-tools/blob/feature/integrate-eodataset3/libs/stats/odc/stats/io.py#L299
4.2  thumbnail.jpeg: https://github.com/opendatacube/odc-tools/blob/feature/integrate-eodataset3/libs/stats/odc/stats/io.py#L329
4.3  checksum file: https://github.com/opendatacube/odc-tools/blob/feature/integrate-eodataset3/libs/stats/odc/stats/io.py#L378

Cheers

Sai

PR #340, opened 2021-08-27
Add eodatasets3 as plugin with CLI flag to turn on

supermarkion:
Hi, @omad 

The previous PR: https://github.com/opendatacube/odc-tools/pull/319 is far away behind develop branch, so I resbumit this PR with latest develop branch again.

In this PR:
1. add eodatasets3 support. We can use eodatasets3 to generate metadata files with other support files.
2. add CLI flag --apply_eodatasets3 to turn on eodatasets3 plugin. The default is still PySATC, in case the integration issue.

Passed the test on WOfS summary and FC percentile integration. Cannot 100% sure the other run are safe with eodatasets3 plugin.

I also dump the Dask DAG as svg, and run batch tasks (1/10/30 tasks) in WOfS, cannot find the processing performance issue after enabling the eodatasets3 plugin.

cc @kieranricardo I will give your the fc percentile config yaml later.


docker/constraints.txt
```diff
@@ -145,6 +145,7 @@ pylint==2.10.2
 pyparsing==2.4.7
 pyproj==3.1.0
 pyrsistent==0.18.0
+eodatasets3==0.22.0
```
Kirill888: this file is not meant to be edited manually, also eodatasets3 is already included, see higher up in this file.

supermarkion: Forget the develop branch already had it. Thanks, Kirill.

kieranricardo (approved):
@supermarkion this looks good to me!

re. discussion about your code and its interaction with dask, it looks like your code is only called after the dask tasks have been run and the results stored in memory so this shouldn't cause any issues.

Kirill888:
> @supermarkion this looks good to me!
> 
> re. discussion about your code and its interaction with dask, it looks like your code is only called after the dask tasks have been run and the results stored in memory so this shouldn't cause any issues.

@kieranricardo how do you figure that? Also unless you use persist and then use result of that for further computation, there is no guarantee that Dask won't discard results you need to compute those thumbnails. Even if pixels are not recomputed they are still being marshalled back to the controlling code in this design, I would expect significant degradation in performance from that alone.

You really need to make thumbnail generation into a Dask operation rather than pushing all these pixels back into the controlling process memory.

kieranricardo:
@Kirill888 
 
`ds` is persisted before being passed to `sink.dump` (which then calls `dump_with_eodatasets3`) so I'm not 100% sure but I don't think the pixels are being re-computed here.

Good point re. the moving the data back into the controlling thread, but the summaries are ~100x smaller than the data used to compute them so this effect may not be noticeable. I think @supermarkion mentioned that he didn't observe any performance difference.

https://github.com/opendatacube/odc-tools/blob/78600e7d30ab01353cdd0fed707aba047408a63f/libs/stats/odc/stats/proc.py#L209-L218

supermarkion:
> @Kirill888
> 
> `ds` is persisted before being passed to `sink.dump` (which then calls `dump_with_eodatasets3`) so I'm not 100% sure but I don't think the pixels are being re-computed here.
> 
> Good point re. the moving the data back into the controlling thread, but the summaries are ~100x smaller than the data used to compute them so this effect may not be noticeable. I think @supermarkion mentioned that he didn't observe any performance difference.
> 
> https://github.com/opendatacube/odc-tools/blob/78600e7d30ab01353cdd0fed707aba047408a63f/libs/stats/odc/stats/proc.py#L209-L218

Yes. I am using the WOfS summary run as test.

- original odc-stats: 10 tasks: 
real	4m33.994s
user	12m12.049s
sys	0m20.738s

- eodatasets3 odc-stats: 10 tasks:
real	4m56.828s
user	12m52.973s
sys	0m19.323s

- original odc-stats: 30 tasks: 
real	30m23.951s
user	81m6.500s
sys	1m30.897s

- eodatasets3 odc-stats: 30 tasks:
real	31m41.682s
user	82m58.543s
sys	1m33.870s

libs/stats/odc/stats/io.py
```diff
@@ -242,5 +284,156 @@ def dump(self, task: Task, ds: Dataset, aux: Optional[Dataset] = None) -> Delaye
...
+                    # the bit operation return tuning_pixels is numapy.array
+                    if 'display_bands' not in single_band: # if no display_band, duplicate it as r,g,b
+                        tuning_pixels = [tuning_pixels, tuning_pixels, tuning_pixels]
+                    else:
+                        display_pixels = []
+                        for display_band in ["red", "green", "blue"]:
+                            display_pixels.append(tuning_pixels) if display_band in single_band["display_bands"] else display_pixels.append(zero_band)
+                        tuning_pixels = display_pixels # make sure the tuning_pixcels format is [numpy.array, numpy.array, numpy.array] == [r, g, b]
+
+                thumbnail_bytes = FileWrite().create_thumbnail_from_numpy(rgb=tuning_pixels,
```
Kirill888: I think preview management code should be taken out of this method, it's already way too long, makes it hard to reconstruct data flows.

Kirill888:
@kieranricardo @supermarkion my reading of the code is that original resolution pixel data is moved back from worker to processing code and it's done inside a function that takes a lazy dataset and produces lazy result, yet it also performs non-trivial operation of fetching data from the worker and computing thumbnails from that.

libs/stats/odc/stats/io.py
```diff
@@ -242,5 +284,156 @@ def dump(self, task: Task, ds: Dataset, aux: Optional[Dataset] = None) -> Delaye
...
+        if task.product.preview_image_singleband:
+            # single_band: {"measurement": "count_clear", "lookup_table": {"0":[150,150,110]}}, or
+            # single_band: {"measurement":'frequency', 'bit':1, 'display_bands': ['blue']} or 
+            # single_band: {"measurement":'frequency', 'bit':1, 'display_bands': ['red', 'green', 'blue']}
+            for single_band in task.product.preview_image_singleband:
+                band = single_band['measurement']
+                thumbnail_path = odc_file_path.split('.')[0] + f"_{band}_thumbnail.jpg"
+
+                tuning_pixels, stretch = FileWrite()._filter_singleband_data(data=ds[band].values.reshape([task.geobox.shape[0], 
+                                                                                                           task.geobox.shape[1]]),
```
Kirill888: how is this moving "thumbnail-sized only data", this is native resolution band data, is this done for every output band?

libs/stats/odc/stats/io.py
```diff
@@ -242,5 +284,156 @@ def dump(self, task: Task, ds: Dataset, aux: Optional[Dataset] = None) -> Delaye
         sha1_done = self._write_blob(sha1_digest, sha1_url, ContentType="text/plain")
 
         return self._write_blob(
-            json_data, json_url, ContentType=self._meta_contentype, with_deps=sha1_done,
+            json_data, json_url, ContentType=self._stac_meta_contentype, with_deps=sha1_done,
         )
+
+    def dump_with_eodatasets3(self, task: Task, ds: Dataset, aux: Optional[Dataset] = None, proc: StatsPluginInterface = None) -> Delayed:
+        """
```
Kirill888, line 376: This is meant to be Lazy->Lazy transform, yet it has non-lazy and expensive to compute components that can fail due to network issues.

Kirill888:
Wofs is not the best test to be honest, it has few bands and they are uint8. My main issue is that dump metadata mixes lazy and non-lazy computations.

It also introduces new constraint that was not there before, namely that `ds` is expected to be persisted before calling code to generate metadata making it unworkable for fully distributed Dask backend.

Previously one could just generate `Task -> Dask` on as many tasks as you want then submit that to cluster and wait for completion, with this change you can no longer do that, because of the expectation of Dataset being already computed before being passed on to metadata generation logic function, which is no longer lazy itself.

libs/stats/odc/stats/io.py
supermarkion: TBH, I am follow this method: https://github.com/opendatacube/odc-tools/blob/develop/libs/stats/odc/stats/io.py#L218 to generate this method. 

Of course, the original method is a lightweight computing component. May I know my version fails because of heavy workload?

Kirill888:
This also most likely breaks SQS renewal logic, as there is an assumption that generating Dask graph for metadata output is a quick operation (because lazy), but now it might take arbitrary amount of time, including complete recomputation of the entire Dask graph (worker might be killed, so even persisting computation is not a guarantee that it won't be repeated).

In general mixing lazy and non-lazy computation is a big no no, it will be a major pain to debug issues that this might produce "some of the time" only. 

There is nothing impossible here, you just need to figure out what thumbnail generation rules are for this product on the primary machine and then make thumbnail generation a Dask operation that runs on the worker, same as save cog.

libs/stats/odc/stats/io.py
supermarkion: You suggested we change the resolution (then reduce the data transmit cost) before pass to next methods? 

In my design, this 'tunning' should only happen to the 'thumbnail_display' bands. E.g. I plan to use WOfS summary frequency as the thumbnail_display band, so I just tuning it and display it it the JEPG 'red, green, blue' layer.

libs/stats/odc/stats/io.py
Kirill888: only non-lazy parts in that method are

- rendering internal structure to json

the rest is all Dask operations. Including COG generation and hash generation from result of cogs. In your case you read pixel data from Dask arrays via `.values`, this forces data transfer from the worker in the best case and a complete recomputation in the worst case.

Kirill888: You need to construct a lazy operation for thumbnail generation. Good way to think about lazy vs non-lazy is this:

Say you have  `Y = should_be_lazy(X)`, this should work even if `X.compute()` throws an error. In your case `Y` will not be returned if `X.compute()` throws error, because `should_be_lazy` is not lazy at all.

supermarkion: Do you suggests that I can follow this: https://github.com/opendatacube/odc-tools/blob/6e2037b3fb8c48b152d3ca613147ff60d661850e/libs/stats/odc/stats/io.py#L182 design. Just move the thumbnail_generation codes to a new method like:

> def  _ds_to_thumbnail_cog() -> List[Delayed]

to convert the thumbnail generation to a lazy method?

Kirill888: Yes, all "dump" operations should be lazy, except for cheap metadata operations: computing output paths and generating eo3 yamls or stac jsons. Any computation that depends on pixel values of the final output should be delayed properly, including thumbnail generation.

supermarkion: I am doing this change now.

libs/stats/odc/stats/io.py
```diff
@@ -209,6 +209,92 @@ def _ds_to_cog(self, ds: xr.Dataset, paths: Dict[str, str]) -> List[Delayed]:
...
+            2. single_band: {"measurement":'frequency', 'bit':1, 'display_bands': ['blue']} or 
+            3. single_band: {"measurement":'frequency', 'bit':1, 'display_bands': ['red', 'green', 'blue']}
+        """
+        band = single_band['measurement']
+        thumbnail_path = odc_file_path.split('.')[0] + f"_{band}_thumbnail.jpg"
+        zero_band = numpy.zeros((task.geobox.shape[0], task.geobox.shape[1]))
+
+        # if use lookup_table to tune pixel, it will return [numpy.Array, numpy.Array, numpy.Array]
+        # if use bit to tune pixel, it will return numpy.Array
+        tuning_pixels, stretch = FileWrite()._filter_singleband_data(data=ds[band].values.reshape([task.geobox.shape[0],
```
supermarkion, line 226: Hi, @Kirill888.
How are you. Sorry to interrupt you again.

I already moved the thumbnail generation code to standalone methods. However, based on my understanding, this new method still have 'un-lazy' section like this line

> tuning_pixels, stretch = FileWrite()._filter_singleband_data(data=ds[band].values.reshape([task.geobox.shape[0], task.geobox.shape[1]])

Can I just added dask.delay to wrap them up and that's enough?

Kirill888: > Can I just added dask.delay to wrap them up and that's enough?

probably not, because of xarray wrapping

see here for example that deals with both lazy and non-lazy inputs:

https://github.com/opendatacube/datacube-core/blob/develop/datacube/utils/cog.py

PR #346, opened 2021-08-31
Support aliases when generating Product from STAC

Kirill888:
Aliases can come from 2 sources
- EO extension (common_name)
- User defined alias map (overrides aliases discovered via EO extension)
- Detect situations when there multiple projections in use per item and raise error in that case
- Adds support for raster extension to extract `dtype, nodata, units` from raster extension when available

If common name is not unique it is ignored


Kirill888:
@gadomski @matthewhanson

I have pushed this branch to pypi and updated demo notebook here:

https://nbviewer.jupyter.org/gist/Kirill888/1c1c885e9bbedcc2871cb785b98cf048?uncache

Main changes are
- `dc_load` moved to `odc.stac.` (was `odc.algo.io.`)
- can refer to bands by common name defined EO extension
   - non-unique common names are ignored, but you can specify your own aliases

libs/stac/odc/stac/_eo3.py
```diff
@@ -83,19 +84,17 @@ def asset_geobox(asset: pystac.asset.Asset) -> GeoBox:
     :raises ValueError: when transform is not Affine.
     """
     _proj = ProjectionExtension.ext(asset)
-    assert _proj.shape is not None
-    assert _proj.transform is not None
-    assert _proj.crs_string is not None
+    if _proj.shape is None or _proj.transform is None or _proj.crs_string is None:
+        raise ValueError("Missing proj data")
```
gadomski:
```suggestion
        raise ValueError("The asset must have the following fields (from the projection extension): shape, transform, and one of an epsg, wkt2, or projjson")
```

Also, does this code path support a projection extension defined on the `Item` rather than the `Asset`?

gadomski (changes requested):

libs/stac/odc/stac/_eo3.py
Kirill888: My assumption was that item level proj data is exposed on an asset automatically by `pystac`, but I have not verified that, I just assumed that if this works for crs it works for others, but maybe not cause crs is special and always shared between assets?

alexgleith:
Asset level Proj information over-rides Item level. But Asset level is not required, so we should default to Item level.

Kirill888:
> Asset level Proj information over-rides Item level. But Asset level is not required, so we should default to Item level.

I agree we should, I was just assuming that this "default to item level metadata" was automatic within the extension that operates on an asset level. I will test my assumption now. and if it is wrong perform the default manually

Kirill888:
so, according to this code item level properties do propagate to asset when accessing them with `ProjectionExtension.ext`

```python
from pystac import Item
from pystac.extensions.projection import ProjectionExtension

assert ProjectionExtension.ext(
    Item.from_dict(
        {
            "type": "Feature",
            "stac_version": "1.0.0",
            "id": "test-item",
            "properties": {
                "datetime": "2012-12-12T00:00:00Z",
                "proj:shape": (100, 200),
                "proj:transform": [10, 0, 2000, 0, -10, 3000],
                "proj:epsg": 3857,
            },
            "geometry": None,
            "links": [],
            "assets": {
                "band": {"href": "http://example.com/some/file.tif", "roles": ["data"]}
            },
            "stac_extensions": [
                "https://stac-extensions.github.io/projection/v1.0.0/schema.json"
            ],
        }
    ).assets["band"]
).shape == (100, 200)
```

Kirill888:
@gadomski I have since added raster extension support, please re-review

gadomski (approved):

PR #350, opened 2021-09-03
Improve eodatasets3 performance

supermarkion:
In this PR, we are doing:
1. Remove the nodata values from Config file. The metadata will extract them from output dataset nodata field
2. Avoid the actually data loading from XArray.Dataset (e.g. dataset[band]. values). Then the eodatasets3 metadata files section can be lightweight.

### DIFF
--- a/libs/stats/odc/stats/_cli_run.py
+++ b/libs/stats/odc/stats/_cli_run.py
@@ -36,7 +36,7 @@ from ._sqs import SQSWorkToken
     help="Mark outputs for public access (default: no)",
 )
 @click.option(
-    "--apply_eodatasets3/--not_apply_eodatasets3",
+    "--apply_eodatasets3",
     is_flag=True,
     default=False,
     help="Apply eodatasets3 plugin to generate metadata files (default: use PySTAC to generate metadata files)",
--- a/libs/stats/odc/stats/model.py
+++ b/libs/stats/odc/stats/model.py
@@ -9,6 +9,7 @@ from pathlib import Path
 
 import json
 import pandas as pd
+import numpy
 import pystac
 import xarray as xr
 from datacube.model import Dataset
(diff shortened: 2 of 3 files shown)


libs/stats/odc/stats/_cli_run.py
```diff
@@ -36,7 +36,7 @@
     help="Mark outputs for public access (default: no)",
 )
 @click.option(
-    "--apply_eodatasets3/--not_apply_eodatasets3",
+    "--apply_eodatasets3",
```
omad, line 39:
```suggestion
    "--apply-eodatasets3",
```
Click will still use the same variable name, but it's prettier and more consistent with other application options.

libs/stats/odc/stats/model.py
```diff
@@ -400,11 +401,11 @@ def render_assembler_metadata(
                 # when we pass grid, the eodatasets will not load file from path
                 dataset_assembler.note_measurement(band,
                                                     path,
-                                                    pixels=output_dataset[band].values.reshape([self.geobox.shape[0], self.geobox.shape[1]]),
+                                                    pixels=numpy.zeros((self.geobox.shape[0], self.geobox.shape[1])),
```
omad, line 404: Why does `.note_measurement()` even require a `pixels` argument?

kieranricardo (approved):
LGTM :)

supermarkion:
Improve the eodatasets3 performance by changing data access pattern.

libs/stats/odc/stats/model.py
supermarkion: In this method, it needs the pixel values to check nodata: https://github.com/GeoscienceAustralia/eo-datasets/blob/3c38a5a740980389452884ba7a53fdee79f7cf1a/eodatasets3/images.py#L308

The calling orders are: 
1. https://github.com/GeoscienceAustralia/eo-datasets/blob/eodatasets3/eodatasets3/assemble.py#L736
2. https://github.com/GeoscienceAustralia/eo-datasets/blob/3c38a5a740980389452884ba7a53fdee79f7cf1a/eodatasets3/images.py#L278
3. https://github.com/GeoscienceAustralia/eo-datasets/blob/3c38a5a740980389452884ba7a53fdee79f7cf1a/eodatasets3/images.py#L297
4. https://github.com/GeoscienceAustralia/eo-datasets/blob/3c38a5a740980389452884ba7a53fdee79f7cf1a/eodatasets3/images.py#L308

PR #354, opened 2021-09-06
Stac.load

Kirill888:
Adding a more direct `odc.stac.load([Item], ...) -> xr.Dataset`

Also generating deterministic UUIDs now, this is actually needed for deterministic pixel loading. When multiple Items share the same timestamp and have overlapping pixels which Item's pixels end up in the output currently depends on UUIDs of those items. This is unfortunate, but it's the behaviour of `dc.load` (see https://github.com/opendatacube/datacube-core/issues/671), it used to be random which is even worse, now it's still arbitrary but at least it's deterministic.

Some other fixes include:
- renaming some STAC properties that EO3 expects to be named differently (because it was finalised before STAC 1.0)
- more tests
- warnings can be disabled via configuration
- introducing `*` config section that applies to all collections, useful for configuration of `warning` and `uuid` strategies
- Credentialization with `pc.sign` is supported by supplying `patch_url=pc.sign`, this same mechanism can be used for remapping urls from http to s3/azure or filesystem to thredds.

Closes #354

### DIFF
--- a/libs/stac/odc/stac/_load.py
+++ b/libs/stac/odc/stac/_load.py
@@ -1,0 +1,88 @@
+"""
+stac.load - dc.load from STAC Items
+"""
+from typing import Any, Callable, Dict, Iterable, Optional, Sequence, Union
+
+import pystac
+import xarray as xr
+from datacube.utils.geometry import GeoBox
+
+from ._dcload import dc_load
+from ._eo3 import ConversionConfig, stac2ds
+
+
+def load(
+    items: Iterable[pystac.Item],
+    bands: Optional[Union[str, Sequence[str]]] = None,
+    geobox: Optional[GeoBox] = None,
+    groupby: Optional[str] = None,
+    resampling: Optional[Union[str, Dict[str, str]]] = None,
+    skip_broken_datasets: bool = False,
+    chunks: Optional[Dict[str, int]] = None,
+    progress_cbk: Optional[Callable[[int, int], Any]] = None,
+    fuse_func=None,
+    stac_cfg: Optional[ConversionConfig] = None,
+    **kw,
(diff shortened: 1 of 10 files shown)


alexgleith (approved):
Looks good to me.

RichardScottOZ (changes requested):
Kirill,
Decide on plural arrays or not here 51-53?

:param chunks:
       Rather than loading pixel data directly construct
       a Dask backed arrays. ``chunks={'x': 2048, 'y': 2048}``

libs/stac/odc/stac/_load.py
```diff
@@ -0,0 +1,100 @@
...
+
+    :param resampling:
+       Controls resampling strategy, can be specified per band.
+
+    :param skip_broken_datasets:
+       Continue processing when IO errors are encountered
+
+    :param chunks:
+       Rather than loading pixel data directly construct
+       a Dask backed arrays. ``chunks={'x': 2048, 'y': 2048}``
```
Kirill888: - find lines that need changing
- select them
- press `+/-` button or `Ctrl-g`
- change them

you get this:
```suggestion
       Rather than loading pixel data directly construct
       Dask backed arrays. ``chunks={'x': 2048, 'y': 2048}``
```

it is then easy to incorporate the change directly from github.com

Kirill888: @RichardScottOZ like this?

RichardScottOZ: Yeah, wasn't somewhere that was easy to do at the time sorry, but yes.

construct... arrays.
or construct a Dask backed array.

Whichever you prefer/wanted in this case

gadomski (approved):

PR #356, opened 2021-09-07
Make stac transform safer

alexgleith:


Kirill888 (approved):

Kirill888:
it's a good practice to add tests that would fail before the change and succeed after when doing fixes like that, but since we gonna replace transformer soon anyway it's not worth it for now

alexgleith:
Thanks @Kirill888, you're right. I'm trying to write tests, but sometimes move too fast.

Appreciate the quick review :-)

PR #357, opened 2021-09-07
Update _load.py

RichardScottOZ:
arrays plural or singular here?


Kirill888:
thanks @RichardScottOZ , but can you please instead make a "suggestion" edit in this PR: https://github.com/opendatacube/odc-tools/pull/354 review, I will then commit it directly from github, I might still be rebasing this branch and it's easier that way.

RichardScottOZ:
No problem.

PR #359, opened 2021-09-08
Rework CI test runner

Kirill888:
- Use conda environment (cached and only gets rebuilt when environment definition changes)
- Use multiple Jobs with dependencies
- Only upload wheels after testing them
- Tests run on actual wheels rather than with editable installs
   - This found a packaging bug in `odc-apps-dc_tools` (fixed in this PR)
- This change is meant to make publishing to PyPI and documentation building  easier to implement.
- Paves the way for adding more checks like pylint/black/mypy
- Faster run time while still maintaining relatively stable test environment
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,12 +97,85 @@ jobs:
         mamba env export -p tests/env
 
 
-  run-tests:
+  test-with-coverage:
     runs-on: ubuntu-latest
 
     needs:
-      - build-wheels
       - build-test-env-base
+
+    steps:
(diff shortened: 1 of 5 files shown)


gadomski (approved):
LGTM. Would `tests/test-env-py38.yml` be usable as a top-level `enviornment.yml` for downstream users looking to get set up on their own systems via conda?

Kirill888:
@gadomski for sure, sections are commented, so if people don't need certain capabilities they are easy to prune, like remove all the sphinx dependencies if you are not planning to build documentation from source.

PR #360, opened 2021-09-08
Disable the inherit_properties warning msg

supermarkion:
Before we finish this feature: https://github.com/GeoscienceAustralia/eo-datasets/issues/194. The inherit_properties method will throw tons of warning message. E.g. 

> /env/lib/python3.8/site-packages/eodatasets3/assemble.py:681: UserWarning: Inheritable property 'odc:region_code' is different from current value ("Overriding property 'odc:region_code' (from '31NGE' to '31NHF')",)

Just disable it to save the log system.


tebadi (approved):

tebadi (approved):
Ideally we should fix the source instead of silencing them here but as I've understood this is a non-trivial change.

PR #361, opened 2021-09-08
Limit 0 not allowed for redrive to queue:

carioca-au:
- Limit 0 or lower, now raises an exception informing that value isn't valid.
- None is the default value, which will redrive the whole queue
- get_messages was changed to not allow 0 as limit and raise an exception in case of limit lower than 1.

### DIFF
--- a/libs/cloud/odc/aws/queue.py
+++ b/libs/cloud/odc/aws/queue.py
@@ -27,9 +27,6 @@ def redrive_queue(
             message.delete()
         return []
 
-    if limit < 1:
-        raise Exception(f"Limit {limit} is not valid.")
-
     dead_queue = get_queue(queue_name)
 
     if to_queue_name is not None:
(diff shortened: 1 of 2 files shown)


libs/cloud/odc/aws/queue.py
```diff
@@ -27,8 +27,10 @@ def post_messages(to_queue, messages):
             message.delete()
         return []
 
+    if limit < 1:
```
Kirill888: at this point `limit=None` is a possibility, so you need to check for that as you can't compare `None < 1`.
```suggestion
    if limit is not None and limit < 1:
```

carioca-au: You are totally right. Also, limit is just used on **get_messages** which already is checking for that, so this check is useless.
Removed.
Thank you

Kirill888 (approved):

PR #363, opened 2021-09-09
Allow fusing products with no file format

alexgleith:
* Fix up issue I introduced in removing FF but not adding a complete test.


Kirill888 (commented):
I know it's not part of the change but catching `KeyError` to detect missing optional values is not the best way of solving this problem.

I would use
- `toolz.get_in` to extract keys of interest with default value being `None`
- feed results into a `set([...])`
- raise error if set has more than one element

Kirill888 (approved):

PR #364, opened 2021-09-09
Starting `odc-stac` documentation

Kirill888:
~For now I'm using `STAC.load` == `stacload` name on readthedocs, see here:~

~https://stacload.readthedocs.io/en/latest/index.html~
https://odc-stac.readthedocs.io/en/latest/

It's kinda sparse, but it's a start.

There is no obvious way to publish multiple projects from the same repo to readthedocs, so we might have to move docs later on when we start documenting other libs properly.

Sample notebook in this PR can be reviewed here:

https://nbviewer.jupyter.org/gist/Kirill888/4845c9f33509650e4a7a6e751d377fb6

Configuration Changes
-------------------------------

Use STAC names to describe band information
- `measurements: ` -> `assets: `
- `dtype -> data_type` to match raster extension
- `units -> unit` to match raster extension


alexgleith:
Is it necessary to use RST to get these happening? I find Markdown a HUGE amount easier to edit and the barrier to contributions will be much lower for others too, I think.

Kirill888:
I think sphinx does support markdown for function annotations now, but I'm not sure how well or how to enable it. RST can be annoying, but it's not that bad, and it is way more precise than markdown. It is also very standard for Python documentation. For me it's easier to use RST than to deal with less supported markdown issues. VSCode has pretty good mode for it too, it's mostly just remembering to use 2 backticks instead of one. And weird url vs reference `_` positioning.

Standalone documents can be written in markdown without problem, code documentation not so much, also markdown would still have some non-standard linking mechanisms you'll need to remember, while RST is all about cross references.

libs/stac/docs/conf.py
```diff
@@ -0,0 +1,101 @@
...
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
+from odc.stac._version import __version__ as _odc_stac_version
+#print(sys.path)
+
+
+# -- Project information -----------------------------------------------------
+
+project = "STAC.load"
```
gadomski: What's the rationale behind this name, vs the PyPI name of `odc-stac`?

libs/stac/odc/stac/_load.py
```diff
@@ -28,17 +28,41 @@ def load(
...
+
+       At the moment one must specify desired projection and resolution of
+       the result. The plan is to make this choice automatic if not configured.
+
+
+    **Parameters**
 
     :param items:
-       STAC Items to load
+       Collection of STAC :class:`stac.Item` to load
```
gadomski:
```suggestion
       Iterable of STAC :class:`stac.Item` to load
```

`Collection` is a noun with specific meaning in the STAC ecosystem.

libs/stac/odc/stac/_load.py
```diff
@@ -59,23 +83,93 @@ def load(
...
+
+       The default behaviour is to use first observed valid pixel.
+       Item timestamp is used to determine order, ``nodata`` is
+       used to determine "valid".
 
     :param stac_cfg:
-       Controls STAC -> Dataset conversion, mostly used to specify
-       "missing" metadata like pixel data types. see ``stac_to_ds``.
+       Controls :class:`pystac.Item` ``->`` :class:`datacube.model.Dataset`
+       conversion, mostly used to specify"missing" metadata like pixel
```
gadomski:
```suggestion
       conversion, mostly used to specify "missing" metadata like pixel
```

gadomski (changes requested):

libs/stac/odc/stac/_load.py
Kirill888: Good catch, maybe `Sequence` is a more "human" name for it rather than `Iterable`? Although in Python `Sequence` is basically a kind of `Iterable` the one that supports `len()` and `[idx]`.

libs/stac/docs/conf.py
Kirill888: no good reason, it also lives on rtd under that name, but we should probably move it to `odc.stac` instead.

libs/stac/odc/stac/_load.py
gadomski: Since it's API documentation, IMO `Iterable` is ok to use because presumably only technical users will care about this function.

libs/stac/docs/conf.py
gadomski: :+1: for trying to keep the same name as much as possible (`odc-stac` for urls and such, `odc.stac` for python imports).

Kirill888:
```suggestion
project = "odc-stac"
```

libs/stac/docs/index.rst
```diff
@@ -0,0 +1,28 @@
+.. STAC.load documentation master file, created by
+   sphinx-quickstart on Wed Sep  8 12:43:50 2021.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+STAC.load
+=========
```
Kirill888:
```suggestion
odc-stac
========
```

Kirill888:
I have addressed comments.

I gonna merge it as I have another branch for publishing to pypi based off this one and I'm already pushing wrong commits to wrong branches.

PR #369, opened 2021-09-12
Fix bug with never loading as yaml and logic problems

alexgleith:
* Fix up the issue with only using `json.loads`
* Actually handle default globs, which were not included
* Fix another bug with actually testing the right variables for the `--stac` flag

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
@@ -65,7 +65,7 @@ def cli(input_directory, update_if_exists, allow_unsafe, stac, glob):
                 if "json" in glob:
                     metadata = json.load(f)
                 elif "yaml" in glob:
-                    metadata = yaml.load(f, Loader=Loader)
+                    metadata = yaml.safe_load(f, Loader=Loader)
                 # Do the STAC Transform if it's flagged
                 if stac:
                     metadata = stac_transform(metadata)
--- a/apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
@@ -14,6 +14,13 @@ from odc.stac.transform import stac_transform
 from typing import Generator, Optional
 import logging
 
+
+import yaml
+try:
+    from yaml import CLoader as Loader
+except ImportError:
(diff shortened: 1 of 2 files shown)


apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
```diff
@@ -45,16 +52,22 @@ def cli(input_directory, update_if_exists, allow_unsafe, stac, glob):
...
     for in_file in files_to_process:
         with in_file.open() as f:
             try:
-                # Yaml loads as JSON
-                metadata = json.load(f)
-                if stac_transform:
+                if "json" in glob:
+                    metadata = json.load(f)
+                elif "yaml" in glob:
+                    metadata = yaml.load(f, Loader=Loader)
```
omad:
```suggestion
                    metadata = yaml.safe_load(f, Loader=Loader)
```
It's not safe to load YAML from untrusted sources.

omad (changes requested):
Thanks Alex, just make that one change and looks good to go.

I'd seen the `stac` whoopsie too, but got waylaid with a whole mass of README updates.

alexgleith:
Thanks @omad! Is the `ImportException` the right way to handle switching on and off the C Loader?

omad:
> Thanks @omad! Is the `ImportException` the right way to handle switching on and off the C Loader?

I think so. It looks ugly, but that's what's shown in the PyYAML docs.

It might be part of why we use [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) in some places, because it's more likely to include C parsing code, which makes a *massive* performance difference.

omad (approved):

apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
```diff
@@ -45,16 +52,22 @@ def cli(input_directory, update_if_exists, allow_unsafe, stac, glob):
...
 
     added, failed = 0, 0
 
     for in_file in files_to_process:
         with in_file.open() as f:
             try:
-                # Yaml loads as JSON
-                metadata = json.load(f)
-                if stac_transform:
+                if "json" in glob:
```
Kirill888, line 65: this would be better as `in_file.endswith(".json")` which is a more direct way than to check glob pattern that might contain `json` as part of the path.

PR #371, opened 2021-09-13
remove negative pixels from input bands

NikitaGandhi:


NikitaGandhi:
@cbur24 and Fang please review this PR - implemented masking to remove negative pixels from input bands as Fag suggested. Let me know is this correct or not. 

Also please provide your input for SR calculation.

cbur24 (approved):

PR #372, opened 2021-09-14
Add publishing to pypi step

Kirill888:
Rules are

- Publish to S3 (`develop` branch only, with patched up dev version)
- Publish to pypi (`stable`or `pypi/publish` branches only, no version editing before publishing)
- Also fixes publishing of sources to S3, previously was only pushing wheels


gadomski (approved):
Seems reasonable to me. One alternative implementation is publishing when a new tag is pushed (with potentially some checking to make sure the tag hasn't been published). This makes publishing a bit more of an intentional step, rather than doing it on every merge to **develop**. But your approach seems sound as well.

Kirill888:
`on-release` event is how I would do it normally, but this repo has many packages. Pushing to pypi is still intentional, you need to edit `_version.py` in the library you want published, merge it to `develop`, then PR to `stable` only when that PR is merged will it be pushed to pypi, and only those packages that had a new version. Pushing to pypi is no-op on `develop` branch, only pushing to S3 is done from `develop` branch. Github actions don't let you filter jobs based on branches/files, you can only do that at the workflow level, so I have to disable individual steps instead manually.

The alternative is to move it into a separate workflow, but I do want tests to run on wheels I'm about to publish, so I either need to figure out cross-workflow data flows (harder than cross-jobs) or duplicate jobs between workflows.

PR #373, opened 2021-09-14
fix tcw percentile fmask

emmaai:
`fmask` is not defined by the `bit` but value, ref. product definition for details


libs/stats/odc/stats/_tcw_percentiles.py
```diff
@@ -44,8 +44,7 @@ def _native_tr(self, xx):
         """
         Loads data in its native projection.
         """
-
-        bad = (xx["fmask"] & 0b0000_1101) != 0 # a pixel is bad if any of the cloud, shadow, or no-data bits are 1
+        bad = (xx["fmask"] == 0) | (xx["fmask"] == 2) | (xx["fmask"] == 3) # a pixel is bad if any of the cloud, shadow, or no-data value
```
Kirill888: there is `enum_to_bool` that does exactly that but with fewer Dask nodes than what this will generate.

https://github.com/opendatacube/odc-tools/blob/47207229441fae4c7560d951e5cf5f2515b27cee/libs/algo/odc/algo/_masking.py#L462

emmaai: 👍

Kirill888 (approved):

PR #378, opened 2021-09-15
Enhancement/mosaic

kieranricardo:
This PR adds a few changes to CLI mosaic:

- allows the option to specify which bands are used
- allows more complex globs to be used for searching s3
- uses some built in dask functions to stack bands - unsure if this is equivalant to what was there before
- no longer creates a dataset cache file (only to be immediately loaded back into memory?)

For the extra flexibility I removed the clamping that was implemented for rgb stuff - I'm not sure what the best API is here. Maybe an `--rgb` flag?

### DIFF
--- a/libs/stats/odc/stats/_cli_generate_mosaic.py
+++ b/libs/stats/odc/stats/_cli_generate_mosaic.py
@@ -12,7 +12,7 @@ from datacube.model import Dataset
 from datacube import Datacube
 from odc.algo import store_to_mem, to_rgba
 from datacube.utils.cog import write_cog
-
+from odc.algo import save_cog
 from datacube.utils.dask import start_local_dask
 from datacube.utils.rio import configure_s3_access
 
@@ -67,7 +67,7 @@ def save(xx, location, product_name, verbose, creds=None, rgb_bands=None):
     if verbose:
         print(f"Writing {location}/{product_name}.tif")
 
-    write_cog(
+    save_cog(
         rgba,
         f"{location}/{product_name}.tif",
         blocksize=1024,
@@ -85,9 +85,9 @@ def save(xx, location, product_name, verbose, creds=None, rgb_bands=None):
 @click.argument("product", type=str)
 @click.argument("input_prefix", type=str)
 @click.argument("location", type=str)
-@click.option("rgb-bands", type=str)
+@click.option("--bands", type=str)
 @click.option("--verbose", "-v", is_flag=True, help="Be verbose")
-def cli(product, input_prefix, location, verbose, rgb_bands):
+def cli(product, input_prefix, location, verbose, bands):
     """
     Generate mosaic overviews of the stats data.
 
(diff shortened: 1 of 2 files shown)


kieranricardo:
Update, actually I think the only difference here is that clamping isn't happening. Maybe adding in a `argument` and clipping to `(0, clamp)` will do the trick.

libs/stats/odc/stats/_cli_generate_mosaic.py
```diff
@@ -77,48 +81,60 @@ def save(xx, location, product_name, verbose):
...
-    save(xx, location, product.name, verbose)
+    if bands is None:
+        suffix = ''
+    else:
+        suffix = '_' + '_'.join(bands)
+
+    if verbose:
+        print(f"Writing {location}/{product.name}{suffix}.tif")
+    xx = xx.squeeze('time').to_stacked_array('bands', ['x', 'y'])
+    yy = save_cog(
```
omad, line 83: Should this be calling `save()`, not `save_cog()`?

libs/stats/odc/stats/_cli_generate_mosaic.py
```diff
@@ -46,7 +50,7 @@ def xr_to_mem(xx, client):
     )
 
 
-def save(xx, location, product_name, verbose):
+def save(xx, location, product_name, verbose, creds=None, rgb_bands=None):
```
omad: Is this function still used?

omad (changes requested):
Kieran, can you please check which function is used, I assume it works as is, but it's messy and unclear whether `save()` or `save_cog()` is being used.

libs/stats/odc/stats/_cli_generate_mosaic.py
kieranricardo: yeah it's no longer used, but I thought I'd keep it in here for now just in case there's some special RGB stuff that's needed for geomedian. Do you know if anyone is using this?

kieranricardo: (if not I'll go ahead and delete it)

libs/stats/odc/stats/_cli_generate_mosaic.py
kieranricardo: nah, `save` is a wrapper for `write_cog` which loads everything into memory!

libs/stats/odc/stats/_cli_generate_mosaic.py
tebadi: I don't think anyone uses this but check with Nikita before deleting it.

kieranricardo: just checked with Nikita and it doesn't affect her stuff :)

omad: Sounds like this isn't used then, might as well remove it.

omad (approved):
LGTM, merge away Kieran.

PR #380, opened 2021-09-15
Float 16 frequency

kieranricardo:
Following a short discussion on Slack:

Kieran Ricardo 15 Sep at 11:36:
> on a side note I noticed that frequency uses float64 and  64 bits of precision might be overkill for representing a fraction in 0-1 with a denominator of <365. Maybe its worth storing this as float 16 ?

daveg
> make it so!

Paul Haesler
> Yeah even if you cared about the difference between once per solar day and once per sidereal day, that's excessive.


omad (approved):
Hey Kieran,

I think this is a good improvement. However can you please include more details on the background of the change, why was it float32 before, what has lead to the change, discussions, previous runs, some context to help out future us. :)

Cheers

PR #384, opened 2021-09-16
reenable surface reflectance calculation

NikitaGandhi:
- recalculate gm resulting input bands to surface reflectance
- recalculate gm resulting emad band between 0-10000


fangfy (commented):
Can we not call this sr_scale but output_scale?

NikitaGandhi:
> Can we not call this sr_scale but output_scale?

updated

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -35,6 +36,7 @@ def __init__(
             work_chunks: Tuple[int, int] = (400, 400),
             scale: float = 0.0000275,
             offset: float = -0.2,
+            output_scale: int = 10000, # scale USGS Landsat bands into surface reflectance
```
cbur24: Can we please change the comment here to instead read 'making SR range match sentinel-2 gm'

cbur24 (approved):

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -162,15 +164,17 @@ def reduce(self, xx: xr.Dataset) -> xr.Dataset:
...
-        #         gm[band] = gm[band].round().astype(np.uint16)
-        #
-        # # Rescale edev to sr_scale
-        # gm['emad'] = scale * sr_scale * gm['emad']
-        # gm['emad'] = gm['emad'].round().astype(np.uint16)
+        # rescale gm bands into surface reflectance scale
+        for band in gm.data_vars.keys():
+            if band in self.bands:
+                # gm[band] with low value (ie gm[band] < 7270) end up in negative values during sr scaling
+                # so resetting nodata and negative values back to 0 before rounding to uint16
```
NikitaGandhi: @fangfy @cbur24 see the comment above and let me know if we need to adjust this calculation

libs/stats/odc/stats/_gm_ls_bitmask.py
```diff
@@ -162,15 +164,17 @@ def reduce(self, xx: xr.Dataset) -> xr.Dataset:
...
+                gm[band] = self.scale * self.output_scale * gm[band] + self.offset * self.output_scale
+                gm[band] = gm[band].where(gm[band] > 0, 0)
+                gm[band] = gm[band].round().astype(np.uint16) # still getting some unexpected behaviour
```
cbur24: Are we still getting unexpected results? Or did we resolve all the no-data issues?

libs/stats/odc/stats/_gm_ls_bitmask.py
cbur24: Matches with what we discussed on Friday. Happy to see this tested.

cbur24 (approved):

libs/stats/odc/stats/_gm_ls_bitmask.py
NikitaGandhi: We resolved an issue by resetting  nodata and negative values back to 0. But we also experiencing slight variation on counts before and after results.

cbur24 (approved):
Nice work, @NikitaGandhi. I like the explicit setting of no-data with `keep_good_only`

PR #388, opened 2021-09-20
change dtype to float 32 for GDAL compatability

kieranricardo:
After a fair bit of work getting `safe_div` to work for `float16` I discovered that GDAL doesn't support `float16` anyway! :(


Kirill888:
if you are after smaller image file size I would suggest switching to scaled integer representation instead of `float16`, nobody needs that much precision for frequency band anyway, `[0, 10_000] -> [0, 1]` would be fine.

omad (approved):
`float32` is better than not working at all!

There's so many ways to decrease file size. Along with scaled ints, the [GDAL GeoTIFF](https://gdal.org/drivers/raster/gtiff.html) driver also supports:
- **NBITS=n**: From GDAL 2.2, n=16 is accepted for Float32 type to generate half-precision floating point values.
- **DISCARD_LSB=nbits or nbits_band1,nbits_band2,…nbits_bandN**: Set the number of least-significant bits to clear, possibly different per band. Lossy compression scheme to be best used with PREDICTOR=2 and LZW/DEFLATE/ZSTD compression.

PR #390, opened 2021-09-21
Integrate OWS to odc-stats and odc-tool

supermarkion:
We are using the [OWS](https://github.com/opendatacube/datacube-ows) to generate odc-stats thumbnail now. 

Because the feature branch always load the develop branch docker to run the test. So add the OWS library to develop branch firstly.

Note: I did the same changes in the feature branch, and run local docker test. It can generate the output without dependency issue.

### DIFF
--- a/docker/constraints.txt
+++ b/docker/constraints.txt
@@ -36,6 +36,7 @@ click-plugins==1.1.1
 cligj==0.7.2
 cloudpickle==1.6.0
 colorama==0.4.3
+colour==0.1.5
 commonmark==0.9.1
 coverage==5.5
 cryptography==3.4.7
@@ -44,6 +45,7 @@ dask==2021.8.1
 dask-image==0.6.0
 dataclasses==0.6
 datacube==1.8.5
+datacube_ows==1.8.19
 debugpy==1.4.1
 decorator==5.0.9
 deepdiff==5.5.0
@@ -53,8 +55,12 @@ docker==5.0.0
 docutils==0.15.2
 entrypoints==0.3
 eodatasets3==0.22.0
+flask==2.0.1
+flask-babel==2.0.0
+flask-log-request-id==0.10.1
 fsspec==2021.7.0
 future-fstrings==1.2.0
+geoalchemy2==0.9.4
 google-api-core==2.0.0
 google-auth==2.0.1
 google-cloud-core==2.0.0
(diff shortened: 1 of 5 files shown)


kieranricardo:
@supermarkion it looks like the tests are failing - is this expected?

supermarkion:
> @supermarkion it looks like the tests are failing - is this expected?

Hi @kieranricardo. Yes. When the Github run the test, its test docker still references to current develop branch docker. It does not have the OWS in its requirement.txt.

I am using the Makefile CMD to test locally. All good.

Kirill888:
@kieranricardo @supermarkion docker is no longer used for tests on github actions, instead you should add missing dependencies to https://github.com/opendatacube/odc-tools/blob/develop/tests/test-env-py38.yml

Kirill888:
also you should not make this dependency compulsory

supermarkion:
> also you should not make this dependency compulsory

Hi, @Kirill888. How are you. Do you mind explain which part I can improve? The dependency compulsory refers to OWS version or?

Kirill888:
What I mean is that one should be able to use `odc-stats` without ows being installed, and it should only be pulled in via a feature flag. How are you expecting to use it? Is it really an essential dependency. I only see it being imported and not used.

Kirill888:
but also version constraint is too tight

supermarkion:
> I only see it being imported and not used.

The implementation detail is under a feature branch now. I add import here aims to trigger the import happen, then test the environment crash or not.

> but also version constraint is too tight

The previous OWS version (before 1.8.19) will bring aiobotocore~=1.4.0 to the environment. This version cannot work with our odc-tool setting: aiobotocore[awscli,boto3]==1.3.3.

> What I mean is that one should be able to use odc-stats without ows being installed, and it should only be pulled in via a feature flag.

I am happy to bind the eodatasets3 and OWS library pull and import based on the _--apply-eodatasets3_ flag. Should not be a challenge.

Kirill888:
I think this PR was made with an incorrect assumption that Docker file on develop branch needs fixing for tests on a branch to work. While this was the case a while back it is not anymore. I think we should close this PR and changes made here should be incorporated with the rest of the changes when they are ready.

supermarkion:
> I think this PR was made with an incorrect assumption that Docker file on develop branch needs fixing for tests on a branch to work. While this was the case a while back it is not anymore. I think we should close this PR and changes made here should be incorporated with the rest of the changes when they are ready.

Make sense to me. Based on the latest design, we can fix the running environment in the feature branch. So see you in the next PR.

PR #392, opened 2021-09-21
Apply OWS styling to thumbnail

supermarkion:
The full version of https://github.com/opendatacube/odc-tools/pull/390.

In this PR:

1. use OWS styling dict (e.g. [color_ramp](https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/wms/ows_refactored/baseline_satellite_data/landsat_annual/style_ls_cfg.py#L6)) to change thumbnail data styling
2. a new "ows" optional dependency in setup.py/setup.cfg. Use can use `python -m pip install odc-stats["ows"]` to setup a environment can generate thumbnail JPEG.
3. update the odc-stats version to 0.2.0a5

Note: if user run odc-stats run with `--apply_eodatasets3` AND put the `preview_image_ows_style` in the [config yaml file](https://github.com/GeoscienceAustralia/dea-config/pull/808/files) , the system will try to import the OWS library and generate the thumbnail.

---

For test: 
`odc-stats run ga_ls8c_ard_3_2019--P1Y.db --config ga_ls8c_nbart_gm_cyear_3.yaml --threads=4 --memory-limit=14Gi --location file:///home/ubuntu/odc-stats-test-data/gm-sample 135:145 --overwrite --apply_eodatasets3`

The db file can download from: s3://dea-dev-stats-processing/gm-ls-temp3/dbs/ga_ls8c_ard_3_2019--P1Y.db
The yaml file is: can be found [here](https://github.com/GeoscienceAustralia/dea-config/pull/808/files#diff-7dca9ff183db1ed9f7461c9c260d34811d82d73d5fc2d25d6b9723c4c3b73d32)

I am using the previous CMD to run 10 tasks to test the performance. The average processing time each task is 22 minutes (23/09/2021) in my EC2 conda environment.

---
As the previous [PR](https://github.com/opendatacube/odc-tools/pull/390) discussion, there is no need to touch odc-tools/docker folder anymore.

Cheers

Sai


libs/stats/odc/stats/utils.py
```diff
@@ -292,3 +292,19 @@ def fuse_ds(
     fused_ds = Dataset(product, prep_eo3(fused_doc), uris=[""])
     fused_doc["properties"]["fused"] = "True"
     return fused_ds
+
+
+def install_ows_environment():
```
omad: I don't like this Sai.
- We want these as optional dependencies, and if we try to use a feature that absolutely depends on them we should fail with an error.
- I don't ever want tools to silently try to install extra libraries when I run them
- In many environments it would fail anyway.

Please remove this function and it's usage.

omad (changes requested):
Hi Sai,

I really like the idea of sharing image rendering code with OWS, but we need to handle that dependency better than this.

There should be an optional dependency in `setup.py/setup.cfg`, and reasonable error messages if the feature is tried to be used without them being available.

libs/stats/odc/stats/utils.py
supermarkion: As we (@omad and Sai) discussed, I will:

1. put the OWS relative dependency as optional dependencies in setup.cfg to manage.
2. make sure the import OWS will display useful message to let user know they should run pip install "odc-stats[ows]" to setup environment.

supermarkion: Already did the change at https://github.com/opendatacube/odc-tools/pull/392/commits/5593cf223a6aa4382fd3835e22cb3a5e3e89dc9c

libs/stats/odc/stats/io.py
```diff
@@ -247,13 +246,14 @@ def _ds_to_thumbnail_cog(self, ds: xr.Dataset, task: Task) -> List[Delayed]:
...
         if task.product.preview_image_ows_style:
-
-            
             try:
                 thumbnail_cog = self._get_thumbnail(ds, task, task.product.preview_image_ows_style, input_geobox, odc_file_path)
                 thumbnail_cogs.append(thumbnail_cog)
             except AttributeError:
                 _log.error(f"Cannot parse OWS styling: {task.product.preview_image_ows_style}.")
+            except ImportError:
+                _log.error('Please run python -m pip install "odc-stats[ows]" to setup environment.')
```
omad: I think that if someone has requested an OWS style preview image in their configuration, it should error out at this point instead of only printing a message. It's the type of thing that is easy to get caught out by when running things for real.

omad (changes requested):
A few more updates before we merge this please Sai.

1. Can you please update the PR description with some extra information:
   - It looks like this has added an extra option to the configuration file, please describe this.
   - Explain what testing you've done. I know it's not automated, and that's okay for now, but it's important to have a record of what has been done.
   - Explain how this relates, or doesn't, to the `eodatasets3` integration. What options are there now for generating preview images, and how are they triggered?
2. Update the code so that if there is a specific request for an OWS styled thumbnail, and OWS isn't available, we should exit with an error instead of logging and not creating one.

libs/stats/odc/stats/io.py
supermarkion: Change the code: Still throw the ImportError, but append more information: https://github.com/opendatacube/odc-tools/pull/392/commits/0d4b05e5375069f0dee1cd4a4f8080b86a79f3d2.

omad (approved):
Thanks for the updated commentary Sai, it really helps putting these changes in context.

omad:
Urgh, I fail at git. It's merged to `develop` now, but not via this PR.

PR #394, opened 2021-09-21
remove more sensitive sns info

GypsyBojangles:
remove more sensitive sns info


Kirill888 (approved):
tests pass, so good enough for me.

PR #399, opened 2021-09-24
Enhancements/stats docs

kieranricardo:
This PR adds some documentation for dataset fusing, caching data from multiple products, and generating mosaics.

It also makes mosaics API a little smoother by allowing more flexible for specify input datasets docs.

### DIFF
--- a/libs/stats/README.md
+++ b/libs/stats/README.md
@@ -59,10 +59,25 @@ GeoJSON files are useful when selecting test regions as well as for debugging sp
 
 <img src="docs/auxiliary/screenshot-L8-2015.png" alt="Screenshot GeoJSON" width="1000"/>
 
-Optionally two products can be fused using the following syntax:
+#### Dataset Fusing
+
+Some summary products require data from bands stored in multiple products. Products can be fused to use bands from both products in the derivitive products, this creates a virtual product that contains the bands from both products. Note that for datasets to be fused they must have the same `center_time` and `region_code`. This process find the matching dataset from each product that are in the same time and place and fuses them into one product.
+
+An example of this is `fc-percentiles`, which uses the fractional covers bands in `ga_ls_fc_3` to calculate the percentiles, and uses the `ga_ls_wo_3` band to mask out bad data.  
+
+Two products can be fused using the following syntax:
 
 ```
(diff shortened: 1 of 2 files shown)


libs/stats/README.md
```diff
@@ -277,3 +292,12 @@ You can now run a query on the log of the selected job:
 
 <img src="docs/auxiliary/screenshot-grafana.png" alt="Screenshot Grafana" width="1000"/>
 
+## Mosaics
+
+To quickly review products stats can be used to generate a low resolution continental mosaic of a summary product. To run the command:
+
+```
+odc-stats generate-mosaic product_definition.yaml s3://input-data-bucket/folder/*/*/metadata-file.stac-item.json s3://output-bucket/folder/subfolder --bands nbart_red,nbart_blue
```
omad: The example glob should have quote characters around it.

What goes into the `product_definition.yaml`?
I assume the output is a TIFF file for each band, but this should be stated.

Is the output actually a COG? How many overview layers are generated? Can the resampling method be specified? 

Does this/How does this parallelise across CPUs? What's the expected runtime and memory usage when running this?

omad (changes requested):
Nice improvements Kieran!

Can you please include some more information about the mosaic generation please.

libs/stats/README.md
kieranricardo: I'll add those quotes in :) 

The product definition is just a standard datacube product definition file. This enables the mosaicing to be done for stuff that isn't indexed. I'll make this a bit cleearer.

Yep its a proper COG! It has 7 overview layers (maybe fewer for smaller images?) And the resampling can't be specified right now but I can add this in.

It's fully parallelised, hopefully run time is `~1/num_cores`. Not exactly sure what the memory usage it's but probably something like `2048 * 2048 * bands * dtype_size * num_cores `, so for 64 cores maybe something like ~1GB ish. I think? Like each core gets a `2048 * 2048` chunk of the whole continent.

omad (approved):

omad:
Thanks Kieran. I've rebased and cleaned up the commits (generally appreciated) and added your comments about runtime resource requirements.

Issue #403, opened 2021-09-28
Task: add feature to support remote YAML file as odc-stats run config path


supermarkion:
Hello, @alexgleith, @omad and @Kirill888,

How are you.

I found the odc-stats run config [option](https://github.com/opendatacube/odc-tools/blob/develop/libs/stats/odc/stats/_cli_run.py#L49) only supports the dict content or local YAML file.

```
1) https://github.com/opendatacube/odc-tools/blob/develop/libs/stats/odc/stats/_cli_run.py#L49
2) https://github.com/opendatacube/odc-tools/blob/45542d24fcbfd8a9f3535ede57c435dff425a4bd/libs/stats/odc/stats/_cli_common.py#L89
3) https://github.com/opendatacube/odc-tools/blob/45542d24fcbfd8a9f3535ede57c435dff425a4bd/libs/io/odc/io/text.py#L38
4) https://github.com/opendatacube/odc-tools/blob/45542d24fcbfd8a9f3535ede57c435dff425a4bd/libs/io/odc/io/text.py#L42
```

I had these config saved as remote [YAML file](https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/master/dev/services/odc-stats/geomedian/ga_ls8c_nbart_gm_cyear_3.yaml) in a Github repo. And I plan to change the code [here](https://github.com/opendatacube/odc-tools/blob/45542d24fcbfd8a9f3535ede57c435dff425a4bd/libs/io/odc/io/text.py#L43) to make it support the feature `loading config from Github raw file`. 

Consider I plan to change the shared methods, then I submit the plan here to start the discussion.

The relative PR will be linked soon.

Cheers

Sai

Kirill888:
I don't think it's such a good idea, `odc.io` does not have a lot of dependencies right now, adding generic "slurp" file or any kind of url is not something I'd like to have in there. Besides this functionality is already available in `datacube` here

https://github.com/opendatacube/datacube-core/blob/d268f49d291cb42769f283b953ffb994c76ed4c5/datacube/utils/documents.py#L88

but we can't make `odc-io` depend on `datacube`. So I suggest you handle urls at the application level with datacube's `load_documents` function (write a simple wrapper around it as it produces a generator rather a single dict)

supermarkion:
Thanks @Kirill888 , I am happy to handle it at application level. 

How about your team @alexgleith on solution `handle it at application level` ?

alexgleith:
I don't mind.

[...]

PR #404, opened 2021-09-28
Use load_documents from odc to handle remote file

supermarkion:
Based on the talk here: https://github.com/opendatacube/odc-tools/issues/403.

The new/expected config YAML loading piepline:

1) if value meet filename pattern (based on urllib), we will use the similar remoate yaml load [approach](https://github.com/opendatacube/odc-tools/blob/develop/apps/dc_tools/odc/apps/dc_tools/add_update_products.py#L24) to load it. Consider we only have `one` file, so only one next() is enough.
2) if value is not filename, we assume it is the dict string, and still use old approach `parse_yaml_file_or_inline()` to process it.

I am using three filename patterns to test and all good:

1) https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/master/dev/services/odc-stats/geomedian/ga_ls8c_nbart_gm_cyear_3.yaml
2) ga_ls8c_nbart_gm_cyear_3.yaml
3) /home/ubuntu/odc-stats-test-data/ga_ls8c_nbart_gm_cyear_3.yaml

Cheers

Sai

### DIFF
--- a/libs/stats/odc/stats/_cli_common.py
+++ b/libs/stats/odc/stats/_cli_common.py
@@ -92,12 +92,25 @@ def click_yaml_cfg(*args, **kw):
     """
     def _parse(ctx, param, value):
         if value is not None:
-            from odc.io.text import parse_yaml_file_or_inline
+            import urllib
+            from urllib.parse import urlparse
+            r = urlparse(value)
 
-            try:
-                return parse_yaml_file_or_inline(value)
-            except Exception as e:
-                raise click.ClickException(str(e)) from None
+            # if referece to file, use documents.load_documents to process
+            if all([r.scheme, r.netloc]):
+                from datacube.utils import documents
+                try:
+                    return dict(next(documents.load_documents(value)))
+                except urllib.error.URLError as e:
+                    raise click.ClickException(str(e) + f". Cannot access file {value}") from None 
+                except Exception as e:
+                    raise click.ClickException(str(e)) from None 
+            else:
+                from odc.io.text import parse_yaml_file_or_inline
+                try:
+                    return parse_yaml_file_or_inline(value)
+                except Exception as e:
+                    raise click.ClickException(str(e)) from None
     return click.option(*args, callback=_parse, **kw)
 
 
--- a/libs/stats/odc/stats/_cli_common.py
+++ b/libs/stats/odc/stats/_cli_common.py
@@ -92,15 +92,16 @@ def click_yaml_cfg(*args, **kw):
     """
     def _parse(ctx, param, value):
         if value is not None:
-            import urllib
             from urllib.parse import urlparse
             r = urlparse(value)
 
-            # if referece to file, use documents.load_documents to process
             if all([r.scheme, r.netloc]):
-                from datacube.utils import documents
+                import urllib
+                import fsspec
+                import yaml
                 try:
-                    return dict(next(documents.load_documents(value)))
+                    with fsspec.open(value, mode="r") as f:
+                        return next(yaml.safe_load_all(f))
                 except urllib.error.URLError as e:
                     raise click.ClickException(str(e) + f". Cannot access file {value}") from None 
                 except Exception as e:


alexgleith:
I'm not convinced this is a good idea, and I don't like how this is implemented!

This is a simple example of loading a YAML file over the network: https://github.com/opendatacube/odc-tools/blob/develop/apps/dc_tools/odc/apps/dc_tools/add_update_products.py#L24

supermarkion:
> I'm not convinced this is a good idea, and I don't like how this is implemented!
> 
> This is a simple example of loading a YAML file over the network: https://github.com/opendatacube/odc-tools/blob/develop/apps/dc_tools/odc/apps/dc_tools/add_update_products.py#L24

Sure, use/apply the recommended method to change it.

libs/stats/odc/stats/_cli_common.py
```diff
@@ -92,12 +92,25 @@ def click_yaml_cfg(*args, **kw):
...
         if value is not None:
-            from odc.io.text import parse_yaml_file_or_inline
-
-            try:
-                return parse_yaml_file_or_inline(value)
-            except Exception as e:
-                raise click.ClickException(str(e)) from None
+            import urllib
+            from urllib.parse import urlparse
+            r = urlparse(value)
```
omad, line 96: If you're going to use `.load_documents()`, it handles checking whether it's a local file or a remote file, so this shouldn't be necessary.

libs/stats/odc/stats/_cli_common.py
```diff
@@ -92,12 +92,25 @@ def click_yaml_cfg(*args, **kw):
...
+                try:
+                    return dict(next(documents.load_documents(value)))
+                except urllib.error.URLError as e:
+                    raise click.ClickException(str(e) + f". Cannot access file {value}") from None 
+                except Exception as e:
+                    raise click.ClickException(str(e)) from None 
+            else:
+                from odc.io.text import parse_yaml_file_or_inline
+                try:
+                    return parse_yaml_file_or_inline(value)
```
omad, line 112: I think this whole section can be removed. Also, there's no need to use `parse_yaml_file_or_inline()` I don't think, since I don't think it makes any sense to try to read a literal yaml file from a CLI argument.

omad (changes requested):
If we're deferring to `.load_documents()` this can be a whole lot simpler.

libs/stats/odc/stats/_cli_common.py
supermarkion: I am OK to remove the support `read a literal yaml file from a CLI argument`. 

But it's a shared method, and should let stakeholders know it: @alexgleith @Kirill888

Kirill888: I found it useful for embedding config in k8s, but also I don't care.

alexgleith: We embed config in k8s so please don't change the current behaviour.

alexgleith:
I actually think that for this one-off process, storing the config remotely where it might be changed is not great.

I already feel that Alchemist is a bit fragile for having config external to the deployment of the thing, and it would be better if it was embedded in the k8s deployment yaml.

For Stats, we are now using a k8s configuration, like this: https://bitbucket.org/geoscienceaustralia/datakube-apps/src/master/workspaces/deafrica-prod-af/processing/argo/once/stats_wofs_ls_summary_annual.yaml

So any changes to this section of config loading _must_ be backwards compatible, please @supermarkion and @omad

supermarkion:
> I actually think that for this one-off process, storing the config remotely where it might be changed is not great.
> 
> I already feel that Alchemist is a bit fragile for having config external to the deployment of the thing, and it would be better if it was embedded in the k8s deployment yaml.
> 
> For Stats, we are now using a k8s configuration, like this: https://bitbucket.org/geoscienceaustralia/datakube-apps/src/master/workspaces/deafrica-prod-af/processing/argo/once/stats_wofs_ls_summary_annual.yaml
> 
> So any changes to this section of config loading _must_ be backwards compatible, please @supermarkion and @omad

I am OK to still support the existing feature `read a literal yaml file from a CLI argument` (that is why I keep it in my current codes). Are you OK with this @omad ?

supermarkion:
Close this change. No plan to crash all existing pipelines and lost backwards compatible.

emmaai:
I'm reviving this PR, reasoning:
- don't think copy/paste one yaml file into another is less fragile than reading one yaml file remotely
- need some consistency in local run  and k8s run, aka, NOT being required to remember copy/paste the config file to another and do 2 PRs
- need a bit simplified and flexible config on argo yaml, it has a no-ending expanding parameters list, anytime anyone wanna add or enable some option on `odc-stats`, those never changing product-wise options can stay in the product yaml
- agree with the compatibility requirements of not changing current behavior by supporting reading CLI yaml

alexgleith:
Please re-create the PR over here, @emmaai https://github.com/opendatacube/odc-stats

emmaai:
> Please re-create the PR over here, @emmaai

yes, I will, only for the sake of keeping history and cherry-picking the commit.

emmaai:
new PR https://github.com/opendatacube/odc-stats/pull/14

PR #405, opened 2021-09-28
Add compatibility for parsing ARD c3 metadata

james-blitzm:
Resolves STAC parsing for ARD c3 metadata.

### DIFF
--- a/libs/stac/odc/stac/transform.py
+++ b/libs/stac/odc/stac/transform.py
@@ -88,7 +88,7 @@ def _stac_product_lookup(
     if constellation in KNOWN_CONSTELLATIONS:
         # This handles the Element 84 and  Microsft PC docs
         if constellation == "sentinel-2":
-            dataset_id = properties.get("sentinel:product_id") or properties.get("s2:granule_id")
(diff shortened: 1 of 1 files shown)


alexgleith:
This won't work for what you're doing, @james-blitzm.

Because the lookup for `sentinel-2` is naming the product too...

I think that this code path should not be running if there's an `odc:product` set.

alexgleith:
> Resolves STAC parsing for ARD c3 metadata.

This isn't what this PR does. You're trying to get a Sentinel-2 derivative product able to be parsed.

uchchwhash:
I don't see what the way forward here is besides this @alexgleith, we are trying to enable the burn area product to be indexed, and it derives from sentinel-2 ARD and therefore is inherits that tag there. We confirmed that our provisional sentinel-2 ARD product (unfortunately) has the tag there. So I think @james-blitzm is correct in that this enables parsing the ARD product too, as well as the burn area product.

uchchwhash:
Sorry @alexgleith I see what you mean. Note to self: check before you post.

uchchwhash:
Superseded by https://github.com/opendatacube/odc-tools/pull/406.

PR #406, opened 2021-09-28
Only search for known const if no product yet

alexgleith:
Only run the KNOWN_CONSTELLATION path if there's no product_name found


uchchwhash (approved):
💯

james-blitzm:
Yep, looks good

james-blitzm (approved):

PR #423, opened 2021-10-06
support save-tasks for zero/no dataset counts

NikitaGandhi:
Current implementation throw error when there is no/zero datasets on cache db creation. This PR fix this issue.

```
TypeError: cannot unpack non-iterable bool object
```


alexgleith (approved):
Looks good to me

PR #425, opened 2021-10-08
Apply a fix to ensure input source datasets are unique per task.

tebadi:


Kirill888 (approved):
this is fine, but a proper fix should be in the query functions, see #426

supermarkion (approved):

PR #428, opened 2021-10-11
Prepare for odc-stac extraction from this repo

### LINKED ISSUE #426: Duplicate entries might be returned from streaming query functions
`ordered_dss` and `chopped_dss` functions emulate streaming query by chopping query along time dimension, making several smaller queries one after another then stitching results and returning as continuous stream of datasets. 

https://github.com/opendatacube/odc-tools/blob/9e126245a0180064ccb933e0731f72a00dc32ad3/libs/stac/odc/index/_index.py#L242-L245

https://github.com/opendatacube/odc-tools/blob/9e126245a0180064ccb933e0731f72a00dc32ad3/libs/stac/odc/index/_index.py#L257-L259

however some datasets valid time period might span multiple query periods, this currently results in this Dataset being returned several times. 

To resolve that, one needs to keep a set of UUIDs returned from a previous query around so that duplicates can be detected and filtered out from the output. There is no need to keep a full set of dataset uuids, the set returned from the last query is sufficient as dataset time range is a single interval, so duplicates can only appear in consecutive queries only (but more than twice is possible, example dataset with a time range of one month will be returned in 4+ consecutive 1-week queries).

cc: @tebadi

Kirill888:
Mostly various cleanups

- Remove unused code
- Remove dependency on `odc-io`
- Update test docker
- Fixes for duplicates returned from chopped queries, Closes #426 
- Basic case for automatic CRS/resolution is working now, there might be some corner cases though, like when CRS is supplied to be geographic but resolution is automatically found to be projected


.github/workflows/dev-docker.yml
```diff
@@ -0,0 +1,92 @@
+name: Update Dev Docker
+
+on:
+  pull_request:
+    paths:
+      - 'docker/**'
+  push:
+    paths:
+      - 'docker/**'
+
```
gadomski, line 10: Is the `docker/**` too restrictive? E.g. if there's a code change (but not a docker change) we would still want to update the dev images, correct?

libs/stac/odc/stac/_eo3.py
```diff
@@ -104,7 +104,7 @@ def band_metadata(asset: pystac.asset.Asset, default: BandMetadata) -> BandMetad
     )
 
 
-def has_proj_ext(item: pystac.Item) -> bool:
+def has_proj_ext(item: Union[pystac.Item, pystac.Collection]) -> bool:
     """
     Check if STAC Item has projection extension
```
gadomski:
```suggestion
    Check if STAC Item or Collection has projection extension
```

gadomski (changes requested):
One little comment/question and one doc edit, otherwise looks good!

.github/workflows/dev-docker.yml
Kirill888: @gadomski this docker image does not include code, it is designed to test code that is brought over in a volume, image only contains dependencies

PR #433, opened 2021-10-18
Access the version from local module

supermarkion:
Found the odc.stats.__version__ values is different when same codebase install by `pip install odc-stats` or `pip install odc-stats --extra-index-url https://packages.dea.ga.gov.au/`.

Also found this is an expected result from [commit](https://github.com/opendatacube/odc-tools/commit/1b7bd03065c05577864b1eb7725d8b6927b600e0).

Before we move to standalone odc-stats repo, access the _version file directly to make sure dump the expected value.


Kirill888:
`_version.py` is overwritten during packaging when pushing to S3, this change won't make any difference

supermarkion:
> `_version.py` is overwritten during packaging when pushing to S3, this change won't make any difference

Thanks a lot for your information. I will use the public PyPI to setup odc-stats in Docker image.

Issue #432, opened 2021-10-18
Proposal: ODC-tools split/clean-up


SpacemanPaul:
odc-tools was conceived as a place for "ODC features that DEA is experimenting with or prototyping with the intention of being integrated into odc-core in the future."

However some parts of odc-tools have already grown into mission-critical tools, and the tangled experimental architectural anti-pattern of the tools repository makes it difficult for them to mature further.  Splitting odc-stac into a separate repository was an important step in the right direction.

Most pressingly for DEA, odc-stats (a.k.a. "statistician") is now a key part of DEA infrastructure and is being recommended to external users as the go-to tool for large scale batch processing in the cloud - it therefore strongly desirable to get odc-stats out of odc-tools and into a separate, cleaner repository.

### Guiding Principles

1. Get odc-stats out of odc-tools.
2. Ensure new odc-stats package has no dependencies on packages inside odc-tools repository.
3. Keep package inter-dependencies as clear and simple as possible.

#### 1.  odc-tools packages to be split into separate repos:

1. odc_cloud
2. odc_dscache
3. odc_algo
4. odc_stats

These packages appear to be fairly tightly focused and uncluttered: particularly odc_cloud and odc_dscache.

These libraries may include some experimental or specialised functions that are not used by odc-stats or other libraries - where practical, these should be moved to other new or existing experimental packages within odc-tools.

#### 2. Move all "algorithmic" code out of odc-stats into odc-algo.

Code specific to particular algorithms (e.g. geomad, fractional cover, wofs, etc.) are currently split between odc-algo and odc-stats.)  Ideally this should be consolidated into odc-algo, leaving odc-stats as containing only the core statistician library, with modules imported from odc-algo.

#### 3. Remove apps/libs split.

Most of the apps in odc-cloud-apps could be more usefully packaged with odc-cloud.

Similarly most of the apps in odc-apps-dc-tools could be more usefully packaged with either odc-stats, odc-cloud, or odc-stac.

Again, some experimental or highly specialised apps in these packages are probably best left inside odc-tools.

#### 4. Move statistician docker build into odc-stats

Currently releasing a new version of Statistician requires coordinated changes to the odc-stats and datacube-docker repos.  Docker builds of packages should live with the package, not in a separate docker repo.

#### 5. Libraries to stay in odc-tools

There seems no pressing reason to move odc-ui out of odc-tools.

[...]

PR #436, opened 2021-10-19
Remove io dependencies

SpacemanPaul:
Remove (direct) dependency of odc-stats on odc-io.

Note stats still depends on odc-io indirectly, via odc-dscache.

Preparatory refactor for #432


SpacemanPaul:
Actually I need to copy across some test coverage from odc-io as well.

supermarkion (approved):
All good to me.

PR #437, opened 2021-10-19
Fix tcwp nodata reprojection and no datasets in searching product

emmaai:
- Fix tcw percentile nodata issue in reprojection
- Fix the issue when no dataset found when searching the product


SpacemanPaul (approved):
Makes sense to me.

PR #440, opened 2021-10-20
Accept func type in proc-info.yaml

supermarkion:
When we are using the WO summary styling from OWS [cfg](https://github.com/GeoscienceAustralia/dea-config/blob/master/prod/services/wms/ows_refactored/inland_water/wofs/style_wofs_cfg.py#L91). It has one section `index_function`. It will add one more field called `band_mapper`. The `band_mapper` type is a Python function. Because the [proc-info YAML setting](https://github.com/opendatacube/eo-datasets/blob/develop/eodatasets3/serialise.py#L65) does not accept function type. Then it will crash when dump the`proc-info.yaml`, and stop processing.

* Add the func type setting to YMAL then can generate the proc-info.yaml
* Change odc-stats version number (if don't do it, CD/CI system will skip `publish PyPI step`)

Cause this `index_function` is a special use case in WO summary. E.g. [GeoMAD OWS cfg ](https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/odc-stats/geomedian/ga_ls8c_nbart_gm_cyear_3.yaml#L50) does not have the `index_function`. Then, the GeoMAD or FCP output will no change. 

Also download the latest odc-stats Docker image:.0.4.1 (https://github.com/opendatacube/datacube-docker/blob/main/statistician/version.txt), and re-install odc-stats from current feature branch. Can pass the GeoMAD, WO summary and FCP run test.

Here is the example GeoMAD output proc-info.yaml:
```yaml
---
input-products:
- ga_ls8c_ard_3
odc-stats-config:
  name: ga_ls8c_nbart_gm_cyear_3
  version: 3.0.0
  short_name: ga_ls8c_nbart_gm_cyear_3
  location: file:///code/odc-stats-test-data/gm-sample
  properties:
    odc:file_format: GeoTIFF
    odc:product_family: geomedian
    odc:producer: ga.gov.au
  measurements:
  - red
  - green
  - blue
  - nir
  - swir1
  - swir2
  - sdev
  - edev
  - bcdev
  - count
  href: https://collections.dea.ga.gov.au/product/ga_ls8c_nbart_gm_cyear_3
  region_code_format: x{x:02d}y{y:02d}
  cfg:
  naming_conventions_values: dea_c3
  explorer_path: https://explorer.dea.ga.gov.au/
  inherit_skip_properties:
  - eo:cloud_cover
  - fmask:clear
  - fmask:snow
  - fmask:cloud
  - fmask:water
  - fmask:cloud_shadow
  - eo:sun_elevation
  - eo:sun_azimuth
  - gqa:iterative_stddev_x
  - gqa:iterative_stddev_y
  - gqa:iterative_stddev_xy
  - gqa:stddev_xy
  - gqa:stddev_x
  - gqa:stddev_y
  - gqa:mean_xy
  - gqa:mean_x
  - gqa:mean_y
  - gqa:abs_xy
  - gqa:abs_x
  - gqa:abs_y
  - gqa:abs_iterative_mean_y
  - gqa:abs_iterative_mean_x
  - gqa:abs_iterative_mean_xy
  - gqa:iterative_mean_xy
  - gqa:iterative_mean_x
  - gqa:iterative_mean_y
  - gqa:cep90
  - landsat:landsat_product_id
  - landsat:landsat_scene_id
  - landsat:collection_category
  - landsat:collection_number
  - landsat:wrs_path
  - landsat:wrs_row
  preview_image_ows_style:
    name: simple_rgb
    title: Simple RGB
    abstract: Simple true-colour image, using the red, green and blue bands
    components:
      red:
        red: 1
      green:
        green: 1
      blue:
        blue: 1
    scale_range:
    - 0
    - 3000
  classifier: ard
  maturity: final
  collection_number: 3
  nodata:
software_versions:
- name: eodatasets3
  url: https://github.com/GeoscienceAustralia/eo-datasets
  version: 0.22.2
- name: odc-stats
  url: https://github.com/opendatacube/odc-tools
  version: 1.0.2
- name: gm_ls_annual
  url: https://github.com/opendatacube/odc-tools
  version: 3.0.0
- name: datacube-ows
  url: https://github.com/opendatacube/datacube-ows
  version: 1.8.20
...

```
When compare with the old [output file](https://dea-public-data-dev.s3-ap-southeast-2.amazonaws.com/ga_ls8c_nbart_gm_cyear_3/3-0-0/x06/y25/2019--P1Y/ga_ls8c_nbart_gm_cyear_3_x06y25_2019--P1Y_final.proc-info.yaml). The only two differences are: 1) `location` (cause new one is local run), and 2) `odc-stats library version`.


libs/stats/odc/stats/io.py
```diff
@@ -380,8 +382,15 @@ def dump_with_eodatasets3(self, task: Task, ds: Dataset, aux: Optional[Dataset]
         serialise.to_stream(odc_meta_stream, meta)
         odc_meta = odc_meta_stream.getvalue() # odc_meta is Python str
 
+        # use ruamel.yaml setup from eodatasets3
+        yaml = serialise._init_yaml()
+        # accept the func type in proc-info.yaml
+        def _represent_function(self, data: types.FunctionType):
+            return Representer.represent_str(self, data.__name__)
+        yaml.representer.add_representer(types.FunctionType, _represent_function)
+
```
tebadi, line 391: I don't know what this"band_mapper" is.  I'm okay with the change but not sure whether this was needed at all. @omad?

supermarkion: I believe @SpacemanPaul could explain the detail/story about the `band_mapper`.

SpacemanPaul: I'm a bit confused what the problem is.  I can't see "index_function" or "band_mapper" anywhere in this PR apart from the comments!

But index_function doesn't have to be an actual function - and indeed using an actual function is only supported by the stand-alone API.  You can always specify it as a string if you need the config to be serialisable, e.g.:

```
"index_function": {
        "function": "datacube_ows.band_utils.single_band",
        "mapped_bands": True,
        "kwargs": {
            "band": "frequency",
        },
    }
```
band_mapper is an internal implementation detail - you shouldn't need to think about it at the level you are calling things from.
The function is simply referenced by name in the config.

supermarkion: Thanks for the pair coding with me. When we are using the OWS config e.g.:
```
"index_function": {
        "function": "datacube_ows.band_utils.single_band",
        "mapped_bands": True,
        "kwargs": {
            "band": "frequency",
        },
    }
```
Everything in this cfg can be serized to string, and dump to YAML file.

However, it has another field called `band_mapper` come from the [internal processing](https://github.com/opendatacube/datacube-ows/blob/master/datacube_ows/ogc_utils.py#L349). Even we did not clarify it in the config file. This field is a Python lambda function, and cannot dump to YAML file before.

Consider of the `proc-info.yaml` is only a traceability file. Then modify the YAML setting to accept func type. Then it can also has `band_mapper` info here. 

Of course, we can also implement tons of if-else statements to remove the `band_mapper` before cfg info dump to proc-info.yaml. But I prefer the current solution.

SpacemanPaul: I still don't understand how the OWS config is being written to YAML.  There should be a way to export the config in a serialisable form so the band_mapper doesn't even show up.

supermarkion: In the current odc-stats processing pipeline:

1. pass the OWS cfg from CLI: https://github.com/opendatacube/odc-tools/blob/develop/libs/stats/odc/stats/_cli_run.py#L159
2. The cfg info finally saved in OutputProduct.product: https://github.com/opendatacube/odc-tools/blob/ba5fe6536027bb28a8660de475feb29ddf5a49fd/libs/stats/odc/stats/model.py#L539

In the step 1 and step 2, I believe they are just pass and assign values.

3. when system detect there is thumbnail_image setting here: https://github.com/opendatacube/odc-tools/blob/ba5fe6536027bb28a8660de475feb29ddf5a49fd/libs/stats/odc/stats/io.py#L243, it will call OWS build-in method: https://github.com/opendatacube/odc-tools/blob/ba5fe6536027bb28a8660de475feb29ddf5a49fd/libs/stats/odc/stats/io.py#L218-L221 and access OWS styling cfg to create the image bytes.
4. when we dump OWS cfg info to YAML at https://github.com/opendatacube/odc-tools/blob/ba5fe6536027bb28a8660de475feb29ddf5a49fd/libs/stats/odc/stats/io.py#L344, it looks like just reference to the dict from step 2. 

Any chance the step 3 may overwrite the cfg info in product?

SpacemanPaul: If so, it shouldn't be happening.  Looking into it now.

SpacemanPaul: Ah I see the problem.  I'll squeeze a fix into the upcoming OWS release.

supermarkion: Thanks a lot. I believe my current PR can be closed. Then I create a new PR to update OWS version in odc-stats side.

supermarkion:
We plan to change it on OWS side: https://github.com/opendatacube/datacube-ows/pull/754. No need to add the current feature on odc-stats side.

PR #443, opened 2021-10-21
fix save task bug

emmaai:
Fix the bug from my last PR


supermarkion (approved):
All good to me. Thanks a lot.

PR #444, opened 2021-10-22
Move plugins to a separate plugins package.

SpacemanPaul:
Next step in refactor.  Just tidying up at this stage.

Yaml files that identify their plugin by name should work transparently.  It looks like you can also identify plugins by their absolute path - if that's the case and any of your yaml files use that feature, they will need to be updated.


supermarkion (approved):
Just test this feature branch by installing it on local docker image test: https://github.com/opendatacube/datacube-docker/pull/67. All good on these four test cases. Thanks a lot for this change.

tebadi (approved):
Good work Paul!

PR #445, opened 2021-10-26
Plugin API cleanup refactor

SpacemanPaul:
Move product method out of Plugin API and replaced with helper method `product_for_plugin`.

Also removed some unused test code from pq plugin.

### DIFF
--- a/libs/stats/odc/stats/model.py
+++ b/libs/stats/odc/stats/model.py
@@ -522,92 +522,93 @@ class StatsPluginInterface(ABC):
         """
         return None
 
-    def product(
-        self,
-        location: str,
-        name: Optional[str] = None,
-        short_name: Optional[str] = None,
-        version: Optional[str] = None,
-        product_family: Optional[str] = None,
-        collections_site: str = "collections.dea.ga.gov.au",
-        producer: str = "ga.gov.au",
-        properties: Dict[str, Any] = dict(),
-        region_code_format: str = "x{x:02d}y{y:02d}",
-        naming_conventions_values: str = "dea_c3",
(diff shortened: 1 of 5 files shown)


Kirill888 (approved):

libs/stats/odc/stats/model.py
```diff
@@ -522,93 +522,94 @@ def rgba(self, xx: xr.Dataset) -> Optional[xr.DataArray]:
...
-            naming_conventions_values=naming_conventions_values,
-            explorer_path=explorer_path,
-            inherit_skip_properties=inherit_skip_properties,
-            preview_image_ows_style=preview_image_ows_style,
-            classifier=classifier,
-            maturity=maturity,
-            collection_number=collection_number,
-            nodata=nodata,
+            version=version_dashed,
+            version_dashed=version_dashed,
```
tebadi, line 582: Do we need to keep both `version` and `version_dashed `now that they point to the same value?

tebadi (approved):

libs/stats/odc/stats/model.py
SpacemanPaul: Probably not.  Just taking trying to keep PRs as small and reviewable as possible.

SpacemanPaul: It's a question of whether there are any scripts anywhere that use `{version_dashed}` in their location templates.

tebadi: We don't provide `version_dashed ` in our K8s stats calls.  If it's not used in odc-stats then it's probably safe to delete.

PR #446, opened 2021-10-26
Isolate plugin layer

SpacemanPaul:
Changed `StatsPluginInterface` API so it is independent of the (rest of the) `stats.model` layer.

Moved `StatsPluginInterface` to `plugins._base`, and moved the plugin registry code to a new module `plugins._registry`.

All plugin-related code is now cleanly isolated in the `stats.plugins` package, which has no internal dependencies on the rest of `odc.stats`.

### DIFF
--- a/libs/stats/odc/stats/plugins/_registry.py
+++ b/libs/stats/odc/stats/plugins/_registry.py
@@ -1,0 +1,54 @@
+import pydoc
+from typing import Callable, Dict, Type
+from functools import partial
+
+from ._base import StatsPluginInterface
+
+
+PluginFactory = Callable[..., StatsPluginInterface]
+
+_plugins: Dict[str, PluginFactory] = {}
+
+
+def _new(plugin_class, *args, **kwargs) -> StatsPluginInterface:
+    return plugin_class(*args, **kwargs)
+
+
+def resolve(name: str) -> PluginFactory:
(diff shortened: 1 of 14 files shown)


libs/stats/odc/stats/plugins/_registry.py
```diff
@@ -0,0 +1,54 @@
...
+    # TODO: make that more automatic
+    modules = [
+        "odc.stats.plugins.fc_percentiles",
+        "odc.stats.plugins.tcw_percentiles",
+        "odc.stats.plugins.gm",
+        "odc.stats.plugins.gm_ls_bitmask",
+        "odc.stats.plugins.pq",
+        "odc.stats.plugins.pq_bitmask",
+        "odc.stats.plugins.wofs",
+    ]
```
Kirill888, line 47: Just a note, with this now being in a separate folder it might better to use `__init__.py` inside to `odc/stats/plugins` to force imports.

Kirill888 (approved):

libs/stats/odc/stats/plugins/_registry.py
SpacemanPaul: Yeah, there's definitely more work to be done with import_all().   Even the name is wrong - it doesn't import all plugins at all, just the ones it knows about that live in odc-stats.

PR #447, opened 2021-10-27
Add ESA Worldcover indexing

alexgleith:
* Adds ESA Worldcover indexing tool
* Fixes naming of tests for the copernicus DEM indexing

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/esa_worldcover_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/esa_worldcover_to_dc.py
@@ -1,0 +1,220 @@
+#!/usr/bin/env python3
+"""
+Index the Copernicus DEM automatically.
+"""
+import concurrent.futures
+import logging
+import sys
+from math import ceil, floor
+from typing import Tuple
+
+import click
+import pystac
+import rasterio
+from datacube import Datacube
+from datacube.index.hl import Doc2Dataset
+from datacube.utils import read_documents
+from odc.apps.dc_tools.utils import bbox, index_update_dataset, limit, update_if_exists
+from rio_stac import create_stac_item
+
+from ._stac import stac_transform
+
+PRODUCT = (
+    "https://raw.githubusercontent.com/opendatacube/"
+    "datacube-dataset-config/master/products/esa_worldcover.odc-product.yaml"
+)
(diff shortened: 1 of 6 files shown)


carioca-au (approved):

apps/dc_tools/odc/apps/dc_tools/esa_worldcover_to_dc.py
```diff
@@ -0,0 +1,220 @@
...
+import rasterio
+from datacube import Datacube
+from datacube.index.hl import Doc2Dataset
+from datacube.utils import read_documents
+from odc.apps.dc_tools.utils import bbox, index_update_dataset, limit, update_if_exists
+from rio_stac import create_stac_item
+
+from ._stac import stac_transform
+
+PRODUCT = (
```
carioca-au, line 22: if you have a function to add the product and a flag to see if the product has to be added, shouldn't you send the path when you want to add or update the product instead have it hardcoded?

apps/dc_tools/odc/apps/dc_tools/esa_worldcover_to_dc.py
```diff
@@ -0,0 +1,220 @@
...
+    bounding_box,
+    limit: int,
+    update: bool,
+    n_workers: int = 100,
+) -> Tuple[int, int]:
+    doc2ds = Doc2Dataset(dc.index)
+
+    # Get a generator of (uris)
+    uris_tiles = list(get_tile_uris(bounding_box))
+    if limit:
```
carioca-au, line 148: Would be better to convert to a list just if you have a limit. There is no need to use a list if you will just loop over that

carioca-au (approved):
Just 2 minor sugestions

apps/dc_tools/odc/apps/dc_tools/esa_worldcover_to_dc.py
alexgleith: Could be.

apps/dc_tools/odc/apps/dc_tools/esa_worldcover_to_dc.py
alexgleith: there's only one product

PR #451, opened 2021-10-28
Stats plugins refactor

SpacemanPaul:
* Consolidate rgba method into base class.
* Consolidate input_data method calling load_with_native_transform into base class. (See notes below)
* Make handling of fuser and native_transform callbacks more consistent.
* Update tests to use new API.

input_data notes:

* Most plugins can completely drop their input_data method and use the base class implementation.
* WOFSFullHistory plugin uses straight dc.load instead of load_with_native_transform - overrides base implementation
* Geomedian base plugin has some additional code, but now extends the base implementation

I think the code is cleaner and easier to follow like this, but I'm open to feedback.


supermarkion (approved):
LGTM. Thanks a lot.

tebadi (commented):
Can we add some unit tests for the key functions?

PR #456, opened 2021-11-08
Fix gm s2 filters

NikitaGandhi:
filters is a tuple([("operation", <radius>), ...) not a dict anymore so I converted it to list object of operations and radius first and then get a max from radius list.

So not changing the implementation but fixed due to change in filters definition.


alexgleith (approved):

libs/algo/odc/algo/io.py
```diff
@@ -307,7 +307,7 @@ def fuser(xx: xr.Dataset) -> xr.Dataset:
     pad = kw.pop("pad", None)
     if pad is None:
         if filters is not None:
-            pad = max(*filters.values())
+            pad = max(list(zip(*filters))[1])
```
alexgleith, line 310: Actually, what is the logic here?

NikitaGandhi, line 310: filters is a tuple not a dict anymore so I converted it to list object of operations and radius first and then get a max from radius list. 

So not changing the implementation but fixed due to change in filters definition.

alexgleith: Ok, I trust you. It's pretty hard to read, but if you're happy I'm happy!

NikitaGandhi: I know it is hart to iterate through tuple but this works :)

PR #467, opened 2022-02-10
Fix ancillary data links

alexgleith:
Using the `relative` argument makes all links relative, whether they should be or not.

This change only makes them relative if their `href` is a subfolder of the `self` `href`.


alexgleith:
Sorry about formatting changes... upgraded black version.

apps/dc_tools/odc/apps/dc_tools/_stac.py
```diff
@@ -171,19 +172,29 @@ def _get_stac_bands(
...
 
-    def _get_path(asset):
-        path = asset["href"]
+    def _get_path(asset, force_relative=False):
+        path = URL(asset["href"])
         if relative:
-            path = Path(path).name
+            try:
+                path = path.relative_to(URL(self_link).parent)
+            except (ValueError, TypeError):
```
Kirill888: `TypeError` looks suss? when does this happen?

Kirill888: if it's when `self_link` is `None` then you should check that before attempting to compute parent of that link.

apps/dc_tools/odc/apps/dc_tools/_stac.py
```diff
@@ -171,19 +172,29 @@ def _get_stac_bands(
...
+                if force_relative:
+                    path = path.name
```
Kirill888, line 194: Should have at least some comment describing what happens and in what situations. I can only guess from looking? If asset doesn't share same path with the stac item, just use filename as relative path?

apps/dc_tools/odc/apps/dc_tools/_stac.py
alexgleith: Cleaned this up to make it more obviousl

apps/dc_tools/odc/apps/dc_tools/_stac.py
alexgleith: On it.

Kirill888 (approved):

PR #473, opened 2022-06-01
Added support to load_native_transform to apply different buffer size…

tebadi:
…s for different cloud classes.

### DIFF
--- a/libs/algo/odc/algo/io.py
+++ b/libs/algo/odc/algo/io.py
@@ -96,7 +96,6 @@ def _load_with_native_transform_1(
 
     xxs = native_transform(xx)
     xxss = []
-
     if groupby is not None:
         if fuser is None:
             fuser = _nodata_fuser  # type: ignore
@@ -108,10 +107,9 @@ def _load_with_native_transform_1(
 
             xx = xxss[0]
             for xs in xxss[1:]:
-                xx =  np.logical_and(xx, xs)
+                xx =  np.logical_or(xx, xs)
         else:
            xx = xxs.groupby(groupby).map(fuser)
-
 
     _chunks = None
     if chunks is not None:
@@ -308,7 +306,7 @@ def load_enum_filtered(
         else:
             _xxs = []
             for x in categories:
-                _xx = enum_to_bool(xx[band], (x, ))
+                _xx = enum_to_bool(xx[band], [x])
                 _xxs.append(xr.Dataset(
                     {band: _xx},
                     attrs={"native": True},  # <- native flag needed for fuser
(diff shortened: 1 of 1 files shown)


emmaai (changes requested):
Sorry it's too hacky for me. Details inline.

libs/algo/odc/algo/io.py
```diff
@@ -84,17 +86,30 @@ def _load_with_native_transform_1(
...
+
     if pad is not None:
         load_geobox = gbox.pad(load_geobox, pad)
 
     mm = ds.type.lookup_measurements(bands)
     xx = Datacube.load_data(sources, load_geobox, mm, dask_chunks=load_chunks)
-    xx = native_transform(xx)
 
+
+    xxs = native_transform(xx)
```
emmaai, line 97: `native_transform` is to return a `xr.Dataset`, all the operations in this function are based on this assumption. Even if the mask bands need to be merged here (which I don't agree, but it's separate), they don't have to be in a list but as variables in the same `Dataset`.

libs/algo/odc/algo/io.py
```diff
@@ -84,17 +86,30 @@ def _load_with_native_transform_1(
...
+            idx = 0
+            for x in xxs:
+                xxss.append(x.groupby(groupby).map(fuser, [idx]))
+                idx += 1
+
+            xx = xxss[0]
+            for xs in xxss[1:]:
+                xx =  np.logical_or(xx, xs)
+        else:
+           xx = xxs.groupby(groupby).map(fuser)
```
emmaai, line 112: I don't agree the mask bands should be merged here. Especially when `fuser` is involved. It could cause the same issue that I observed in wofs plugin. When the datasets are in the same crs, they're merged in native crs, when not, they're merged in reprojected crs. We need some consistency here, either before or after reprojection.

libs/algo/odc/algo/io.py
```diff
@@ -281,35 +297,51 @@ def load_enum_filtered(
     """
 
     def native_op(xx: xr.Dataset) -> xr.Dataset:
-        _xx = enum_to_bool(xx[band], categories)
-        return xr.Dataset(
+        if len(np.array(filters).shape) < 3:
```
emmaai, line 300: For all the change below, I don't think we should hack inside, but consider to change the interface of this function. Currently it accepts a union of values and convert the input band to a boolean `xr.DataArray`. It could accepts a list of values and convert each of them to a boolean `xr.Dataset` with mask/s as variable/s. Again we should avoid the list for `dask`'s sake, since it'll introduce unnecessary troubles in the future development.

tebadi:
Filtering will happen at the stats level specifically at the geomedian plugin.  This PR is redundant.

PR #483, opened 2022-06-24
rework test_stac_tranform

### LINKED ISSUE #481: test cases for stac_transform are incomplete and not integrated to odc-core
Test file: https://github.com/opendatacube/odc-tools/blob/develop/apps/dc_tools/tests/test_stac_transform.py

### Issues

1. The test cases uses `deepdiff` to compare originally stac file convert to odc dataset yaml file with an expected odc dataset yaml file. This `deepdiff` should be replaced with functions available from `datacube-core`
2. https://github.com/opendatacube/odc-tools/blob/62c7356b9728f6cbd217a8be9dc5dece7cddf57e/apps/dc_tools/tests/test_stac_transform.py#L25 and https://github.com/opendatacube/odc-tools/blob/62c7356b9728f6cbd217a8be9dc5dece7cddf57e/apps/dc_tools/tests/test_stac_transform.py#L29 are incomplete tests that should either be marked with `pytest.mark.skip` or be completed to serve a purpose

pindge:
fixes #481 

replace `deepdiff` with `get_doc_changes` from datacube-core

### DIFF
--- a/apps/dc_tools/tests/test_stac_transform.py
+++ b/apps/dc_tools/tests/test_stac_transform.py
@@ -17,11 +17,11 @@ def test_sentinel_stac_transform(sentinel_stac, sentinel_odc):
     assert len(get_doc_changes(transformed_stac_doc, sentinel_odc)) == 2
 
 
-@pytest.mark.skip(reason="Incomplete test")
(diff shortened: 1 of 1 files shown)


SpacemanPaul (approved):
Nice - so much cleaner and more readable!

PR #486, opened 2022-06-28
add test case for s3-to-dc single absolute s3 path uri

### LINKED ISSUE #485: s3-to-dc does not index when s3 url doesn't contain *

pindge:
- allow s3-to-dc to index when provided `uri` is a absolute s3 path


pindge:
Won't merge

PR #489, opened 2022-06-28
add instruction for local development with conda

### LINKED ISSUE #488: README is missing instruction on local setup that simulates action workflow
The `main.yaml` workflow uses mamba to build test environment, but there is no instruction on how to setup test locally.

pindge:
add instruction for local testing using the steps used in github action workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,9 @@ dc-index-from-tar --protocol gs --env mangroves --ignore-lineage metadata.tar.gz
 Local Development
 =================
 
+Docker
+------
+
 Requires docker, procedure was only tested on Linux hosts.
 
 ```bash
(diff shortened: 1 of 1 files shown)


omad:
Thanks Pin, I think this looks like a good improvement.

I think we should go further still however, and extract this code into a script. Something like `scripts/run-tests`, that could even include a few more smarts so that it only regenerates the database and python environment if required.

There's a few reasons for this suggestion:
- It's important to make it as simple and fast as possible to run the tests Just one command, not copying and pasting multiple pieces.
- Have exactly the same script and steps run locally as well as in CI, to avoid chasing failures due to different environments.
- To avoid drift between the different setup steps.

omad (approved):

PR #492, opened 2022-07-05
add landsat9 product name

mohamad-mohammad:
This will fix the error raised when trying to index Landsat Collection 2 Level-2 UTM Surface Reflectance (SR) (landsat-c2l2-sr) with `stac-to-odc` tool
the error(odc.apps.dc_tools.utils.IndexingException: Failed to create dataset with error No matching Product found for dataset ...)


Kirill888 (approved):
thanks for this @mohamad-mohammad 

there is now a more robust method for transforming stac into datacube datasets

https://odc-stac.readthedocs.io/en/latest/_api/odc.stac.eo3.stac2ds.html#odc.stac.eo3.stac2ds

but that was originally developed for loading data directly from stac and so lacks tooling for indexing right now,

PR #493, opened 2022-07-08
Option to overwrite product name

SeffatChowdhury:


SeffatChowdhury:
@alexgleith please review. Note one issue with the 9-class collection, it didn't return the number of item in the pystac client search, hence not all datasets were indexed (it indexed the first 100, not sure where that limit is set)

alexgleith (approved):
Looks good, Seffat! Thanks for this.

PR #499, opened 2022-08-16
add skipped for dataset not indexed but skipped

pindge:
- added is misleading when clearly no action was performed but skipped
- this will increase clarity for metrics

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/utils.py
+++ b/apps/dc_tools/odc/apps/dc_tools/utils.py
@@ -26,6 +26,13 @@ class IndexingException(Exception):
     pass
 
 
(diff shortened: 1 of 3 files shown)


jmettes (approved):
This is great - fixing something that's confused me.

apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -177,6 +184,9 @@ def index_update_dataset(
                         )
                 else:
                     logging.warning("Dataset already exists, not indexing")
```
omad, line 186: We know the `id` of the dataset we're processing here, it'd be worth including in the log output.

PR #500, opened 2022-08-24
support kwargs in reproject

emmaai:
Support `kwargs` taken by `rasterio.warp` as gdal options.
e.g.,
```
kwargs = {"COORDINATE_OPERATION": '+proj=pipeline +step +inv +proj=utm +zone=49 +ellps=WGS84 \
        +step +proj=aea +lat_0=0 +lon_0=132 +lat_1=-18 +lat_2=-36 +x_0=0 +y_0=0 +ellps=GRS80'}

xr_reproject(src_data, dst_geobox, **kwargs)
```  
Changes:
- add function to choose the transform path explicitly by given "authority:code", e.g., `epsg:9688`
- suppress the superfluous warnings from `numpy`, the computation is checked and doing as intended
- require `rasterio>=1.3.2` since the transform path only works this version thereafter
- adopt pre-commit check and lint

Note that the meaningful changes happened in `_warp.py` and `io.py`, the rest is merely trivial formatting by `black`.


SpacemanPaul (approved):
There's a little bit of dask stuff that I don't totally follow, but otherwise looks good.

libs/algo/odc/algo/io.py
```diff
@@ -52,6 +56,32 @@ def compute_native_load_geobox(
     )
 
 
+def choose_transform_path(
+    src_crs: str, dst_crs: str, transform_code: Optional[str] = None
+) -> str:
+    # generally it is a bad idea to make aoi global
+    # here is to accommodate the fact that 326xx not overlapping with 3577
+    area_of_interest = aoi.AreaOfInterest(-180.0, -90.0, 180.0, 90)
```
SpacemanPaul: Maybe you should be able to pass in an area of interest?  It would make sense (for us) to be able to restrict to Australia-ish for example, but obviously Africa (and Antarctica) would need to be able to specify their own AOIs.

In an ideal world where people used CRSs properly we could default AOI to the valid region of the destination CRS.

SpacemanPaul (approved):
Mostly makes sense to me.  The global Area of Interest makes me uncomfortable but I understand why we need it.

libs/algo/odc/algo/io.py
emmaai: agreed. fixed in the commit below.

SpacemanPaul (approved):
Thanks Emma.

Kirill888:
@emmaai @SpacemanPaul what's up with rasterio pin in this PR? This is very bleeding edge dependency, and you only "need it" to run tests, normal usage can keep working with older versions just fine.

Emma can you please commit large formatting changes like this in a separate commit in the future, it makes it harder to review during the PR process, but also makes history less useful going forward as there are a lot of unrelated changes scattered all over the code-base now that claim they were needed to enable this feature.

SpacemanPaul:
rasterio>=1.3.2 is required for Python3.10...

We could generalise to `rasterio!=1.3.0,!=1.3.1` or something fairly safely though I guess.

emmaai:
rasterio>=1.3.2 also supports gdal options and transform options correctly, which is not on 1.2.10.

re formatting, my idea was to gradually correct them with actual changes needed. Otherwise, there is too much manual work to please pylint and flake. If the trouble of keeping track outweighs the  benefit of auto-formatting, then I guess I can remove precommit in the next pr.

Kirill888:
@emmaai when working with older code I usually apply formatting first and check it in using one large commit that changes nothing else, then do actual change, ideally formatting fixes would be a separate PR as well...

> rasterio>=1.3.2 also supports gdal options and transform options correctly, which is not on 1.2.10.

is it actual `rasterio` change that fixes it, or is it just gdal version that older `rasterio` shipped with? Can you point me to some issues on github describing the problem you have experienced?

emmaai:
I'll keep this in mind next time I change things.

The issue was raised here https://github.com/rasterio/rasterio/issues/2354

This commit fixed the issue:
https://github.com/rasterio/rasterio/commit/58b8e7095f5ed176182a7faa038da989d49961a0#diff-06869bf8e8a5bab7295de3b08a914e9b4d536d8c10c6403d7ea05063b83533d1L487

The transform option should not be "uppered".  If so, gdal would fail to recognise that option. I didn't use the gdal lib shipped with `rasterio`, but the local one.

PR #503, opened 2022-08-29
add skipped to stac-api-to-dc

### LINKED ISSUE #358: Return number skipped as well as success and failed from stac_api_to_odc
*Simply a suggestion:*
When adding datasets using stac_api_to_odc (and I assume other equivalent functions), it would be useful to be able to differentiate between datasets that have been legitimately skipped (because they are already in the database) vs those which have actually been added. 

If the function could return success, failure and skipped, that would be very useful, but perhaps this is tricky given the way that the success and failure are currently calculated, whereas "skipping" is handled inside index_update_dataset() of utils.py

This would help report more accurate information about whether there was really any new data added to the DB.

pindge:
fixes #358 

- Add skipped reporting to `stac-api-to-dc` and `cop-dem`
- extend test case to cover skipped
- fix cop-dem test


SpacemanPaul (approved):
Looks good.

PR #505, opened 2022-09-15
fix due to gdal 3.5.2

emmaai:
It is probably not a good practice to send an arg with nothing. `gdal>=3.5.0` is not happy with it, hence the fix.


omad (approved):
I don't understand all of this, but I think it makes sense 🤷

🤞👍

PR #507, opened 2022-10-05
Dataset maturity handling

SpacemanPaul:
Changes to all dc-tools indexing CLIs to add `--archive-less-mature` flag.

If set, then for each indexed dataset, if there is a dataset in the product that has matching `region_code` and `time` values but a less mature value of `dataset-maturity` then the less mature dataset is archived as the new dataset is indexed.  Both operations take place in an atomic transaction, using the new database transaction API introduced in datacube-1.8.8.


apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -156,52 +166,100 @@ def index_update_dataset(
...
+        if archive_less_mature:
+            dupes =  dc.index.datasets.search(
+                product=ds.type,
+                **dupe_query
             )
+            for dupe in dupes:
+                if dupe.id == ds.id:
+                    # Same dataset, for update.  Ignore
+                    continue
+                if dupe.metadata.dataset_maturity <= ds.metadata.dataset_maturity:
```
mpaget, line 232: I'm curious: How does this line (212) work?
I see that typical values are 'final|interim|nrt' but are these a controlled vocabulary and how/where are their order defined?

SpacemanPaul: The "allowed" values are defined in the EO3 definition at [datacube/index/default-metadata-types.yaml:42](https://github.com/opendatacube/datacube-core/blob/develop/datacube/index/default-metadata-types.yaml#L41-L44) and are: `nrt, interim, final` listed in of increasing maturity and decreasing alphabetic order.

There is no code that enforces the use of only these values, so if somebody decided to use new dataset maturity values that didn't neatly sort in reverse alphabetic order, they would have problems if they tried to use this code.  But the supported values are well documented (and have been for quite some time), and the `dataset_maturity` metadata element is defined within the `dea` namespace, so I feel comfortable with this implementation.

(Apart from the fact the test is still failing.)

I guess nothing explicitly says that nrt, interim, final is the order of increasing maturity (outside of internal DEA documentation) but it seems to me be pretty clear implicitly - I can't see why amyone would sort those values in any other order.

apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -156,52 +166,100 @@ def index_update_dataset(
...
+        )
+    if ds is None:
+        raise IndexingException(
+            f"Failed to create dataset with error {err}\n The URI was {uri}"
+        )
+
+    if archive_less_mature:
+        if archive_less_mature == True:
+            # if set explicitly to True, default to [region_code, time]
+            archive_less_mature = ["region_code", "time"]
```
emmaai, line 208: alright, now I understand what `archive_less_mature` kinda changed between `bool` and `None` here and there.

emmaai (approved):
The logic makes sense to me and there is a test so I'd be lazy in checking all the details. :P

apps/dc_tools/odc/apps/dc_tools/utils.py
SpacemanPaul: Yes, the idea is you could pass in a different list of columns to group by if you really wanted.

`None` and `False` mean don't do maturity checking and `True` means do maturity checking, grouping by `region_code` and `time`.   The CLIs only use True and False, but it is extensible in future if we need to group by something else.

SpacemanPaul: I'll try and make the comments a bit clearer.

apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
```diff
@@ -64,8 +62,8 @@ def cli(input_directory, update_if_exists, allow_unsafe, stac, statsd_setting, g
     for in_file in files_to_process:
         with in_file.open() as f:
             try:
-                if in_file.endswith(".yml") or in_file.endswith(".yaml"):
-                    metadata = yaml.safe_load(f, Loader=Loader)
```
omad, line 68: It is almost definitely worth keeping the `CLoader` for YAML parsing. For bulk loading it makes a big difference, just depends on where this tool is used.

apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -153,54 +163,122 @@ def index_update_dataset(
...
+    :param archive_less_mature: Enforce dataset maturity.
+           * If None (the default) or False or an empty iterable, ignore dataset maturity.
+           * If True, enforce dataset maturity by looking for existing datasets with same product, region_code and time
+             values. If a less mature match is found, it is archived and replaced with the new dataset being inserted.
+             If a match of the same or greater maturity is found an IndexException is raised.
+           * If an iterable of valid search field names is provided, it is used as the "grouping" fields for
+             identifying dataset maturity matches.
+             (i.e. `archive_less_mature=True` is the same as `archive_less_mature=['region_code', 'time'])
+    :return: Returns nothing.  Raises an exception if anything goes wrong.
+    """
```
omad, line 190: ♥

omad (approved):

apps/dc_tools/odc/apps/dc_tools/fs_to_dc.py
SpacemanPaul: I would agree but `yaml.safe_load` doesn't take a Loader parameter - only `yaml.load` does.

PR #509, opened 2022-10-11
Docker py3.10

SpacemanPaul:
* Move test Docker image to osgeo/gdal (Ubuntu 22.04/Python3.10) from geobase which is now unmaintained.
* Increase minimum Python version to 3.7 for odc-cloud and odc-dc-tools.
* Remove `loop` parameter from `asyncio.Queue()` calls.  This parameter has been deprecated since Py3.7 and is removed in Py3.10.

(Split from #507)


alexgleith (approved):
Looks good to me!

emmaai (approved):
💯

PR #513, opened 2022-10-18
Add json parsing options

alexgleith:


apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
```diff
@@ -39,6 +35,11 @@ def _parse_options(options: Optional[str]) -> Dict[str, Any]:
         for option in options.split("#"):
             try:
                 key, value = option.split("=")
+
+                try:
+                    value = json.loads(value)
```
alexgleith, line 40: This is the important change.

It allows you to send in a query, like this: `--options="query={\"platform\":{\"in\":[\"landsat-8\"]}}"` which works on the Microsoft PC like this: 

`https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=landsat-c2-l2&limit=1&query={%22platform%22:%20{%22in%22:%20[%22landsat-8%22]}}`

SpacemanPaul (approved):
LGTM

PR #514, opened 2022-10-18
Various cleanups and proper handling of SkippedException

alexgleith:
The SkippedException was being handled by raising an Exception, which it shouldn't.


SpacemanPaul (approved):
All good.

PR #519, opened 2022-11-01
Publish index and archive actions to SNS topic

### LINKED ISSUE #516: Publish index actions to SNS topic
With the introduction of the [`--archive-less-mature` maturity functionality](https://github.com/opendatacube/odc-tools/pull/507), now less mature (i.e., `NRT` and `interim`) datasets get archived on indexing of more mature datasets (i.e., `final`).

We have a need to also delete these datasets from S3, beyond just archiving them - as the NRT data can get quite large, and they're not needed anymore. It would be really handy if dc-tools were to publish `archive` actions to an SNS topic (i.e., the STAC document + message attribute: `{"status": "ARCHIVED", ...}`). This way, a job can handle tagging the relevant objects in S3 for deletion.

---

Additionally, this functionality might also be handy for indexing events. We have downstream derivative products (wofs and fc) that are processed on arrival of upstream ARD datasets. However, currently they run into race-conditions, because they are triggered for based on SNS events triggered by S3 bucket notifications, rather than by indexing events. So it might be more sensible to drive this downstream processing based on that actual availability of data in the datacube index (i.e., STAC document + message attribute: `{"status": "ADDED"}`, as opposed to their arrival on S3.

Ariana-B:
In response to [issue #516](https://github.com/opendatacube/odc-tools/issues/516), publishing indexing or archiving action to an SNS topic.

- Added a flag `--publish-action [sns topic arn]` which publishes action ("ARCHIVED"/"ADDED") and STAC document to the SNS topic when set. 
- SNS publication function defined in queue lib since it is similar/related to other SQS functionality (and wasn't sure where else would be better).
- Wrote function `ds_to_stac` to convert from ds eo3 metadata to STAC
- Updated dc-tools readme to include `archive-less-mature` and `publish-action` flags
- Added a script that sets up the testing database
- Updated test instructions in README and added tests for sns publication


libs/cloud/odc/aws/queue.py
```diff
@@ -180,3 +184,16 @@ def get_messages(
...
 
     return itertools.islice(messages, limit)
+
+
+def publish_to_topic(action, stac):
+    """
+    Publish a message to an sns topic
+    """
+    sns = boto3.resource("sns")
+    sns_topic = sns.create_topic(Name=SNS_TOPIC_NAME)
```
jmettes: We'll have a topic already existing - `dea-public-data-sentinel-2`, and will want the tool to accept the topic name (or ARN [the Absolute Resource Name in AWS]) as a parameter - e.g., `--publish arn:aws:sns:ap-southeast-2:xxxxxxxxxxxxxxxxx:dea-public-data-sentinel-2` (where the xxx's are the account number, which I've forgotten)

libs/cloud/odc/aws/queue.py
```diff
@@ -180,3 +184,16 @@ def get_messages(
...
+
+    stac = json.dumps(stac)
+    sns_topic.publish(
+        Message='{"action": {action}, "stac": {stac}}',
```
jmettes: I think you probably want

```
message_attributes = {
  "action": action,
   ... # more attributes? e.g., product_name?
}
sns_topic.publish(Message=stac, MessageAttributes=message_attributes)
```

jmettes:
Awesome, thanks so much for working on this!

libs/cloud/odc/aws/queue.py
jmettes: prob enough to use [SNS client](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#SNS.Client.publish)

libs/cloud/odc/aws/queue.py
```diff
@@ -180,3 +181,30 @@ def get_messages(
...
 
     return itertools.islice(messages, limit)
+
+
+def publish_to_topic(topic_name: str, action: str, stac: dict):
+    """
+    Publish 'added' or 'archived' action to the provided sns topic
+    """
+    sns = boto3.client("sns")
+    sns_topic = sns.create_topic(Name=topic_name)
```
jmettes: It's not really the job of these tools to create infrastructure like SNS topics - we set all that up elsewhere. I think this code should accept an already existing topic (i.e., the ARN), and not create it.

libs/cloud/odc/aws/queue.py
```diff
@@ -180,3 +181,30 @@ def get_messages(
...
+
+    stac = json.dumps(stac)
+    message = {
+        "action": action,
+        "stac": stac,
+    }
+    sns.publish(
+        TopicArn=sns_topic.get("TopicArn"),
+        Message=json.dumps(message),
+        MessageAttributes={
```
jmettes: Is it worth adding more `MessageAttributes` such as the `product` (so we can filter on it, beyond just whether it was added/archived)? Maybe `maturity` is worth adding too?

libs/cloud/odc/aws/queue.py
```diff
@@ -180,3 +181,30 @@ def get_messages(
...
+    sns_topic = sns.create_topic(Name=topic_name)
+
+    stac = json.dumps(stac)
+    message = {
+        "action": action,
+        "stac": stac,
+    }
+    sns.publish(
+        TopicArn=sns_topic.get("TopicArn"),
+        Message=json.dumps(message),
```
jmettes: The `Message` payload is just the STAC document itself. The `MessageAttributes` are used to describe it, so they can be filtered on when receiving lots of messages.

Our convention is as follows:

```
sns.publish(Message=stac_document, MessageAttributes={
    "action": { "DataType": "String", "StringValue": action },
    "product": {
        "DataType": "String", 
        "StringValue": stac_document["odc:product"] if "odc:product" in stac_doc else stac_document["collection"]
    },
    "maturity": {
        "DataType": "String",
        "StringValue": stac_doc["properties"]["dea:dataset_maturity"]
    }
})
```

Other message attributes I've come across are: `bbox.ll_lon`, `bbox.ll_lat`, `bbox.ur_lon`, `bbox.ur_lat`, `datetime`, `cloudcover`.

libs/cloud/odc/aws/queue.py
```diff
@@ -180,3 +181,30 @@ def get_messages(
...
+        MessageAttributes={
+            "action": {
+                "DataType": "String",
+                "StringValue": "ARCHIVED {or} ADDED",
+            },
+            "stac": {
```
jmettes: We don't want to include the STAC doc inside the message attributes - that's the message payload

libs/cloud/odc/aws/queue.py
```diff
@@ -180,3 +181,30 @@ def get_messages(
...
+        "action": action,
+        "stac": stac,
+    }
+    sns.publish(
+        TopicArn=sns_topic.get("TopicArn"),
+        Message=json.dumps(message),
+        MessageAttributes={
+            "action": {
+                "DataType": "String",
+                "StringValue": "ARCHIVED {or} ADDED",
```
jmettes: You'll want to pass the `action` variable value here

apps/dc_tools/README.md
```diff
@@ -126,6 +126,12 @@ Options:
                                   include
 
   --absolute                      Use absolute paths when converting from stac
+
+  --archive-less-mature           Find less mature versions of the dataset and
+                                  archive them
+                                  
+  --publish-action TEXT           Publish indexing action to SNS topic
```
jmettes: Maybe `TEXT` be `SNS topic ARN` for clarity

libs/cloud/odc/aws/queue.py
Ariana-B: `create_topic` only creates a topic if it doesn't exist, otherwise it simply returns the arn. This is the only way to get a topic arn based off the name, though I can change the flag to take in the arn instead if that's preferable.

libs/cloud/odc/aws/queue.py
Ariana-B: What would the expected behaviour be for products that don't have `maturity`?

libs/cloud/odc/aws/queue.py
jmettes: Oh I see. I don't think `odc-tools` should create infra. Rather do that explicitly elsewhere. I think ARN is fine

libs/cloud/odc/aws/queue.py
jmettes: Not sure, haven't come across that. I'd guess don't pass it as a `MessageAttribute` if it's not in the STAC document.

apps/dc_tools/README.md
Ariana-B: Done

libs/cloud/odc/aws/queue.py
Ariana-B: Done

libs/cloud/odc/aws/queue.py
Ariana-B: Fixed

libs/cloud/odc/aws/queue.py
Ariana-B: Done

libs/cloud/odc/aws/queue.py
Ariana-B: Changed

apps/dc_tools/README.md
```diff
@@ -178,6 +184,11 @@ Options:
   --request-payer                 Needed when accessing requester pays public
                                   buckets
 
+  --archive-less-mature           Find less mature versions of the dataset and
+                                  archive them
+
+  --publish-action TEXT           Publish indexing action to SNS topic
```
jmettes: SNS ARN here too?

apps/dc_tools/tests/test_sns_publishing.py
```diff
@@ -0,0 +1,261 @@
...
+    assert len(messages) == 1
+    message_attrs = json.loads(messages[0]["Body"]).get("MessageAttributes")
+    assert message_attrs["action"].get("Value") == "ADDED"
+    assert message_attrs["product"].get("Value") == "ga_ls5t_nbart_gm_cyear_3"
+    assert message_attrs["maturity"].get("Value") == "final"
+
+
+@mock_sns
+@mock_sqs
+def test_sqs_publishing_archive(aws_credentials, aws_env, sqs_message):
```
jmettes: The only test case this doesn't cover is the scenario of using `--archive-less-mature`  + `--publish-action`. Might be worth adding, because it's the main thing I'll immediately be using it for

jmettes: Or is this essentially covered via functionality of `---archive`?

Ariana-B: Added

jmettes (approved):
Woohoo! tyvm @Ariana-B - this looks fantastic. well done navigating the jungle that is odc-tools

apps/dc_tools/tests/test_sns_publishing.py
```diff
@@ -0,0 +1,354 @@
...
+    attrs = sqs.get_queue_attributes(
+        QueueUrl=queue.get("QueueUrl"), AttributeNames=["All"]
+    )
+    queue_arn = attrs["Attributes"]["QueueArn"]
+    sns_arn = topic.get("TopicArn")
+    sns.subscribe(
+        TopicArn=sns_arn,
+        Protocol="sqs",
+        Endpoint=queue_arn,
+    )
```
omad: These lines could go nicely into a fixture, the fixture itself could be annotated with `@mock_sns` and `@mock_sqs`, and could provide the SNS ARN and a [boto3 SQS Queue](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#queue) ready to receive messages from.

This would reduce the duplication and make the purpose of the tests easier to follow.

apps/dc_tools/odc/apps/dc_tools/_stac.py
```diff
@@ -404,3 +413,21 @@ def stac_transform(input_stac: Document, relative: bool = True) -> Document:
...
+def ds_to_stac(ds: Dataset) -> dict:
+    """Get STAC document from dataset with eo3 metadata"""
+    if ds.is_eo3:
+        if not ds.uris:
+            raise ValueError("Can't find dataset location")
+        location = ds.uris[0]
+        stac = to_stac_item(
+            from_doc(ds.metadata_doc, skip_validation=True),
+            location,
+            explorer_base_url="https://explorer.dea.ga.gov.au",
```
omad: If this is optional, we should leave it out instead of defaulting to DEA. We're not the only ones to use these tools.

tests/test-env-py38.yml
```diff
@@ -32,6 +32,7 @@ dependencies:
   - lxml # needed for thredds-crawler
   - urlpath
   - datadog
+  - eodatasets3
```
omad, line 35: Can you please add this new dependency into `apps/dc_tools/setup.cfg` too. 🙏

omad (changes requested):
Looks great Ariana. Lets try to merge this today before I get too stuck into updating tests and linting errors (conflict hell).
I love the tests!

apps/dc_tools/tests/test_sns_publishing.py
Ariana-B: Done

apps/dc_tools/odc/apps/dc_tools/_stac.py
Ariana-B: Removed

tests/test-env-py38.yml
Ariana-B: Done

omad (approved):
Yay! Thanks!

omad:
I've rebased these changes and force-pushed them, but there are still some failing tests. Some are due to internal stuff, some due to external data changes.

I'm going to merge this, and keep working on fixes in a new branch and PR.

Issue #521, opened 2022-11-02
Improve Developer Testing


omad:
In **ODC Tools** we have a several automated tests written for some of the tools and libraries. However, having recently tried to do some work on the code here, I've run across a few hiccups.

**Problems**
- Most tests required an available PostgreSQL database to be setup.
- Interdependencies between tests, many tests have been written such that they depend on previous tests being run. This makes it impossible to run individual tests during development.

The tests have been setup with [pytest-depends](https://pypi.org/project/pytest-depends/), but it's being used in a manner other than advised.

From the third paragraph documentation of **pytest-depends**:
> This isn't a dependency in the sense of test A sets up stuff for test B, but more in the sense of if test A failed there's no reason to bother with test B either.

Two different options for development setups have been documented in [README.md#Local Development](https://github.com/opendatacube/odc-tools#local-development), but this does nothing to solve the test interdependency problem to be able to run individual tests, and is awkward at best in automating a PostgreSQL setup.

omad:
**Another Issue**

Tests written for the AWS capable tools depend on S3 resources hosted by third parties, and some of those have recently become unavailable. Some of the STAC documents store in `s3://sentinel-cogs/` are no longer publicly readable.

I tried replacing the reads of real S3 documents with [moto](https://github.com/spulec/moto), but failed due to the use of [aiobotocore](https://github.com/aio-libs/aiobotocore), see: #522 .

An alternate possible solution for this could be using a mock server after implementing a solution to #470 . But using [moto](https://github.com/spulec/moto) would have been simpler.

omad:
Done

PR #523, opened 2022-11-09
Support changing the S3 Endpoint

### LINKED ISSUE #470: AWS_S3_ENDPOINT env for s3-to-dc
Lately we have been producing STAC documents and indexing them with something like `s3-to-dc --stac s3://glob/*/*json`. However, our data is not stored on AWS S3, it is stored on-prem in a Ceph storage cluster. It appears that `s3-to-dc` only supports AWS buckets by default.

The solution right now is a very quick hack in our Dockerfile
```
FROM docker.io/opendatacube/datacube-index:0.1.5

RUN sed -i '1i import os' /usr/local/lib/python3.8/dist-packages/odc/aio/__init__.py
RUN sed -i 's/create_client("s3"/create_client("s3", endpoint_url=os.environ["AWS_S3_ENDPOINT"]/g' /usr/local/lib/python3.8/dist-packages/odc/aio/__init__.py
```

The AWS_S3_ENDPOINT exists in GDAL (https://gdal.org/user/virtual_file_systems.html#vsis3-aws-s3-files) and ODC picks up on this when loading data. Using it for indexing as well would be nice!

omad:
There are multiple tests failing at the moment that this PR does not address. They are due to external changes changes in the Sentinel-2 STAC metadata.

This PR adds support for changing the S3 Endpoint to be used, by setting the environment variable `AWS_S3_ENDPOINT`. This will allow changing the tests to use a local `moto` server instead of relying on external data sources for tests, which have changed and broken tests, and are liable to do so again in the future.

See #470 and #521.

Closes: #470


omad:
There's also loads of linting and pre-commit checks that are failing, since they've never been enabled or enforced. I tried fixing them here, but it was getting out of hand, so I'm going to merge and fix later.

The failing checkers are
- yamllint (when run via pre-commit it doesn't use the config file in this repo)
- flake8
- pylint

PR #524, opened 2022-11-14
Improve Developer Testing Experience

omad:
# Background
See #521 

# Changes in this PR
- Add [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) which automatically start and stop a PostgreSQL Docker Container, and provision an ODC Database containing Metadata Types and Products. This is just for the `odc.apps` tests.
- Use [moto](http://docs.getmoto.org/en/latest/) to mock AWS services instead of relying on external S3 objects, which have recently changed and broken tests.
- Remove [pytest-depends](https://pypi.org/project/pytest-depends/) and replace with [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) , making it possible to run any test individually.
- Unfortunately: Lots of linting and formatting related fixes


omad:
As described in #526 I'm going to fix the outstanding linting issues caught by pre-commit.ci in a follow up PR.

apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -273,7 +269,7 @@ def index_update_dataset(
                     updates = {tuple(): changes.allow_any}
                 # Do the updating
                 try:
-                    dc.index.datasets.update(ds, updates_allowed=updates)
+                    dc.index.datasets.update_flag(ds, updates_allowed=updates)
```
Ariana-B: accidental replace?

apps/dc_tools/tests/conftest.py
```diff
@@ -81,12 +169,144 @@ def final_dsid():
...
+def odc_test_db_with_products(odc_db: Datacube):
+
+    local_csv = str(Path(__file__).parent / "data/example_product_list.csv")
+    added, updated, failed = add_update_products(odc_db, local_csv)
+
+    assert failed == 0
+    yield odc_db
 
 
 @pytest.fixture
```
Ariana-B, line 308: Should this fixture also be updated?

libs/cloud/odc/aws/_find.py
```diff
@@ -64,7 +66,7 @@ def is_glob(s):
         elif "**" not in qq_set:
             depth = len(qq)
         else:
-            raise ValueError("Bad query: %s" % url_query)
+            raise ValueError("Bad query: {url_query}")
```
Ariana-B: Missing an `f` in front

Ariana-B (dismissed):
Let's either determine what's causing the pylint warning with `update` and `update-if-exists` and revert the name change, or update the `cli` parameters in each file accordingly

Ariana-B (commented):
Other than the couple minor comments, looks good!

apps/dc_tools/tests/conftest.py
omad: Yep, thanks!

omad:
Thanks for catching those Ariana!

PR #535, opened 2022-11-30
updated world cover date, name

rokrak1:


omad (approved):
Thank you!

omad:
Hi rokrak1!

I should have thought harder about this before merging it, because I now think it should work in a different way.

We shouldn't need two separate ODC Products for the 2 years of ESA WorldCover, the better solution is to have all the datasets under the same Product, but with different start/end dates. This will fit better with how other tools expect ODC Data to be structured.

We should still allow choosing which year you need.

omad:
Okay, more thought, this is totally fine, thank you again!

rokrak1:
> Hi rokrak1!
> 
> I should have thought harder about this before merging it, because I now think it should work in a different way.
> 
> We shouldn't need two separate ODC Products for the 2 years of ESA WorldCover, the better solution is to have all the datasets under the same Product, but with different start/end dates. This will fit better with how other tools expect ODC Data to be structured.
> 
> We should still allow choosing which year you need.

Hi Damien  :) I thought so too, but our researcher said these are two separate products.

PR #541, opened 2023-01-31
fix xarray groupby

emmaai:
`groupby` has trouble on `MultiIndex` since `xarray > 2022.3.0`. 

The fix/workaround:

- remove `MultiIndex`
- set index on the coords on which to `groupby`, e.g., `solar_day`
- apply `groupby` and `map`
- reinstate `MultiIndex`

### DIFF
--- a/libs/algo/odc/algo/_masking.py
+++ b/libs/algo/odc/algo/_masking.py
@@ -689,23 +689,40 @@ def _fuse_with_custom_op(x: xr.DataArray, op, name="fuse") -> xr.DataArray:
 
     Expects data in _,y,x order. Works on Dask inputs too.
     """
-    if dask.is_dask_collection(x.data):
-        data = _da_fuse_with_custom_op(x.data, op, name=name)
+    if x.shape[0] > 1:
+        if dask.is_dask_collection(x.data):
+            data = _da_fuse_with_custom_op(x.data, op, name=name)
+        else:
+            slices = [x.data[i : i + 1] for i in range(x.shape[0])]
+            data = op(*slices)
     else:
-        slices = [x.data[i : i + 1] for i in range(x.shape[0])]
-        data = op(*slices)
+        data = x.data
 
     coords = dict(x.coords.items())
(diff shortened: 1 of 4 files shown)


Kirill888:
@emmaai can I have more context as to what the problem is? What's with all the `MultiIndex` things? When are things broken and how exactly? Can I have an example of failure?

libs/algo/odc/algo/_masking.py
```diff
@@ -689,23 +689,40 @@ def _fuse_with_custom_op(x: xr.DataArray, op, name="fuse") -> xr.DataArray:
...
+        return res
+    else:
+        return res.set_xindex(indexes)
 
 
 def _xr_fuse(xx, op, name):
     if isinstance(xx, xr.Dataset):
         return xx.map(partial(_xr_fuse, op=op, name=name))
-    if xx.shape[0] <= 1:
-        return xx
```
Kirill888, line 708: I'm not sure this is safe to remove

Kirill888:
I'm not sure we even need `MultiIndex` here https://github.com/opendatacube/odc-tools/blob/c4b509cefab178b15a960e0735ab21b02ca5b21d/libs/algo/odc/algo/_grouper.py#L90

This was just me not understanding that you can have multiple coordinates linked to the same dimension.

But, whatever makes this work.

Kirill888 (approved):

emmaai:
> @emmaai can I have more context as to what the problem is? What's with all the `MultiIndex` things? When are things broken and how exactly? Can I have an example of failure?

The issue is here, as well as info related
https://github.com/pydata/xarray/issues/6836

libs/algo/odc/algo/_masking.py
emmaai: Moved here to get a consistent index on dims
https://github.com/opendatacube/odc-tools/blob/f2797912e816667200c8b8f17026baf695ab3ce5/libs/algo/odc/algo/_masking.py#L692-L699

PR #543, opened 2023-02-06
Archive if action is provided in message attributes

### LINKED ISSUE #542: sqs-to-dc decide to archive or index based on SNS message attributes
Currently sqs-to-dc is told whether to archive by giving a `--archive` flag. It would be nifty if it was capable of deciding whether to archive based on the message attributes in the SNS messages it's consuming from.

It looks like it's already parsing the message attributes, perhaps it can check whether the message attribute contains an archived action (i.e., `{"action": "ARCHIVED"}`):
https://github.com/opendatacube/odc-tools/blob/c4b509cefab178b15a960e0735ab21b02ca5b21d/apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py#L234-L239

Then, in my particular use case, I wouldn't need two processes (I currently have two KEDA objects: one for indexing, and another for archiving).

Ariana-B:
Currently, the only way to archive a dataset using sqs_to_dc is by providing the `--archive` flag. In order to simplify pipelines that need to both archive and add datasets, check for `action: ARCHIVED` in the message attributes as well.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
@@ -61,6 +61,13 @@ def extract_metadata_from_message(message):
         raise IndexingException("Failed to load metadata from the SQS message")
 
 
+def extract_action_from_message(message):
+    attributes = message.message_attributes
+    if attributes:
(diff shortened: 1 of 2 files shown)


jmettes (approved):
Looks good! Thanks @Ariana-B!

Issue #550, opened 2023-02-28
Excessive `datacube_index` metrics created via StatsD


omad:
When `sqs-to-dc`,  and other indexing tools from this repo report metrics to StatsD, they are creating excessive numbers of unique metric labels. Time series databases like Prometheus scale poorly when given too many unique metrics to track.

The specific unique label in question is recording the hostname, which would more appropriately be recorded in a log, not in a metric.

This picture shows 35k unique metrics for `datacube_index`  from Digital Earth Australia. 

![image](https://user-images.githubusercontent.com/108979/221725605-0fbed7ac-e023-4834-bb31-865b704bd41f.png)

omad:
Fixed by #552

PR #551, opened 2023-02-28
Fix excessive numbers of indexing metrics being created

omad:
See https://github.com/opendatacube/odc-tools/issues/550


pjonsson:
@omad Did #552 solve the problem this was meant for, so this PR can be closed?

PR #552, opened 2023-02-28
Fix metrics cardinality explosion

benjimin:
Metrics are not logs. Pods are short lived. If we ask statsd to keep tracking a separate set of metrics for every unique pod name it has ever encountered, then we will accumulate tens of thousands of separate metrics over the course of a week, all of which being actively exported to prometheus. This impacts the performance and cost of monitoring.


omad (approved):
Thanks Ben!

omad:
There's one **known bad test** that's failing, due to relying on remote network resources. 🤦

I'm going to bypass branch protections and merge.

PR #554, opened 2023-03-01
Hack in multipolygon support using convex_hull

alexgleith:
ODC Tools currently supports indexing [STAC Items](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md) into an ODC Database. It does this by converting them into the native eo3 dataset format used by ODC. Unfortunately, ODC only supports geometry stored as a single Polygon.

This PR adds support for indexing MultiPolygon STAC Items by using a [Convex Hull](https://en.wikipedia.org/wiki/Convex_hull) operation to merge the multiple polygons into a single large polygon covering the entire footprint.

These types of STAC Items are used in the wild, including for example in the [ESRI 10m Annual Land Use Land Cover (9-class) data available on Planetary Computer](https://planetarycomputer.microsoft.com/dataset/io-lulc-9-class).

While it would be preferable for multi-polygons to be supported, until then, the best we can do is create a workaround that will still represent the valid data region appropriately.


tests/test-env-py38.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
   - python=3.8
 
   # Datacube
-  - datacube>=1.8.8
+  - datacube<=1.8.9
```
omad, line 13: Why the change to `<=`? Seems a bit strange to me. I'd expect a minimum bound at least.

omad (commented):
Hey Alex, can you please provide a better description of what this change is, and what it's useful for.

Something like (please update if I've made any mistakes):

> 
> ODC Tools currently supports indexing [STAC Items](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md) into an ODC Database. It does this by converting them into the native eo3 dataset format used by ODC. Unfortunately, ODC only supports geometry stored as a single Polygon.
> 
> This PR adds support for indexing MultiPolygon STAC Items by using a [Convex Hull](https://en.wikipedia.org/wiki/Convex_hull) operation to merge the multiple polygons into a single large polygon covering the entire footprint.
> 
> These types of STAC Items are used in the wild, including for example in the [ESRI 10m Annual Land Use Land Cover (9-class) data available on Planetary Computer](https://planetarycomputer.microsoft.com/dataset/io-lulc-9-class)
>

alexgleith:
Hey @omad, you're right, sorry. I was rushing and didn't complete the description.

Your text was perfect, as usual!

tests/test-env-py38.yml
alexgleith: Oh, because of this: https://github.com/opendatacube/datacube-core/issues/1412

omad: Ooooh, okay! I think I hit that today and didn't even realise it, I'll have to investigate further. Thanks

alexgleith: Paul is aware of it too.

omad (approved):
Looks good to me! Thanks Alex!

PR #556, opened 2023-03-02
Mark dataset as skipped instead of failed if more mature version exists

Ariana-B:
Currently, trying to index a less mature dataset logs an error and is marked as failed. However, this is expected behaviour, so the dataset should be marked as skipped instead, and there is no need for a loud exception.


JM-GA (approved):
Thank you!

Issue #548, opened 2023-02-23
s3-to-dc --stac fails if asset is jp2 and crs != "epsg:4326"


joelsundholm:
I have some Sentinel-2 MSI Level-1C data in an S3 bucket that I want to index using STAC.  This data is just the unzipped archives that were downloaded from ESA and are otherwise unaltered. The image files are in jp2 format and the epsg code corresponds to whatever UTM zone the data is in.  

This fails because the code doesn't attempt to extract the crs for anything other than geotiffs, see here:

https://github.com/opendatacube/odc-tools/blob/e5e987f7bb55a794f60308b3ce85ce0d29930a9a/apps/dc_tools/odc/apps/dc_tools/_stac.py#L218

... and eventually we run into exceptions here where we try to fetch some default transform that we didn't place in the dict.
https://github.com/opendatacube/odc-tools/blob/e5e987f7bb55a794f60308b3ce85ce0d29930a9a/apps/dc_tools/odc/apps/dc_tools/_stac.py#L385

Is there any reason why it shouldn't be possible to index the jp2 files?

alexgleith:
It is possible, it's just laziness on the original implementer's behalf (mine).

You're welcome to raise a PR that allows for jp2 files :-)

PR #557, opened 2023-03-08
Add jp2 to set of supported STAC asset types

joelsundholm:
The purpose of this PR is to solve https://github.com/opendatacube/odc-tools/issues/548


omad (approved):
Makes a lot of sense to me. Thank you!

PR #562, opened 2023-05-04
Abort when doc stream empty

joelsundholm:
This commit fixes that s3-to-dc waits indefinitely whenever a glob fails to produce any documents.

Instead we raise a click.Abort exception which properly ends execution.


omad (approved):
Brilliant! Thank you.

PR #579, opened 2023-10-09
update archive_less_mature option to expect int value

Ariana-B:
To allow for timestamp discrepancies between dataset versions, archive_less_mature in core now expects an optional int value instead of a boolean. Update the `--archive-less-mature` click option and `index_update_dataset` to reflect this.


SpacemanPaul (approved):
Thanks for dealing with this.

PR #584, opened 2023-10-27
Enable list of absolute urls in s3-to-dc

Ariana-B:
- Make it so that s3-to-dc no longer requires a glob for the uri, enable passing multiple uris, and make product argument optional. 
- More informative error for empty doc stream by using `click.ClickException` instead of `click.Abort`, which doesn't support a custom error message. 
- Rename test environment to `odc-tools-tests`


benjimin (approved):
This (not needing to fake globs, not needing to specify product, keeping python current) will be helpful, thanks!

tests/test-env.yml
```diff
@@ -48,7 +48,7 @@ dependencies:
   - sphinx-autodoc-typehints
   - nbsphinx
 
-  - pip=20
+  - pip=23
   - pip:
       # odc.apps.dc-tools
       - thredds-crawler
```
omad, line 54: Don't change it in this PR, but all of `thredds-crawler`, `rio-stac` and `jupyter-ui-poll` are available from _conda-forge_, so should be installed from there.

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -130,7 +130,15 @@ def cli(
...
     product,
 ):
-    """Iterate through files in an S3 bucket and add them to datacube"""
+    """
+    Iterate through files in an S3 bucket and add them to datacube.
+
+    File uris can be provided as a glob, or as a list of absolute URLs.
+    If more than one uri is given, all will be treated as absolute URLs.
+
+    Product is optional; if one is provided, it must match all datasets.
```
omad, line 139: This documentation isn't right (as far as I can see).

Product is optional, can be a single product, or can be a space separated list of multiple products.

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -110,8 +110,8 @@ def dump_to_odc(
 @request_payer
 @archive_less_mature
 @publish_action
-@click.argument("uri", type=str, nargs=1)
-@click.argument("product", type=str, nargs=1)
+@click.argument("uri", type=str, nargs=-1)
```
omad: This `type=str` is ignored, which makes it confusing. I've tested a simple mock script here, and by specifying `nargs=-1`, `uri` will always be a `tuple`.

So, I think remove `type=str`, and rename `uri` to `uris`.

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -139,32 +147,59 @@ def cli(
...
-            f"Error: Requested Product/s {', '.join(missing_products)} {'is' if len(missing_products) == 1 else 'are'} "
-            "not present in the ODC Database",
-            file=sys.stderr,
-        )
-        sys.exit(1)
+
+    # if it's a uri, a product wasn't provided, and 'product' is actually another uri
+    if product.startswith("s3://"):
+        candidate_products = []
+        if isinstance(uri, str):
```
omad: See above, `uri` is always a `tuple` so this check can be removed.

apps/dc_tools/tests/test_s3_to_dc.py
```diff
@@ -133,3 +131,55 @@ def test_s3_to_dc_partially_succeeds_when_given_invalid_and_valid_dataset_yamls(
...
+            "--stac",
+            "s3://odc-tools-test/sentinel-s2-l2a-cogs/31/Q/GB/2020/8/S2B_31QGB_20200831_0_L2A/*_L2A.json",
+        ],
+        catch_exceptions=False,
+    )
+    assert result2.exit_code == 0
+    assert (
+        result2.output
+        == "Added 1 datasets, skipped 0 datasets and failed 0 datasets.\n"
+    )
```
omad, line 185: Hip hip array for tests!! I thought this case wasn't going to work because `product` is a greedy argument and I hadn't read all the code, but I saw this first and it gave me confidence.

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -90,7 +90,7 @@ def dump_to_odc(
         except SkippedException:
             ds_skipped += 1
     if not found_docs:
-        raise click.Abort("Doc stream was empty")
+        raise click.ClickException("Doc stream was empty")
```
omad, line 93: Nice! You should mention this fix in the PR description too.

omad (changes requested):

tests/test-env.yml
Ariana-B: Is there a reason not to change it now?

omad: In case it makes tests fail, or otherwise complicates a pleasantly straightforward PR. :)

omad (approved):
Love it! Useful improvement while maintaining backwards compatibility.

Thanks Ariana.

PR #587, opened 2023-11-09
Fix Python versions

SpacemanPaul:
Updated supported Python version to >= 3.9 throughout.  
Updated test conda environment to Python 3.9
Pinned odc_apps_dc_tools to Python 3.9-3.10 as per https://github.com/opendatacube/datacube-core/issues/1511

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -103,7 +103,7 @@ jobs:
       shell: bash -l {0}
       if: steps.conda_cache.outputs.cache-hit != 'true'
       run: |
-        mamba env create -f tests/test-env-py38.yml -p tests/env
+        mamba env create -f tests/test-env-py39.yml -p tests/env
 
     - name: Check Python Env
       shell: bash -l {0}
@@ -128,7 +128,7 @@ jobs:
         path: |
           tests/env
 
-        key: ${{ runner.os }}-test-env-py38-${{ hashFiles('tests/test-env-py38.yml') }}
+        key: ${{ runner.os }}-test-env-py39-${{ hashFiles('tests/test-env-py39.yml') }}
 
     - name: Update PATH
       shell: bash
@@ -211,7 +211,7 @@ jobs:
         path: |
           tests/env
 
-        key: ${{ runner.os }}-test-env-py38-${{ hashFiles('tests/test-env-py38.yml') }}
+        key: ${{ runner.os }}-test-env-py39-${{ hashFiles('tests/test-env-py39.yml') }}
 
     - name: Update PATH
       shell: bash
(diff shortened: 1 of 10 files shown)


tests/test-env-py39.yml
```diff
@@ -46,7 +46,7 @@ dependencies:
   - pytest-httpserver
   - pytest-cov
   - pytest-timeout
-  - moto
+  - moto[server]
```
Kirill888: not valid conda syntax, this only works with `pip`, unless something changed recently in conda

SpacemanPaul: Yeah, I had a feeling that was the case.  I've installed conda and I'll deal with it tomorrow.

SpacemanPaul: It's the moto version that's the issue - conda is pulling in an ancient version for some reason.

SpacemanPaul: Lol.  I just removed the ancient pin on `aiobotocore` and it built in 1/5 the time and works fine.

emmaai (approved):

apps/dc_tools/setup.cfg
```diff
@@ -16,7 +16,7 @@ url = https://github.com/opendatacube/odc-tools/
 include_package_data = true
 zip_safe = false
 packages = find_namespace:
-python_requires = >=3.7
+python_requires = >= 3.9,<=3.10
```
omad: At least make it up to and include 3.11. I'd prefer not to have an upper bound on Python version.

tests/test-env-py39.yml
```diff
@@ -49,7 +49,6 @@ dependencies:
   - moto
   - deepdiff
 
-
   # for docs
   - sphinx
```
omad, line 53: Don't do it now, but, omg, testing everywhere is such a mess. I'm sure we don't need sphinx here...

.github/workflows/main.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.8
+        python-version: 3.9
```
omad, line 31: At some point, we should use the GitHub actions matrix to parameterise this and test more than a single Python version.

omad (changes requested):
Thanks Paul. Other than the <=3.10 restriction on Python this looks okay.

apps/dc_tools/setup.cfg
SpacemanPaul: Upper bound was because of issue with aiohttp reported in https://github.com/opendatacube/datacube-core/issues/1511

omad (approved):
ty

PR #595, opened 2023-12-19
When a geometry is invalid, fall back to the native transform

alexgleith:
Basically, we have a native geometry that crosses the anti-meridian when converted to lat/lon. So we split the feature there, maintaining coordinates that are between `-180` and `180`. But that then makes it really hard to put back into projected coordinates.

Rather than do that, if there's an invalid geometry, this proposed change falls back to native transform information to create a native geometry.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/_stac.py
+++ b/apps/dc_tools/odc/apps/dc_tools/_stac.py
@@ -2,17 +2,18 @@
 Tools for STAC to EO3 translation
 """
 import math
-import numpy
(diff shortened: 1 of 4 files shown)


SpacemanPaul (approved):
I can't imagine this will fix everything, but seems to be an incremental improvement at least.

alexgleith:
>I can't imagine this will fix everything

No, and all this STAC code is a bit of a mess now... whoops. But yeah, I think this is a good fallback for now. Thanks!

PR #597, opened 2024-01-16
Gather s3 operations into class

emmaai:
It's related to fixing the credentials rotation issues with service account binding assumed role on k8s. The change is required as I will also make changes on how `s3_client` is managed on dask in `odc-stats`, i.e., one client per worker process among all threads on the worker, instead of one client per thread (ref: https://github.com/opendatacube/odc-stats/pull/114). 

`boto3.client` handles the credentials refresh under the hood if we let it follow the credentials provider chain.

The reason of the changes split into two parts, here and `odc-stats`, is to manage the scope of effect. They may be moved into `core` when the community deems fit.

### DIFF
--- a/libs/cloud/odc/aws/s3_client.py
+++ b/libs/cloud/odc/aws/s3_client.py
@@ -48,9 +48,8 @@ class S3Client:
         """Open whole or part of S3 object
 
         :param url: s3://bucket/path/to/object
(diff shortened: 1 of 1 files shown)


SpacemanPaul (approved):
Looks good to me.

PR #601, opened 2024-02-21
add log level option for indexing tool

emmaai:
as the title. 
- The `WARNING` level is too spamming to catch the real failure on indexing. 
- Airflow dag would fail if the log chats too much


SpacemanPaul (approved):
LGTM.  I like the log level as click option.

emmaai:
can be sure this time that I'm not "authorised" to merge.

SpacemanPaul:
> can be sure this time that I'm not "authorised" to merge.

Well let's take another look and see if we can see why.....   try now.

Issue #588, opened 2023-11-10
ODC Tools Broken on Python 3.12 - broken urlpath dependency


omad:
The [urlpath · PyPI](https://pypi.org/project/urlpath/) package is used in ODC for a bit of URL handling. However, the package hasn't been updated in a couple of years and is broken in Python 3.12.

See: 
![image](https://github.com/opendatacube/odc-tools/assets/108979/3968ace7-162d-4a9a-bc39-58e859a83dbe)

and https://github.com/brandonschabell/urlpath/issues/6

We should just drop the dependency, doing whatever trivial rewriting is necessary.

EReaso:
Still having this issue on MacOS 15.4 with Python 3.12.3.

PR #602, opened 2024-03-04
Remove urlpath dependency and update relative path logic

Ariana-B:
- Remove broken urlpath dependency as per #588 
- Change relative path logic to be consistent across tools and be default behaviour, notably removing `--rewrite-assets` option from `stac_to_dc`


apps/dc_tools/odc/apps/dc_tools/_stac.py
```diff
@@ -179,10 +179,23 @@ def _find_self_href(item: Document) -> str:
     return self_uri[0]
 
 
+def _get_relative_path(asset_href, self_link):
+    if self_link is None:
+        return asset_href
+
+    self_path = urlparse(self_link).path
+    href_path = urlparse(asset_href).path
```
SpacemanPaul, line 187: It seems unlikely, but for completeness, I think we should be explicitly checking that both `self_link` and `asset_href` have the same `scheme` and `netloc` (and keeping as an absolute link if they differ).

SpacemanPaul (approved):
Approved with minor comment.

(Oh, and bump the version number while you're at it, and I'll do a release once merged.)

apps/dc_tools/odc/apps/dc_tools/_stac.py
Ariana-B: Maybe we could refrain from checking `scheme`, but check for some degree of similarity between the `netloc` values?

PR #608, opened 2024-05-15
Replace deprecated `urlpath` package with standard library URL manipulation APIs.

SpacemanPaul:
The stac implementation uses the external `urlpath` package which is orphaned and only supported to Python 3.10.

This PR replaces it with `urlparse` and `posixpath` from the standard library.


Ariana-B:
I previously did something similar in #602

PR #609, opened 2024-05-16
Attempt to fix pypi/publish action.  Not clear how it came to be broken.

SpacemanPaul:
The pypi/publish githib action is broken.

Script expected files like `odc-io*.tar.gz` and `odc_io*.whl`  - the tar.gz with a dash and the whl with an underscore, but both are created with an underscore.

None of the relevant scripts have changed for years, it is not clear to me how this came to be broken, but I think this might fix it.


Ariana-B (approved):
Worth a shot!

PR #613, opened 2024-07-22
Add docker to test requirements.

SpacemanPaul:
dc-tools failing because docker can't be imported.


SpacemanPaul:
This is not the right approach.

PR #616, opened 2024-11-18
Do not hang on empty stac file

pjonsson:
Check that there is a document
returned from parse_doc_stream,
so a zero byte file does not
crash s3_to_dc.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
@@ -62,6 +62,9 @@ def dump_to_odc(
 
     found_docs = False
     for uri, metadata in uris_docs:
+        if metadata is None:
+            ds_skipped += 1
+            continue
         found_docs = True
         stac_doc = None
         if transform:


pjonsson:
The CI failure on pre-commit.ci has been present for several months and is not something this PR introduced.

Last release of odc-apps-dc-tools was in July, would it be possible to get a new release after this PR is merged?

pjonsson:
@alexgleith I can't figure out the "code owners" here but I see you have recent commits in the repository, is this something you can review?

alexgleith (approved):
Looks fine to me

pjonsson:
@alexgleith can you evoke a new PyPi-release of the package this code belongs to?

alexgleith:
We need to increment the version right, @pjonsson?

pjonsson:
@alexgleith I have no idea how the PyPi releases work, but I guess we (well, or at least I) would like to have a 0.2.19 version that includes this fix.

alexgleith:
We need to increment the version file, here: https://github.com/opendatacube/odc-tools/blob/develop/apps/dc_tools/odc/apps/dc_tools/_version.py

Then run the script on a local machine that pushes a branch up.

alexgleith:
Ok, should be building now: https://github.com/opendatacube/odc-tools/actions/runs/12019501195

pjonsson:
Oh, that is nice. I don't know anything about PyPi or how these releases are done, but it looks like you hit what #621 is trying to fix.

omad:
There's a bunch of brokenness in the GitHub Actions builds!

I was working on it last week, and fixed a few things, but got stuck on the copernicus dem to ODC tool tests passing locally but not in GitHub Actions.

Back on it now...

alexgleith:
>but got stuck on the copernicus dem to ODC tool 

We should probably just kill it. There's simple ways to index that data from STAC now, which means one less tool to maintain.

omad:
This is the bug https://github.com/opendatacube/odc-tools/issues/622
`cop-dem-to-dc` and `esa-wc-to-dc`.

alexgleith:
>cop-dem-to-dc and esa-wc-to-dc

Both are available via a reasonably well known [STAC Catalog (Microsoft)](https://planetarycomputer.microsoft.com/catalog), so I still think we should remove them both.

PR #621, opened 2024-11-21
Fix GHA: PG admin tools have moved

omad:


pjonsson:
@omad the second commit seems to fix issues with running/accessing postgres. Do you think there are more issues with postgres, or are the remaining issues you have discovered related to other things?

If the remaining issues are not related to postgres, this PR is a strict improvement over main and you might as well merge it. I understand there will still be test failures on the main branch after merging, but regular test failures are easier to debug for most people. (If you get time later to fix those issues that is great, but I understand you no longer work with opendatacube and I know how hard it is to do this kind of stuff on weekends and evenings.)

omad:
Hey @pjonsson , the remaining issues are described in #622 and isn't related to Postgres.

I changed jobs recently, from Geoscience Australia to CSIRO, but ODC is still part of my responsibilities there too. :)

PR #630, opened 2025-03-12
Restore missing dependency

pjonsson:
The s3-to-dc command does
not start without having
importlib_resources available,
so restore this dependency
that was removed in #625.

### DIFF
--- a/apps/dc_tools/setup.cfg
+++ b/apps/dc_tools/setup.cfg
@@ -40,6 +40,7 @@ install_requires =
     urlpath
     datadog
     eodatasets3>=1.9
+    importlib_resources>=6.0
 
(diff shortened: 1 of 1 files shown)


omad:
Thanks for pointing this out! I'd prefer to solve it another way though, I'll raise a PR.

We shouldn't be importing `importlib_resources` anyway, since Python 3.7 [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html) is available in the standard library and we should use that. We don't support earlier Python versions, but if we did, we could add `importlib_resources` as a dependency just for those Python versions.

Digging further, that import is only used in one function, and that function is never called. So I'm going to remove the dead code to fix it instead.

PR #631, opened 2025-03-13
Fix missing import - remove dead code

omad:
@pjonsson pointed out in PR #630 that s3-to-dc is broken due to `importlib_resources` not being listed as a dependency. 

Upon investigation, it's not even being used, so lets delete the import and dead code instead.

We shouldn't be importing `importlib_resources` anyway, since Python 3.7 [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html) is available in the standard library and we should use that. We don't support earlier Python versions, but if we did, we could add `importlib_resources` as a dependency just for those Python versions.

Digging further, that import is only used in one function, and that function is never called. So I'm going to remove the dead code to fix it instead.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/utils.py
+++ b/apps/dc_tools/odc/apps/dc_tools/utils.py
@@ -2,7 +2,6 @@ import logging
 from typing import Optional
 
 import click
-import importlib_resources
 from datacube import Datacube
 from datacube.model import Dataset
 from datacube.index.hl import Doc2Dataset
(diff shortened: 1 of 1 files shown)


apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -195,7 +182,7 @@ def index_update_dataset(
     auto_add_lineage: Optional[bool] = False,
     publish_action: Optional[str] = None,
     stac_doc: Optional[dict] = None,
-) -> int:
+):
```
pjonsson, line 185: @omad thanks for fixing this. In this case the type annotation was wrong, but for people who aren't familiar with the code base it's good to have return type annotations to inform what the function is expected to return, so it's nice if you replace faulty return types with the right ones in the future.

PR #632, opened 2025-03-13
test-env: remove importlib_resources

pjonsson:
Commit 236714099d removed the use
of this package, so remove it from
the test-env as well.


omad:
Thank you! I didn't even realise this was here.

PR #647, opened 2025-06-23
Add a URL replace argument for STAC API to DC

alexgleith:
This re-adds some functionality that I swear I added a long time ago, and must have been removed.

In short, there is sometimes a need to rewrite asset URLs at indexing time. An example is for USGS data.

I have tested manually with the below command, which works. And written a test, which I'm hoping will pass in CI, because I didn't get it working.

``` bash
stac-to-dc --catalog-href='https://landsatlook.usgs.gov/stac-server/' \
                    --bbox='114,-9,116,-7' \
                    --collections='landsat-c2l2-st' \
                    --datetime='2025-01-01/2025-05-30' \
                    --rename-product='ls9_c2l2_st' \
                    --url-string-replace='https://landsatlook.usgs.gov/data,s3://usgs-landsat' \
                    --limit=100 \
                    --options="query={\"platform\":{\"in\":[\"LANDSAT_9\"]}}"
```

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py
@@ -66,6 +66,7 @@ def item_to_meta_uri(
     item: Item,
     dc: Datacube,
     rename_product: Optional[str] = None,
+    url_string_replace: tuple[str, str] | None = None,
 ) -> Generator[Tuple[Dataset, str, bool], None, None]:
     for link in item.links:
         if link.rel == "self":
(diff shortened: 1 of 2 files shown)


SpacemanPaul (approved):
We have STAC->EO3 conversion code in two places - here and in eodatasets. Perhaps some variant of this functionality exists there and is what you are remembering.

In the longer term we need to consolidate the code in these two repos.

alexgleith:
>We have STAC->EO3 conversion code in two places - here and in eodatasets. Perhaps some variant of this functionality exists there and is what you are remembering.

This is actually using the `eo3` [code now](https://github.com/opendatacube/odc-tools/blob/b1a12fa3adf90aa5e9f565fa405d2881e1baea01/apps/dc_tools/odc/apps/dc_tools/stac_api_to_dc.py#L13).

(Conversion code is available in odc-stac and eo3 now, as well as here, so it's in 3 places, but hopefully only used from 2!)

Ariana-B:
Just for the record - @alexgleith remembers correctly, and it was me who removed it in #602 as part of updates to relative path logic. I'm sure there was more discussion/context behind the decision, though I no longer recall - or perhaps we were simply unaware of this particular use case.

alexgleith:
Thanks, @Ariana-B. Looks like a complicated PR, which you executed smoothly!

This url replacer is a much simpler implementation, so hopefully easy to use and to predict the impact of.

PR #651, opened 2025-07-10
Fix single url indexing

alexgleith:
The simple case of passing in URLs was not working with the USGS S3 URLs, which need requester-pays.

I used a very simple implementation of the equivalent of the S3Fetcher. Clearly it's not optimised, but I don't think it needs to be, as a URL list should be fairly short.

This PR also aligns the `s3-to-dc` tool to the `stac-to-dc` tool, sharing code and arguments between them.

Tests updated to handle the new, more flexible and less hard-coded way of renaming products.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
@@ -2,17 +2,19 @@
 """Build S3 iterators using odc-tools
 and index datasets found into RDS
 """
-import click
+
 import logging
 import sys
-from odc.aio import S3Fetcher, s3_find_glob
-from typing import Tuple
+from types import SimpleNamespace
+from typing import Any, Dict, Tuple
 
+import botocore
+import click
 from datacube import Datacube
 from datacube.index.hl import Doc2Dataset
 from datacube.ui.click import environment_option, pass_config
+from odc.aio import S3Fetcher, s3_find_glob
 from odc.apps.dc_tools._docs import parse_doc_stream
-from odc.apps.dc_tools._stac import stac_transform
 from odc.apps.dc_tools.utils import (
     IndexingException,
     SkippedException,
@@ -20,7 +22,10 @@ from odc.apps.dc_tools.utils import (
     archive_less_mature,
     fail_on_missing_lineage,
     index_update_dataset,
+    item_to_meta_uri,
     no_sign_request,
+    publish_action,
+    rename_product,
     request_payer,
     skip_check,
     skip_lineage,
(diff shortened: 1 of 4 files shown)


Ariana-B:
I'm not sure I fully understand what wasn't working in the first place. Is there a test case you could add?

Also, it seems the failing test case isn't compatible with `stac2ds` as it requires this logic from `stac_transform`: https://github.com/opendatacube/odc-tools/blob/677738c0530aac0209f33889a0813e5f821be484/apps/dc_tools/odc/apps/dc_tools/_stac.py#L118T-L125
The comments denote this as hacky, but I'm not sure what a proper fix looks like.

alexgleith:
>I'm not sure I fully understand what wasn't working in the first place.

I don't think it was working at all... I'll see if I can do some work on the tests.

alexgleith:
> I don't think it was working at all...

Maybe I was wrong here, and it's about requester-pays buckets not working. Investigating more.

alexgleith:
I don't think it's possible to mock an S3 Requester Pays bucket.

These changes make things significantly more consistent between the S3 and STAC API indexing. I think it's better generally!

pjonsson:
@alexgleith I had a 2 minute look and this wasn't obvious to me. How urgent is this for you, can I take a closer look early next week?

alexgleith:
Not urgent. I'm doing some testing of it still.

alexgleith:
I feel like this is ready to merge. It's a significant change, so needs some care, but I think the changes are needed.

Any thoughts @SpacemanPaul, @omad and @Ariana-B?

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -29,16 +34,73 @@
...
+                SimpleNamespace: For each input, yields an object with:
+                    - url (str): The S3 URL.
+                    - data (bytes or None): The fetched data, or None on failure.
+                    - last_modified (datetime, optional): Last modified timestamp (if available).
+                    - range (tuple or None): The requested range, if any.
+                    - error (Exception or None): Error information, if any.
+            Notes:
+                - The order of results is not guaranteed to match the input order.
+                - One result is yielded for each input URI.
+    """
```
SpacemanPaul, line 70: The docstring here doesn't seem to match the code: 
 - mentions taking "a sequence of S3 URLs (or (url, range) tuples)" but appears to only take a sequence of urls.
 - SimpleNameSpace is not being returned with last_modified, range, or error.
 - says it loads uris concurrently but appears to be serialised.
 
 Would also be nice to properly typehint this new class.

apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -293,3 +301,52 @@ def statsd_gauge_reporting(value, tags=None, statsd_setting="localhost:8125") ->
...
 
     statsd.gauge("datacube_index", value, tags=tags)
+
+
+def item_to_meta_uri(
+    item: Item,
+    dc: Datacube,
+    rename_product: Optional[str] = None,
+    url_string_replace: tuple[str, str] | None = None,
+) -> Generator[Tuple[Dataset, str, bool], None, None]:  # noqa: F821
```
SpacemanPaul: The typehints here seem totally out of whack with the code.  This function has a `return` statement, not a `yield` statement, I'm not sure how it can be said to return a generator?

Isn't it just `-> Tuple[Dataset, str, dict]`?     (Note third element of the tuple is a dict - technically `dict[str, Any]` - not a bool.)

I know you are just moving the method from one module to another, but may as well clean up the type hints at the same time.

SpacemanPaul (changes requested):
I'm basically happy with the change, but the existing code is badly commented and typehinted, and we should be cleaning this up as we refactor, not perpetuating it.

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
alexgleith: Ok, on it. Fixed.

apps/dc_tools/odc/apps/dc_tools/utils.py
alexgleith: Good point. Thanks!

apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -29,16 +34,75 @@
...
+from odc.aws import _aws_unsigned_check_env, auto_find_region, s3_client, s3_fetch
+from pystac import Item
 
 
 def doc_error(uri, doc) -> None:
     """Log the internal errors parsing docs"""
     logging.exception("Failed to parse doc at %s", uri)
 
 
+class SimpleFetcher:
```
Ariana-B, line 49: Should this perhaps live in `odc.aio` alongside `S3Fetcher`?

alexgleith: I really think it should NOT!

`odc.aio` is really fragile, and this Class is not at all async...

I guess we could make it async, but I'd do it using [obstore](https://github.com/developmentseed/obstore) and I reckon it's a separate PR.

(I'd be inclined to replace all the custom `odc.aio` code with obstore, which is probably faster already.)

Ariana-B: Ok, I didn't realise it would have to be made async to be put there.

alexgleith: Maybe it's just me being pedantic! But that's what the `aio` stands for :)

SpacemanPaul: It's not an asynchronous implementation (although perhaps it should be).

SpacemanPaul (approved):
I tightened the typehint of the last element returned by `item_to_meta_uri` to that return type of `Item.to_dict()`

I think we're good to go.

PR #653, opened 2025-07-16
STAC/EO3 refactor: import stac2ds from core instead of odc-stac.

SpacemanPaul:
- remove direct dependency on odc-stac.
- require `datacube>=1.9.6`
- import `stac2ds` from core instead of odc-stac
- Bump version number for release.

**See Also**
- https://github.com/opendatacube/datacube-core/pull/2027
- https://github.com/opendatacube/odc-stac/issues/219


alexgleith (approved):
Looks nice and straightforward!

omad:
Updated description with links to the odc-stac PR and the datacube-core PR that necessitated this change.

PR #655, opened 2025-07-23
Remove urlpath from requirements

Ariana-B:
Although `urlpath` is no longer used in dc-tools, it is still present in the requirements - remove it.


omad (approved):

omad:
Might fix #588 (again).

Issue #331, opened 2021-08-25
Proposal: reduce number of top level packages


Kirill888:
Problem
======

This repository grew "organically" for a while and some of the earlier experiments have proven to be less relevant than other. The side effect of this growth is a large number of packages and namespaces. Currently there is a one to one mapping between namespace and package, like `odc.algo.*` is shipped by `odc-algo`, and `odc-algo` ships only files in `odc/algo`. While this is a reasonable and clean relation it does make for a larger number of packages. On one hand this allows for higher granularity when declaring dependencies, on the other, packages are not "free". It adds to CI delays, makes renaming and moving code around  harder, and makes publishing/managing to pypi/conda harder too as more secrets need to be managed, maintainers need to be added to more projects etc. (I spent some time adding @GypsyBojangles as an owner to every pypi project pushed from this repo, and will need to do a similar thing for generating publishing tokens.)

Stocktake
=======

First let's decide which thing we definitely keeping as is. I'd say that apps can stay as they are. And as far as user facing libraries go I have this list:

- `odc.algo` - mostly xarray + dask tool: example `xr_rerpoject`
- `odc.ui` - tools for visualizations in jupyter
- `odc.stac` -  (previously `odc.index`) STAC and missing datacube index utilities
- `odc.stats` - large scale data processing libs/apps (work in progress)
- `odc.dscache` - used by `odc.stats`, but has other possible future use cases (odc index export/import for example)

Then we have "cloud io helper libs" that are mostly used via apps or other higher level libs and not directly by users.

- `odc.aws` - AWS S3 and SQS
- `odc.azure` - Azure blob storage
- `odc.thredds`- Crawling THREDDS
- `odc.aio` - AWS S3, but async, has annoying `aiobotocore` dependency

The reason why these are all separate is due to dependencies they pull in. For example `odc.aio` depends on `aiobotocore` which is really challenging to install in the presence of dependencies on `boto*` libraries. One option is to put them all into one package, say `odc-cloud`, and use feature flags to enable/disable features, so instead of depending on `odc-azure` one would declare dependency on `odc-cloud[AZURE]`.

Finally there is an odd bunch of libs

- `odc.io` - poor name, used by cli apps, for various unrelated text processing helpers
- `odc.ppt` - "paralllell processing tools", some generic "Future" object handling(not used) and `Async->Thread` adapter
   - only used by `odc.aio`
[...]

PR #663, opened 2025-08-14
Merge odc-apps-cloud into odc-apps-dc-tools

pjonsson:
There used to be a requirement that cloud
did not depend on datacube, but that
requirement has not been present for
the past 4+ years, so merge all apps
into a single package.

Refs #331


pjonsson:
I'm 4 years late, but my impression is that there was no disagreement about taking this step 4 years ago, it was just never done.

alexgleith (approved):
Love a bit of consolidation and simplification!

Thanks for the contribution.

PR #664, opened 2025-08-15
convert-bools flag for s3-to-dc

Ariana-B:
Add a `convert-bools` flag to `s3-to-dc` to convert boolean values to strings as a temporary backwards-compatibility for data that needs to be indexed into a 1.8 db (which does not support boolean field types)


apps/dc_tools/odc/apps/dc_tools/s3_to_dc.py
```diff
@@ -115,6 +115,7 @@ def dump_to_odc(
     publish_action=None,
     rename_product: None | str = None,
     url_string_replace: None | tuple[str, str] | None = None,
+    convert_bools=False,
```
pjonsson: Can this get a type annotation?

pjonsson:
>.. for data that needs to be indexed into a 1.8 db (which does not support boolean field types)

The dependencies for this package specifies `datacube >= 1.9.6`, does that work with a 1.8 db?

Ariana-B:
> The dependencies for this package specifies `datacube >= 1.9.6`, does that work with a 1.8 db?

This is a workaround for a specific internal use case in which data is still being indexed into a 1.8 datacube for the time being but is then used by a 1.9 explorer.

caitlinadams (approved):
This looks like it should fix our issue, thanks for writing up this work-around!

SpacemanPaul:
> > .. for data that needs to be indexed into a 1.8 db (which does not support boolean field types)
> 
> The dependencies for this package specifies `datacube >= 1.9.6`, does that work with a 1.8 db?

Ariana means here "a 1.8-compatible postgres database".  I.e. a postgres database where we can't use the 1.9-only support for boolean types.

SpacemanPaul (approved):
Ideally we would add this option to the other tools too, but given this is already a temporary compatibility hack for a specific internal use case, let's run with this. We can expand later if necessary.

Can you please just bump the dc-tools version number ready for a release in this PR?  Happy to merge after that.

PR #666, opened 2025-09-02
Use function from datacube

pjonsson:
The parse_doc_stream function
has been integrated into datacube
with the 1.9.9 release.


pjonsson:
Conda can't find Datacube 1.9.9 for some reason. If someone can fix that, I believe the PR could just be retriggered and it should pass the tests.

omad (approved):

PR #670, opened 2025-09-09
STAC/EO3 refactor: replace stac conversion methods with those from core

Ariana-B:
Replace `stac_transform` and `to_stac_item` with `stac2ds` and `ds2stac` from core. Also mark `index-from-tar` command as deprecated.


apps/dc_tools/odc/apps/dc_tools/_stac.py
```diff
@@ -8,10 +8,9 @@
 from uuid import UUID
 
 import numpy
+from datacube.metadata import ds2stac
```
pjonsson, line 11: Is there anything left in the package that depends on eodatasets3, or should that be removed from its dependencies?

apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
```diff
@@ -58,7 +59,7 @@ def process_doc(
         if rename_product is not None:
             # This possibly should be possible for yaml loading too
             doc["properties"]["odc:product"] = rename_product
-        doc = stac_transform(doc)
+        doc = next(stac2ds([pystac.Item.from_dict(doc)]))
```
pjonsson, line 62: Is there anything left that uses `stac_transform`, or can that function be removed as well?

Ariana-B: `index-from-tar` still uses it (didn't seem worth figuring out how to replace it when the command is getting deprecated).

apps/dc_tools/odc/apps/dc_tools/sqs_to_dc.py
```diff
@@ -181,15 +183,17 @@ def get_uri(metadata, rel_value):
     return uri
 
 
-def do_archiving(metadata, dc: Datacube, publish_action):
+def do_archiving(metadata, dc: Datacube, publish_action, stac):
```
pjonsson: What is the type of `stac`, bool? Can the parameter get that type annotation?

apps/dc_tools/pyproject.toml
pjonsson, line 1: If you do a `git grep eodatasets3` you will find a few more places that needs updating.

omad (approved):

PR #673, opened 2025-09-11
inventory: do not create 1000 threads

pjonsson:
Use the n_threads parameter instead
of creating 1000 threads.


alexgleith (approved):
Haha, this is a nice silly bug!

PR #674, opened 2025-09-11
Make exception names PEP-8 compliant

pjonsson:
This is the PEP-8 naming convention,
so follow it.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
+++ b/apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
@@ -16,7 +16,7 @@ from datacube.index.hl import Doc2Dataset
 from datacube.metadata import stac2ds
 from datacube.ui.click import environment_option, pass_config
 from odc.apps.dc_tools.utils import (
-    SkippedException,
+    DatasetExists,
     allow_unsafe,
     archive_less_mature,
     fail_on_missing_lineage,
@@ -124,7 +124,7 @@ def dump_list_to_odc(
                 success += 1
                 if success % 10 == 0:
                     sys.stdout.write(f"\rAdded {success} datasets...")
-            except SkippedException:
+            except DatasetExists:
                 skipped += 1
             except Exception as e:  # pylint:disable=broad-except
                 logging.exception("Failed to handle item %s with exception %s", item, e)
--- a/apps/dc_tools/odc/apps/dc_tools/utils.py
+++ b/apps/dc_tools/odc/apps/dc_tools/utils.py
@@ -16,15 +16,15 @@ from ._stac import ds_to_stac
 MICROSOFT_PC_STAC_URI = "https://planetarycomputer.microsoft.com/api/stac/v1/"
 
 
-class IndexingException(Exception):
+class IndexingError(Exception):
     """
     Exception to raise for error during SQS to DC indexing/archiving
     """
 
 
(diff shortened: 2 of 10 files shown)


apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
```diff
@@ -124,7 +124,7 @@ def dump_list_to_odc(
                 success += 1
                 if success % 10 == 0:
                     sys.stdout.write(f"\rAdded {success} datasets...")
-            except SkippedException:
+            except SkippedError:
```
omad: We should log the `SkippedException`/whatever it gets called, and the reason it was raised. Ideally it should include both the message, and the dataset/url that was skipped, but I'm not sure we have all that information available.

Same with the other places this case it caught.

It's pretty annoying when things don't happen how you intended, and there's no information about why.

omad (changes requested):
The `SkippedException` isn't an error case, so shouldn't be named one, it's used here as a signal to shortcut the normal return chain. Not sure it's a good idea or not, but, it's working for now. It could be renamed to something like `SkippedDataset`.

`IndexingException` does mostly appear to be used for error cases, so that rename probably makes things a bit clearer.

apps/dc_tools/odc/apps/dc_tools/utils.py
```diff
@@ -340,7 +336,7 @@ def item_to_meta_uri(
             "Couldn't find matching product for product name: %s",
             product_name_sanitised,
         )
-        raise SkippedException(
+        raise DatsetExists(
```
pjonsson: This is really a mis-use of the documented behavior, but there is a warning emitted in the previous line.

apps/dc_tools/odc/apps/dc_tools/azure_to_dc.py
pjonsson: I have renamed `SkippedException` to `DatasetExists`.

The already existing doc string for the exception details when datasets are skipped (when they already exist and not in updating mode). There is one mis-use of this exception where it is used for something else and I marked that in this PR, but that mis-use is preceded by a line that logs a warning like you want.

I'm not suggesting that the user should need to check the source code and read the doc string to figure out why datasets are skipped, but that is a documentation or UX issue, this PR is trying to give the classes PEP-8 compliant names.

I think it was 300 000+ lines of logs from our daily indexing jobs the last time I checked, getting another 500 000+ lines of "Dataset XYZ already exists, skipping.." added on top of that is not workable for us.

omad:
The failure message from `fs-to-dc` in the tests is:

    "Usage: fs-to-dc [OPTIONS] INPUT_DIRECTORY\nTry 'fs-to-dc --help' for help.\n\nError: Invalid value for '--archive-less-mature': '--env' is not a valid integer.\n"

It looks like `--archive-less-mature` is defined to expect an `int`, and is given a `None` by default. Which, erroneously works fine in `click<8.3`, and fails in `click>=8.3`.

It needs to be fixed, but for now I've added to your PR to install `click<8.3` in the test env, and to print more useful messages in the test failure logs.

(After typing this, I realised I probably should have pushed these two commits as a new PR, but, I also should have just fixed the error, but I can't do it now because it's time to finish working for the day...)

omad (approved):

PR #693, opened 2025-11-18
Switch to using uv

pjonsson:


omad (approved):
Awesome! Thank you.

Next step, turn it into a uv workspace to manage the different packages in the repo. 😃

pjonsson:
Well, yes, that is probably the way forward if this repository is going to be developed. Personally I was hoping I wouldn't need this repository, but it seems I'm stuck with it for a bit longer for now.

PR #698, opened 2025-11-19
utils: handle not having a self-href

pjonsson:
The current code in odc-stac will happily
return a self-href that is None, so handle
that situation without immediately crashing
with a UnboundLocalError in item_to_meta_uri.


pjonsson:
@alexgleith looks like you are the one who touched this code last, and it was fairly recently, so adding you as reviewer.

alexgleith (approved):
Nice one!

PR #700, opened 2025-11-23
item_to_meta_uri: pass asset_absolute_paths

pjonsson:
This is required for s3-to-dc to accept
STAC files with assets that have relative
URIs.


pjonsson:
This is the odc-tools part of https://github.com/opendatacube/datacube-core/pull/2264. Don't think it will have any effect (good or bad) until there is a released datacube with the linked PR.

SpacemanPaul (approved):

PR #702, opened 2025-11-25
s3-to-dc: set self_href if missing

pjonsson:
This will populate the uri_* columns in `odc.dataset` with `s3` and the bucket/path to the STAC file when passing in STAC files without a self-href/canonical in its links.


pjonsson:
I'd like to avoid adding more command line parameters unless we absolutely need them, is there a reason to not add this?

SpacemanPaul (approved):

PR #721, opened 2026-02-06
Update to click 8.3.1

pjonsson:
Click 8.3.1 is broken, but we need
it for datacube 1.9.14, so remove
the default value for
--archive-less-mature. This seems
to get a None when the flag isn't
passed, and the default value
otherwise.

### DIFF
--- a/apps/dc_tools/odc/apps/dc_tools/utils.py
+++ b/apps/dc_tools/odc/apps/dc_tools/utils.py
@@ -114,19 +114,14 @@ archive_less_mature = click.option(
     flag_value=500,
     type=int,
     help=(
-        "Archive existing any datasets that match product, "
-        "time and region-code, but have lower dataset-maturity."
-        "Note: An error will be raised and the dataset add will "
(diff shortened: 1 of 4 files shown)


pjonsson:
I'm sure there is some change of semantics of `--archive-less-mature` with this, but while the bug in Click has been fixed, there are no signs of a release anytime soon.

The PR that updates datacube-index to the latest version (https://github.com/opendatacube/datacube-docker/pull/244) also depends on this change.

Ariana-B (approved):

apps/dc_tools/odc/apps/dc_tools/utils.py
Ariana-B: Could we also take this opportunity to improve the help message:
```suggestion
"Archive any existing datasets that match product, "
"time, and region-code, but have a lower maturity. "
"If any matching datasets have higher or equal maturity, "
"the dataset add/update will fail.\n"
"Can specify a millisecond amount of leniency for comparing timestamps "
"to account for the possibility of slight differences between matching datasets "
"(default is 500ms)."
```

PR #722, opened 2026-02-06
Stop printing in tests

pjonsson:
The assert shows the output
if it fails, so stop flooding
the logs with noise.


pjonsson:
This is failing on datacube 1.9.14 changes, so it needs to be rebased after #721 is merged.

